PR #4, opened 2020-09-21
Add ZFS driver tests

benjamreis:


test_zfs_sr.py
```diff
@@ -0,0 +1,121 @@
+import pytest
+from lib.common import wait_for
+import time
+
+# Requirements:
+# - one XCP-ng host >= 8.2 with an additional unused disk for the SR
+# - access to XCP-ng RPM repository from the host
+
+@pytest.fixture(scope='module')
```
stormi, line 9: Not needed anymore, a `sr_disk` fixture is now available by default (through `conftest.py`)

test_zfs_sr.py
```diff
@@ -0,0 +1,121 @@
...
+class TestZFSSR:
+    sr = None
+    vm = None
+
+    def test_create_zfs_sr_without_zfs(self, host, sr_disk):
+        # This test must be the first in the series in this module
+        assert not host.file_exists('/usr/sbin/zpool'), \
+            "zfs must not be installed on the host at the beginning of the tests"
+        try:
+            # though it is expected to fail, result assigned to TestXFSSR.sr for teardown in case it succeeds
```
stormi, line 38: s/TestXFSSR.sr/TestZFSSR.sr/

test_zfs_sr.py
```diff
@@ -0,0 +1,121 @@
...
+            assert not sr.all_pbds_attached()
+            host.yum_install(['zfs'])
+            zfs_installed = True
+            sr.plug_pbds(verify=True)
+            sr.scan()
+        finally:
+            if not zfs_installed:
+                host.yum_install(['zfs'])
+
+    # *** End of tests with reboots
```
stormi, line 104: We also need a test to check "SR activation and scan fails if the ZFS mountpoint is not mounted (start from a working setup then "break" the mount). The scan must not appear like it succeeded and the SR was empty."

And also, if you have ideas: "Think of other situations where the scan could 'work' despite there being an issue with the underlying FS and thus remove all VDI metadata because it thinks it's empty. Test them."

test_zfs_sr.py
```diff
@@ -0,0 +1,121 @@
...
+    # Prerequisites: SR attached but empty
+    def test_destroy_sr(self, host):
+        TestZFSSR.sr.destroy(verify=True)
+
+    @classmethod
+    def teardown_class(cls):
+        if cls.sr is not None and cls.sr.exists():
+            if cls.vm is not None and cls.vm.exists():
+                cls.vm.destroy()
+            cls.sr.destroy(verify=True)
```
stormi, line 121: Best to add a newline at the end of file

test_zfs_sr.py
benjamreis: Yeah I saw the checklist, i'm not sure if I want to start the tests of migration or that first.

test_zfs_sr.py
benjamreis: Yes :+1:

PR #5, opened 2020-09-21
Add --sr-deviceconfig parameter for remote SR config

stormi:
Also add related fixtures sr_deviceconfig and cephfs_deviceconfig. For
the latter, the deviceconfig will be read first from CLI if provided,
else from data.py's defaults

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -36,6 +36,13 @@ def pytest_addoption(parser):
         default=[],
         help="VM keys or OVA URLs for tests that require several VMs",
     )
+    parser.addoption(
+        "--sr-device-config",
+        action="append",
+        default=[],
+        help="device-config keys and values for a remote SR. " \
+             "Example: 'server:10.0.0.1,serverpath:/vms,nfsversion:4.1'.",
+    )
 
 def host_data(hostname_or_ip):
     # read from data.py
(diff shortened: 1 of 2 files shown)


conftest.py
```diff
@@ -190,6 +197,38 @@ def running_linux_vm(imported_vm):
...
+        # SR device config from data.py defaults
+        try:
+            from data import DEFAULT_CEPHFS_DEVICECONFIG
+        except ImportError:
+            DEFAULT_CEPHFS_DEVICECONFIG = {}
+        if DEFAULT_CEPHFS_DEVICECONFIG:
+            config = DEFAULT_CEPHFS_DEVICECONFIG
+        else:
+            raise Exception("No default CephFS deviceconfig found, neither in CLI nor in data.py defaults")
+    return config
```
benjamreis, line 230: Shouldn't this be in a ceph dedicated file?

conftest.py
```diff
@@ -211,3 +250,10 @@ def pytest_generate_tests(metafunc):
         if not vm_lists:
             vm_lists = [None] # no --vms parameter does not mean skip the test, for us, it means use the default
         metafunc.parametrize("vm_refs", vm_lists, indirect=True, scope="module")
+    if "sr_deviceconfig" in metafunc.fixturenames:
+        configs = metafunc.config.getoption("sr_deviceconfig")
+        if not configs:
+            # No --sr-deviceconfig parameter doesn't mean skip the test.
+            # For us it means use the defaults.
+            configs = [None]
```
benjamreis, line 258: Why not  `[]`?

conftest.py
stormi: We don't know yet exactly how tests will be hierarchised and fixtures shared and there are no ceph test yet (well there are, but in a separate WIP PR that will need this), so I put it in the main conftest for now.

conftest.py
stormi: Because of the way fixtures work. To pytest, an empty list means "no fixtures, skip the tests that require it", so I provide a non-empty list with a fixture initialized to value None and then I handle that case in the fixture function.

stormi: And why it is a list and not a single value: because you can provide several values for the same fixture and then pytest will execute the tests for each fixture.

benjamreis: Thanks for the info :)

benjamreis (approved):

Wescoeur (approved):

PR #8, opened 2020-09-22
Refactor existing SR tests + test polishing

stormi:
- test files hierarchy
- driver-specific fixtures
- multi-host tests separated into intrapool and crosspool tests
- single-host tests reworked to make better use of fixtures and remove
  the need for a specific execution order
- other minor changes


tests/storage/cephfs/conftest.py
```diff
@@ -0,0 +1,45 @@
+import pytest
+
+@pytest.fixture(scope='session')
+def host_with_ceph(host):
+    assert not host.file_exists('/usr/sbin/mount.ceph'), \
+        "mount.ceph must not be installed on the host at the beginning of the tests"
+    host.yum_install(['centos-release-ceph-jewel'], enablerepo="base,extras", save_state=True)
+    host.yum_install(['ceph-common'], enablerepo="base,extras")
```
benjamreis, line 8: Perhaps we should one day allow to install more than 1 package with this method.

tests/storage/cephfs/conftest.py
```diff
@@ -0,0 +1,45 @@
...
+    yield host
+    # teardown
+    host.yum_restore_saved_state()
+
+@pytest.fixture(scope='session')
+def cephfs_device_config(sr_device_config):
+    if sr_device_config is not None:
+        # SR device config from CLI param
+        config = sr_device_config
```
benjamreis, line 17: Are we sure `sr_device_config` has been properly parsed here?

stormi: Yes, see the definition in the root conftest.py

tests/storage/cephfs/conftest.py
stormi: It's already possible, but that wouldn't work here because the release package adds a new repo so you need a second yum install call to install ceph-common.

benjamreis (approved):

Wescoeur (approved):

PR #10, opened 2020-10-12
Update tests for compatibility with Windows VMs

stormi:
- Require SSH + bash + user named "root" on the windows VM
- Use an alternate method for starting background processes on the VM
- Rename fixtures and functions that used to be specific to linux
- Wait for the management agent to be up before attempting other
  operations
- Always use the mini linux VM for SR creation tests.

### DIFF
--- a/lib/common.py
+++ b/lib/common.py
@@ -1,5 +1,6 @@
 import json
 import subprocess
+import tempfile
 import time
 from subprocess import CalledProcessError
 from uuid import UUID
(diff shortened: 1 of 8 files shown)


lib/common.py
```diff
@@ -455,38 +473,48 @@ def shutdown(self, force=False, verify=False):
     def try_get_and_store_ip(self):
         ip = self.param_get('networks', '0/ip', accept_unknown_key=True)
 
-        if not ip:
+        if not ip or ip.startswith('169.254.'):
```
benjamreis, line 476: I'd add a comment about this fake IP so we remember in the future.

stormi: Done

PR #11, opened 2020-10-12
Add export tests (zstd, gzip, uncompressed)

stormi:


lib/common.py
```diff
@@ -244,13 +244,21 @@ def xo_server_reconnect(self):
         xo_cli('server.enable', {'id': self.xo_srv_id})
         wait_for(self.xo_server_connected, timeout_secs=10)
 
-    def import_vm_url(self, url, sr_uuid=None):
-        print("Import VM %s on host %s" % (url, self))
-        params = {
-            'url': url
-        }
+    def import_vm(self, filename=None, url=None, sr_uuid=None):
+        assert filename is not None or url is not None
```
benjamreis, line 248: what if both are provided?

stormi: Then filename wins :)

benjamreis (approved):

PR #17, opened 2021-03-09
Use `yum history rollback` to restore yum state after tests:

benjamreis:
Previous version didn't support downgrade to a previous version, or install disbaled repo (testing, update etc...)
This version rollback to the last yum transaction before the tests and check that the same RPMs with same versions, releases, archs and epochs are installed.

### DIFF
--- a/lib/common.py
+++ b/lib/common.py
@@ -149,6 +149,7 @@ class Host:
         self.user = None
         self.password = None
         self.saved_packages_list = None
+        self.saved_rollback_id = None
 
     def __str__(self):
         return self.hostname_or_ip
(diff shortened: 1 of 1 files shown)


lib/common.py
```diff
@@ -329,9 +351,10 @@ def yum_remove_added_packages(self, base_packages):
             print("No added packages to remove.")
 
     def yum_restore_saved_state(self):
-        """ Currently only able to remove added packages """
-        assert self.saved_packages_list is not None, "Can't restore previous state without a package list"
-        self.yum_remove_added_packages(self.saved_packages_list)
+        """ Restore yum state to saved state """
+        assert self.saved_packages_list is not None and self.saved_rollback_id, "Can't restore previous state without a package list"
```
stormi, line 355: I'd rather have two different asserts, one for package list and one for rollback id, each with its own message.

lib/common.py
```diff
@@ -329,9 +351,10 @@ def yum_remove_added_packages(self, base_packages):
...
+        self.ssh(['yum', 'history', 'rollback', '--enablerepo=*', self.saved_rollback_id, '-y'])
+        assert self.saved_packages_list == self.packages()
```
stormi, line 357: If it fails, the output will be very hard to read. I would prefer a diff and assert that the diff is empty.

benjamreis: What about keeping this assert, with an error message ?

benjamreis: Diffing would add algo complexity to the tests

stormi: I think it's worth it to make the output usable. By the way you don't need to do the detailed comparison before you detect that there's a difference. When it's OK, the complexity will not be higher.

PR #19, opened 2021-03-11
GlusterFS auto tests

benjamreis:


benjamreis:
All tests are succesfull, but need [this fix](https://github.com/xcp-ng-rpms/blktap/commit/3b885573bc0cd41bfecf1103b2f68b4c2bb9f27b) comming in `blktap-3.37.2-1.0.2.xcpng8.2.x86_64`

tests/storage/glusterfs/test_glusterfs_sr.py
```diff
@@ -0,0 +1,90 @@
...
+
+    # *** tests with reboots (longer tests).
+
+    def test_reboot(self, host, glusterfs_sr, vm_on_glusterfs_sr):
+        sr = glusterfs_sr
+        vm = vm_on_glusterfs_sr
+        host.reboot(verify=True)
+        # Stop and restart volume to bring the bricks online (I don't know why this is required)
+        host.ssh(['gluster', '--mode=script', 'volume', 'stop', 'vol0'])
+        host.ssh(['gluster', '--mode=script', 'volume', 'start', 'vol0'])
```
Wescoeur, line 62: Well, It would be probably more interesting to find a solution to avoid these calls. Maybe in the driver or when the volumes are created. If each user must execute similar commands after a reboot, it's not very friendly. :smile:

benjamreis: testing without it.

tests/storage/glusterfs/conftest.py
```diff
@@ -0,0 +1,128 @@
...
+        pass
+    host.yum_restore_saved_state()
+
+    for h in host.pool.hosts:
+        hostname_or_ip = h.hostname_or_ip
+        if hostname_or_ip != host.hostname_or_ip:
+            host.ssh(['iptables', '-D', 'INPUT', '-p', 'all', '-s', hostname_or_ip, '-j', 'ACCEPT'])
+
+    for service in ['iptables', 'ip6tables']:
+        host.ssh(['cp', '/etc/sysconfig/%s.orig' % service, '/etc/sysconfig/%s' % service])
```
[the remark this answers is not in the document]
benjamreis: indeed

tests/storage/glusterfs/conftest.py
```diff
@@ -0,0 +1,128 @@
...
+def _setup_host_with_glusterfs(host):
+    assert not host.file_exists('/usr/sbin/glusterd'), \
+        "glusterfs-server must not be installed on the host at the beginning of the tests"
+    host.yum_install(['glusterfs-server'], save_state=True)
+    host.ssh(['systemctl', 'enable', '--now', 'glusterd.service'])
+
+    for h in host.pool.hosts:
+        hostname_or_ip = h.hostname_or_ip
+        if hostname_or_ip != host.hostname_or_ip:
+            host.ssh(['iptables', '-I', 'INPUT', '-p', 'all', '-s', hostname_or_ip, '-j', 'ACCEPT'])
```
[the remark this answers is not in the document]
benjamreis: no idea, need to see the driver's code

tests/storage/glusterfs/test_glusterfs_sr.py
benjamreis: tested successful i removed the lines.

tests/storage/glusterfs/conftest.py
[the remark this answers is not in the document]
stormi: The driver's code or rather probably gluster docs

tests/storage/glusterfs/conftest.py
```diff
@@ -0,0 +1,128 @@
...
+        host.ssh(['mv', '/etc/sysconfig/%s.orig' % service, '/etc/sysconfig/%s' % service])
+
+@pytest.fixture(scope='session')
+def pool_with_glusterfs(host):
+    for h in host.pool.hosts:
+        _setup_host_with_glusterfs(h)
+
+    yield
+    for h in host.pool.hosts:
+        _teardown_host_with_glusterfs(h)
```
stormi, line 41: If one host fails, we should collect the error, still do it on the others, and then raise with all errors grouped

tests/storage/glusterfs/conftest.py
```diff
@@ -0,0 +1,128 @@
...
+
+    gluster_cmd.append('force')
+    host.ssh(gluster_cmd)
+
+    host.ssh(['gluster', 'volume', 'start', 'vol0'])
+    yield
+    print("<< stop and delete gluster volume vol0")
+    host.ssh(['gluster', '--mode=script', 'volume', 'stop', 'vol0'])
+    host.ssh(['gluster', '--mode=script', 'volume', 'delete', 'vol0'])
+    host.ssh(['rm', '-rf', '/mnt/sr_disk/vol0'])
```
stormi, line 87: If one of the teardown commands fails we probably want to try the next ones.

This is a need we often have in teardown code. One of us should probably design or find a existing standard lib that allows for running several commands, collects their succes/error status, and only fails at the end with the list of collected errors.

tests/storage/glusterfs/conftest.py
```diff
@@ -0,0 +1,128 @@
...
+    assert len(hosts) > 1, "GlusterFS must be run on at least a 2 hosts pool"
+
+    host_probed = False
+    for h in hosts:
+        h.ssh(['mkdir', '-p', '/mnt/sr_disk/vol0/brick0'])
+        if h.hostname_or_ip != host.hostname_or_ip:
+            host.ssh(['gluster', 'peer', 'probe', h.hostname_or_ip])
+        elif not host_probed:
+            h.ssh(['gluster', 'peer', 'probe', host.hostname_or_ip])
+            host_probed = True # Only probe host once from one of the other hosts.
```
stormi, line 72: I'd rather move this out of the loop and always do it from the first host that isn't the master host (either the `hostA2` fixture which would remove the need to test for the pool length in your first two lines, or host.pool.hosts[1]).

tests/storage/glusterfs/conftest.py
```diff
@@ -0,0 +1,128 @@
...
+
+@pytest.fixture(scope='session')
+def gluster_volume_started(host, gluster_disk):
+    hosts = host.pool.hosts
+    assert len(hosts) > 1, "GlusterFS must be run on at least a 2 hosts pool"
+
+    host_probed = False
+    for h in hosts:
+        h.ssh(['mkdir', '-p', '/mnt/sr_disk/vol0/brick0'])
+        if h.hostname_or_ip != host.hostname_or_ip:
```
stormi, line 68: You can avoid testing for hostname_or_ip by iterating over hosts[1:] instead since the master host is always the first one in the pool. Or, to make it more readable, create a `Pool.members` method (decorated with `@property` to be used as an attribute).

tests/storage/glusterfs/conftest.py
```diff
@@ -0,0 +1,128 @@
...
+def glusterfs_sr(host, pool_with_glusterfs, gluster_volume_started, glusterfs_device_config):
+    """ A GlusterFS SR on first host. """
+    # Create the SR
+    sr = host.sr_create('glusterfs', "GlusterFS-SR", glusterfs_device_config, shared=True)
+    yield sr
+    # teardown
+    try:
+        sr.destroy()
+    except Exception as e:
+        print("SR destroy failed: %s" % e)
```
stormi, line 114: Can you test (not on a physical pool because this would leave things dirty) whether an exception raised in this fixture's teardown would stop the other fixtures from being teared down. I think they would still be, so it would be better to let the exception be raised.

tests/storage/glusterfs/conftest.py
```diff
@@ -0,0 +1,128 @@
...
+def vm_on_glusterfs_sr(host, glusterfs_sr, vm_ref):
+    print(">> ", end='')
+    vm = host.import_vm(vm_ref, sr_uuid=glusterfs_sr.uuid)
+    yield vm
+    # teardown
+    print("<< Destroy VM")
+    try:
+        vm.destroy(verify=True)
+    except Exception as e:
+        print("VM destroy failed: %s" % e)
```
stormi, line 127: Same comment as above. At some point I believed that other fixtures wouldn't be teared down, but I think they are in reality. What isn't executed is any command after the failure in the same fixture's teardown.

tests/storage/glusterfs/conftest.py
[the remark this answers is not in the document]
benjamreis: so there are several ports that need to be opened from my understanding, so it's simpler to keep all, and since the rules are removed at the end and its for testing purposes I think its acceptable.

tests/storage/glusterfs/conftest.py
benjamreis: I know for a fact that first exception raised in teardown stop the rest of the teardown
i had a lot of cleanings by hand to do why developping the tests x)

tests/storage/glusterfs/conftest.py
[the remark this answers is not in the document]
stormi: Opening just the required ports would still have two advantages:
- serve as documentation
- detect if (though unlikely) future glusterfs updates require more ports to be opened

tests/storage/glusterfs/conftest.py
```diff
@@ -0,0 +1,127 @@
...
+    host.ssh(['cp', '-f', '/etc/fstab', '/etc/fstab.orig'])
+    host.ssh(['echo', '/dev/sdb /mnt/sr_disk ext4 defaults 0 0', '>>/etc/fstab'])
+    host.ssh(['mount', '/mnt/sr_disk'])
+    yield
+    host.ssh(['cp', '-f', '/etc/fstab.orig', '/etc/fstab'])
+    try:
+        host.ssh(['umount', '/mnt/sr_disk'])
+        host.ssh(['rmdir', '/mnt/sr_disk'])
+    except Exception:
+        pass
```
stormi, line 72: I had missed this one

benjamreis:
Good to be merged!

stormi (approved):
Looks good. Let's not forget the squash before or during the merge.

Wescoeur (approved):

tests/storage/glusterfs/conftest.py
```diff
@@ -65,15 +65,18 @@ def pool_with_glusterfs(host):
...
+        h.ssh(['echo', '/dev/sdb /mnt/sr_disk ext4 defaults 0 0', '>>/etc/fstab'])
+        h.ssh(['mount', '/mnt/sr_disk'])
     yield
-    host.ssh(['cp', '-f', '/etc/fstab.orig', '/etc/fstab'])
-    host.ssh(['umount', '/mnt/sr_disk'])
-    host.ssh(['rmdir', '/mnt/sr_disk'])
+    for h in hosts:
+        h.ssh(['cp', '-f', '/etc/fstab.orig', '/etc/fstab'])
+        h.ssh(['umount', '/mnt/sr_disk'])
+        h.ssh(['rmdir', '/mnt/sr_disk'])
```
stormi: The sr_disk fixture only guarantees the presence of an available disk on the first host, so this will not fail nicely for other hosts if they don't have such a disk.

conftest.py
```diff
@@ -166,6 +166,15 @@ def sr_disk_wiped(host, sr_disk):
...
 
+@pytest.fixture(scope='session')
+def sr_disk_for_all_hosts(host, sr_disk):
+    for h in host.pool.hosts[1:]:
+        disks = h.disks()
+        # there must be at least 2 disks
+        assert len(disks) > 1, "at least two disks are required on all pool's hosts, missing on host: %s" % h
+        # Using the second disk for SR
+        print(">> a second disk for a local SR is present on host %s: %s" % (h, disk))
+
```
stormi, line 177: As the code in gluster_disk assumes that all hosts have the same device name for the second disk, I would like that we'd also verify that a disk named {sr_disk} exists on each host.

benjamreis:
Solves https://github.com/xcp-ng/xcp/issues/427

tests/storage/glusterfs/conftest.py
```diff
@@ -81,6 +82,45 @@ def gluster_disk(host, sr_disk_for_all_hosts):
...
+        teardown_for_host(host)
+    except Exception as e:
+        print("< Fallback teardown failed on master: %s, attempting to teardown other hosts" % e)
+        for h in host.pool.hosts[1:]:
+            try:
+                teardown_for_host(h)
+            except Exception as e:
+                print("< Fallback teardown failed on host: %s with error: %s" % (h, e))
+                pass
+
```
stormi, line 123: If even the fallback fails on master, I think we can just acknowledge that something went very wrong and stop here. The expected result of running the teardown on the host hosts after having maybe already half cleaned things up appears rather random to me.

tests/storage/glusterfs/conftest.py
```diff
@@ -81,6 +82,45 @@ def gluster_disk(host, sr_disk_for_all_hosts):
...
+        try:
+            # Volume might already be stopped if failure happened on delete
+            h.ssh(['gluster', '--mode=script', 'volume', 'stop', 'vol0'])
+        except Exception as e:
+            pass
+
+        h.ssh(['gluster', '--mode=script', 'volume', 'delete', 'vol0'])
+
+    try:
+        teardown_for_host(host)
```
stormi, line 114: I got umount errors but that was probably caused by the way I tested (cutting gluster traffic between two hosts), so I think it's OK.

tests/storage/glusterfs/conftest.py
```diff
@@ -81,6 +82,45 @@ def gluster_disk(host, sr_disk_for_all_hosts):
...
+        gluster_cmd = ['gluster', '--mode=script', 'volume', 'remove-brick', 'vol0', 'replica', '1']
+        for h2 in hosts:
+            if h.hostname_or_ip != h2.hostname_or_ip:
+                gluster_cmd.append('%s:/mnt/sr_disk/vol0/brick0' % h2.hostname_or_ip)
+        gluster_cmd.append('force')
+        h.ssh(gluster_cmd)
+
+        for h2 in hosts:
+            if h.hostname_or_ip != h2.hostname_or_ip:
+                h.ssh(['gluster', '--mode=script', 'peer', 'detach', h2.hostname_or_ip])
```
stormi, line 103: This should also be done in normal teardown.

benjamreis: why? gluster is uninstall after, so there won't peers anymore.

stormi: The configuration will likely remain on the system. Or can you attest that it won't?

tests/storage/glusterfs/conftest.py
```diff
@@ -0,0 +1,202 @@
...
+
+@pytest.fixture(scope='session')
+def gluster_disk(host, sr_disk_for_all_hosts):
+    sr_disk = sr_disk_for_all_hosts
+    device = '/dev/' + sr_disk
+    hosts = host.pool.hosts
+    for h in hosts:
+        print(">> Format sr_disk %s and mount it on host %s" % (sr_disk, h))
+        h.ssh(['mkfs.ext4', '-F', device])
+        h.ssh(['rm', '-rf', '/mnt/sr_disk']) # Make sure directory doesn't exist to avoid weird behavior in teardown
```
benjamreis, line 90:
```suggestion
        h.ssh(['rm', '-rf', '/mnt/sr_disk']) # Remove any existing leftover dir to ensure rmdir will not fail in teardown
```

PR #23, opened 2021-03-26
Add guest tools testing

stormi:


lib/common.py
```diff
@@ -514,6 +524,16 @@ def shutdown(self, force=False, verify=False):
             wait_for(self.is_halted, "Wait for VM halted")
         return ret
 
+    def reboot(self, force=False, verify=False):
+        print("Reboot VM")
+        ret = self.host.xe('vm-reboot', {'uuid': self.uuid, 'force': to_xapi_bool(force)})
+        if verify:
+            # a way to test that the reboot is happening is that SSH doesn't respond anymore
+            wait_for_not(self.is_ssh_up, "Wait for SSH down")
```
stormi, line 532: This is probably fragile, because it just failed on a VM here.

tests/guest-tools/unix/test_guest_tools_unix.py
```diff
@@ -0,0 +1,124 @@
+import pytest
+from lib.common import wait_for, wait_for_not
+
+# Requirements:
+# - a two host XCP-ng pool >= 8.0 (the second host is only used for the migration test)
+# - a unix VM supported by the Linux/install.sh script of the guest tools ISO
```
beshleman, line 6: I like seeing this in the test file, maybe it should be a standard for new tests?

tests/guest-tools/unix/test_guest_tools_unix.py
```diff
@@ -0,0 +1,124 @@
...
+        vm = running_vm
+        # migrate to default SR on hostA2
+        vm.migrate(hostA2, local_sr_on_hostA2)
+        wait_for(lambda: vm.all_vdis_on_host(hostA2), "Wait for all VDIs on destination host")
+        wait_for(lambda: vm.is_running_on_host(hostA2), "Wait for VM to be running on destination host")
+        self._check_tools_version(vm, state.tools_version)
+        self._check_os_info(vm, state.vm_distro)
+        # We don't migrate the VM back since the test module ends here and vm fixtures are module-scoped.
+        # However, if in the future we manage to mutualize VMs for different test modules, then
+        # we will need to either migrate it back, or create a throw-away clone before migrating.
```
beshleman, line 124: duly noted

lib/common.py
```diff
@@ -123,6 +123,16 @@ def scp(hostname_or_ip, src, dest, check=True, suppress_fingerprint_warnings=Tru
 def to_xapi_bool(b):
     return 'true' if b else 'false'
 
+def parse_xe_dict(xe_dict):
+    """ Parses a xe param containing keys and values, e.g. "major: 7; minor: 20; micro: 0; build: 3"
+        Data type remains str for all values.
+    """
```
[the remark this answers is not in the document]
stormi: If that's the standard way, why not.

tests/guest-tools/unix/test_guest_tools_unix.py
stormi: Yes, probably. Note that in some tests we've used a more formal and probably better form, where we also specify what command line arguments or parameters in data.py you need to feed to the test.

lib/common.py
```diff
@@ -677,6 +697,55 @@ def start_background_process(self, cmd):
...
+        """
+        return parse_xe_dict(self.param_get('PV-drivers-version'))
+
+    def tools_version(self):
+        """ Returns the tools version in the form major.minor.micro-build """
+        version_dict = self.tools_version_dict()
+        return "{major}.{minor}.{micro}-{build}".format(**version_dict)
+
+    def file_exists(self, filepath):
+        '''Test that the file at filepath exists.'''
```
benjamreis, line 731: Why simple quote here instead of double as in other pydocs?

benjamreis: We probably should decide on one and use always the same, even for strings in the code.

lib/common.py
```diff
@@ -677,6 +697,55 @@ def start_background_process(self, cmd):
...
+        return self.ssh_with_result(['test', '-f', filepath]).returncode == 0
+
+    def detect_package_manager(self):
+        """ Heuristic to determine the package manager on a unix distro """
+        if self.file_exists('/usr/bin/rpm') or self.file_exists('/bin/rpm'):
+            return 'rpm'
+        elif self.file_exists('/usr/bin/apt-get'):
+            return 'apt-get'
+        else:
+            return 'unknown'
```
benjamreis, line 741: Why returning `unknown` instead of throwing? Also using an enum would be safer.

tests/guest-tools/unix/test_guest_tools_unix.py
benjamreis: Yes we should normalize that as well.

tests/guest-tools/unix/test_guest_tools_unix.py
```diff
@@ -0,0 +1,124 @@
...
+    def test_check_tools(self, running_vm, state):
+        vm = running_vm
+        self._check_tools_version(vm, state.tools_version)
+        self._check_os_info(vm, state.vm_distro)
+
+    def test_check_tools_after_reboot(self, running_vm, state):
+        vm = running_vm
+        vm.reboot(verify=True)
+        self._check_tools_version(vm, state.tools_version)
+        self._check_os_info(vm, state.vm_distro)
```
benjamreis, line 93: This tests are not independant, if 1st fails, 2nd makes no sense.

lib/common.py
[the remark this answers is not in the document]
benjamreis: Is there something to put in the `setup.cfg` of pycodestyle to enforce that?

lib/common.py
stormi: Because I copy-pasted that function from Bobbye's PR :D
I'll fix it

lib/common.py
stormi: I don't want the caller to be forced to catch the exception. Agreed about an enum.

tests/guest-tools/unix/test_guest_tools_unix.py
stormi: Yes, the tests in this series depend on each other. I didn't want to put the whole installation test into a fixture because 
, that's why the class is decorated with `@pytest.mark.incremental`. If a test fails, the others are automatically skipped.

lib/common.py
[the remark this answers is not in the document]
stormi: Given the change would affect more than my PR, I suggest that a volunteer creates a dedicated PR for that.

lib/common.py
stormi: > We probably should decide on one and use always the same, even for strings in the code.

I usually use double quotes for multi word strings and single quotes for single-word strings, but that's debatable.

lib/common.py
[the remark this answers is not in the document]
benjamreis: Agreed, and this is not blocking for the merge

lib/common.py
benjamreis: yeah the exception could be raised by the caller if wanted.

benjamreis (approved):

PR #25, opened 2021-03-29
Add pydocstyle validation in a PR - ignore D10* in tests files

benjamreis:


lib/common.py
```diff
@@ -304,15 +302,13 @@ def get_last_yum_history_tid(self):
...
-    35 | install -y --enablerepo= | 2021-03-08 15:19 | Install        |    1
-    34 | remove -y ceph-common    | 2021-03-08 15:13 | Erase          |    1
-
+        Loaded plugins: fastestmirror
+        | ID   | Command line             | Date and time    | Action(s)      | Altered
+        -------------------------------------------------------------------------------
+        |   37 | install -y --enablerepo= | 2021-03-08 15:27 | Install        |    1
+        |   36 | remove ceph-common       | 2021-03-08 15:26 | Erase          |    1
+        |   35 | install -y --enablerepo= | 2021-03-08 15:19 | Install        |    1
+        |   34 | remove -y ceph-common    | 2021-03-08 15:13 | Erase          |    1
```
stormi, line 311: Adding the pipes makes the example output unparseable by the code below. I suggest to start each line with a dot and explain why the added dots in the comment.

Or, even better, maybe there's a special syntax that would tell pydocstyle that it is raw output not meant to be analyzed?

stormi:
Looks good to me as a first step. We may choose to enforce more rules in a subsequent PR.

PR #26, opened 2021-03-31
MooseFS SR functional tests

xandrus:


tests/storage/moosefs/test_moosefs_sr.py
```diff
@@ -0,0 +1,92 @@
...
+        vm.start()
+        vm.wait_for_os_booted()
+        vm.shutdown(verify=True)
+
+    #def test_snapshot(self, vm_on_moosefs_sr):
+    #    vm = vm_on_moosefs_sr
+    #    vm.start()
+    #    vm.wait_for_os_booted()
+    #    vm.test_snapshot_on_running_vm()
+    #    vm.shutdown(verify=True)
```
benjamreis, line 54: Why disabling snapshot tests?

xandrus:
Hi,

My fault.
One second and all snapshot will be enabled.

tests/storage/moosefs/conftest.py
```diff
@@ -0,0 +1,48 @@
...
+    assert not host.file_exists('/usr/sbin/mount.moosefs'), \
+        "mount.moosefs must not be installed on the host before all tests"
+    host.ssh(['sh', '-c', '"curl https://ppa.moosefs.com/RPM-GPG-KEY-MooseFS > /etc/pki/rpm-gpg/RPM-GPG-KEY-MooseFS"'])
+    host.ssh(['sh', '-c', '"curl http://ppa.moosefs.com/MooseFS-3-el7.repo > /etc/yum.repos.d/MooseFS.repo"'])
+    host.yum_install(['fuse'])
+    host.yum_install(['moosefs-client'])
+    yield host
+    # teardown
+    host.yum_remove(['moosefs-client'])
+    host.yum_remove(['fuse'])
```
benjamreis, line 14:
```suggestion
    host.yum_install(['fuse'], save_state=True)
    host.yum_install(['moosefs-client'])
    yield host
    # teardown
    host.yum_restore_saved_state()
    host.ssh(['sh', '-c', '"rm -f /etc/pki/rpm-gpg/RPM-GPG-KEY-MooseFS"'])
    host.ssh(['sh', '-c', '"rm -f /etc/yum.repos.d/MooseFS.repo"'])
```

tests/storage/moosefs/test_moosefs_sr.py
```diff
@@ -0,0 +1,92 @@
...
+            sr.destroy()
+            assert False, "MooseFS SR creation should failed!"
+
+    def test_create_and_destroy_sr(self, host_with_moosefs, moosefs_device_config):
+        # Create and destroy tested in the same test to leave the host as unchanged as possible
+        host = host_with_moosefs
+        sr = host.sr_create('moosefs', "MooseFS-SR-test", moosefs_device_config, shared=True, verify=True)
+        # import a VM in order to detect vm import issues here rather than in the vm_on_xfs_fixture used in
+        # the next tests, because errors in fixtures break teardown
+        vm = host.import_vm(vm_image('mini-linux-x86_64-bios'), sr_uuid=sr.uuid)
```
benjamreis, line 36: As for other storage tests can you use the `--vm` option and not always use `'mini-linux-x86_64-bios'`?

benjamreis (changes requested):

benjamreis:
Also we like to test intra and cross pool migrations which would require 3 test hosts (A1, A2 in same pool and B in another pool).
It means you'd need to setup all pool's A hosts with MooseFS and test migration between A1 and A2 on the shared SR and on a different one, and migration between A1 and B.

You can take a look at how its done for CephFS. :)

Thanks for your contribution, this is greatly appreciated!

conftest.py
```diff
@@ -79,6 +79,7 @@ def setup_host(hostname_or_ip):
 def hosts(request):
     # a list of master hosts, each from a different pool
     hostname_list = request.param.split(',')
+    print(hostname_list)
```
Wescoeur, line 82: Could you remove this line? :) It's not directly related to the PR.

xandrus: Absolutely yes :)

tests/storage/moosefs/conftest.py
```diff
@@ -0,0 +1,49 @@
+import pytest
+
+@pytest.fixture(scope='session')
+def host_with_moosefs(host):
+    assert not host.file_exists('/usr/sbin/mount.moosefs'), \
+        "mount.moosefs must not be installed on the host before all tests"
+    host.ssh(['sh', '-c', '"curl https://ppa.moosefs.com/RPM-GPG-KEY-MooseFS > /etc/pki/rpm-gpg/RPM-GPG-KEY-MooseFS"'])
+    host.ssh(['sh', '-c', '"curl http://ppa.moosefs.com/MooseFS-3-el7.repo > /etc/yum.repos.d/MooseFS.repo"'])
+    host.yum_install(['fuse'], save_state=True)
+    host.yum_install(['moosefs-client'])
```
Wescoeur, line 10: You should use this statement instead:
```python
host.yum_install(['fuse', 'moosefs-client'], save_state=True)
```
The saved state must be applied on the two packages.

xandrus: I had some problems with `save_state=True` statement and moosefs-client package.
Basically, the test fails when this option was set to true. I will investigate this topic.

Wescoeur: Sorry, ignore my previous comment. There are things to improve internally regarding the `yum_install` method, but in your use case the code is ok. :+1:

xandrus: OK,

Thanks!

tests/storage/moosefs/test_moosefs_sr.py
Wescoeur: The vm param is given to functions like `vm_on_XXX_sr`, it's ok to use a hardcoded VM image here. (It's already the case in other driver tests.)

xandrus: I have just make a commit connected with this thread: f3d3f2b
Please check if this type of solution meets your expectations.

tests/storage/moosefs/test_moosefs_sr.py
```diff
@@ -0,0 +1,94 @@
...
+    """
+    Tests that do not use fixtures that setup the SR or import VMs,
+    because they precisely need to test SR creation and destruction,
+    and VM import.
+    """
+
+    def test_create_moosefs_sr_without_mfsmount(self, host, moosefs_device_config, pass_vm_ref):
+        # This test must be the first in the series in this module
+        assert not host.file_exists('/usr/sbin/mount.moosefs'), \
+            "MooseFS client shound not be installed on the host"
```
Wescoeur, line 21: `should` :)

tests/storage/moosefs/test_moosefs_sr_intrapool_migration.py
```diff
@@ -0,0 +1,25 @@
...
+# - access to MooseFS packages repository: ppa.moosefs.com
+# From --hosts parameter:
+# - host(A1): first XCP-ng host >= 8.2 with an additional unused disk for the SR.
+# - hostA2: Second member of the pool. Can have any local SR. No need to specify it on CLI.
+# From --vm parameter
+# - A VM to import to the MooseFS SR
+
+@pytest.mark.usefixtures("hostA2", "local_sr_on_hostA2")
+class Test:
+    def test_cold_intrapool_shared_migrationa(self, host, hostA2, vm_on_moosefs_sr, moosefs_sr):
```
Wescoeur, line 15: `test_cold_intrapool_migration`

tests/storage/moosefs/test_moosefs_sr.py
Wescoeur: Like I said you can keep the previous code with the hardcoded 'mini-linux-x86_64-bios' VM, because we just want to test quickly the SR creation like in other drivers. :wink: It's sometimes useful to use a specific image using the CLI but not in this test.

tests/storage/moosefs/test_moosefs_sr.py
xandrus: Wow - I'm creating new words :)
Fixed!

tests/storage/moosefs/test_moosefs_sr_intrapool_migration.py
xandrus: Fixed!

xandrus:
OK, I returned to the hardcoded VM image.
Thanks!

PR #27, opened 2021-03-31
After reconnecting XO to a host, wait for the host object

stormi:
Cross pool migration tests tend to fail because we might attempt the
migration before XO has time to add the Host object in its database, so
we wait for the object to be available.


lib/common.py
```diff
@@ -260,6 +260,9 @@ def xo_server_reconnect(self):
         xo_cli('server.disable', {'id': self.xo_srv_id})
         xo_cli('server.enable', {'id': self.xo_srv_id})
         wait_for(self.xo_server_connected, timeout_secs=10)
```
benjamreis, line 262: Is this still necessary with the new code?

stormi: Which new code? XO?

benjamreis (approved):

PR #28, opened 2021-04-09
FreeBSD support in tests

stormi:
This PR brings better support for FreeBSD VMs in our tests. The price to pay for this is I needed to add bash to the requirements for test VMs. Among our existing VMs, all VMs already have it, except Alpine VMs, but we can add it if this PR is accepted.

* Guest tools tests are skipped
* uefistored tests will probably not work. It's difficult to install FreeBSD in UEFI mode anyway, so I can't even test and think it's acceptable for now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,11 +83,12 @@ Many tests expect VMs with:
 
 Here are instructions that should help creating such VMs.
 
-### Linux
+### Linux and other unixes
 
 * Eject installation ISO
 * Setup network
 * Install openssh-server and enable it
+* Install bash in order to ensure a common shell is available in all test VMs
 * Install guest tools, then eject guest tools ISO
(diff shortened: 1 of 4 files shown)


stormi:
Currently broken on our alpine VMs because they don't have bash installed yet.

gduperrey (approved):

benjamreis (approved):
A explanation of `-x` => `-f` would be nice :)

stormi:
> A explanation of `-x` => `-f` would be nice :)

I totally forgot what exactly it changed, but it was the way I found so that it works in all VMs from our test environment: alpine, debian, centos, windows through ssh, freebsd...

Wescoeur (approved):

PR #33, opened 2021-05-10
WIP: Support write append tests and therefore test uefistored v0.6.0

beshleman:


tests/uefistored/test_auth_var.py
```diff
@@ -39,25 +39,29 @@ def test_auth_variable(imported_vm):
     if vm.is_windows:
         pytest.skip('not valid test for Windows VMs')
 
-    vm.start()
-    vm.wait_for_vm_running_and_ssh_up()
+    try:
```
stormi, line 42: This commit should be squashed to the previous one.

I think my comment on previous commit about using a fixture for all this could still apply. Maybe not a "running VM" as it looks like the second test expects a shutdown VM (or does it care?).

In any case, we should really avoid the following as it reveals we are unclear about what to expect at this stage
```
        if not vm.is_running():
            vm.start()
```

beshleman:
The approach I took here ended up not working as planned. I'm closing this PR, as it is being succeeded by PR 57 https://github.com/xcp-ng/xcp-ng-tests/pull/57

PR #34, opened 2021-05-17
Add tests to check error when device is missing on LVM/EXT SRs

Wescoeur:

### DIFF
--- a/tests/storage/ext/test_ext_sr.py
+++ b/tests/storage/ext/test_ext_sr.py
@@ -1,4 +1,5 @@
 import pytest
+import subprocess
 from lib.common import wait_for, vm_image
 
 # Requirements:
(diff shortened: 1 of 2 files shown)


stormi (approved):
Looks good but maybe this would be a good candidate for extraction to a common lib for SR tests.

Wescoeur:
I agree with that. There are helpers like `cold_migration_then_come_back` or `live_storage_migration_then_come_back` in `common.py`, but I'm not sure it's a good idea to put all helpers in a single file. We can probably create a new `storage.py` file or something like that. What do you think? @benjamreis @stormi

benjamreis:
Yes a specific storage helper file would be good IMHO.
I'm not sure how much of the existing code should be mutualize though.
And probably in a separate PR as well.

stormi:
Agreed. Yes if the file contains more than just this function, a separate PR will be better and then I'll merge this one. As you wish.

About the file location, I don't know what's best, put it in lib or in `tests/storage`. Probably `tests/storage` if it's just test code mutualized (we can still extract storage-related stuff from `lib/common.py` such as the SR class to a storage.py file... Or to per class files)

Wescoeur:
Perfect, ok for `tests/storage`, I will create a new PR for that.

PR #37, opened 2021-06-02
Improve usage of xe commands + robustify calls

Wescoeur:


benjamreis (approved):

lib/common.py
```diff
@@ -505,7 +511,7 @@ def all_vdis_on_host(self, host):
                 return False
         return True
 
-    def export(self, filepath, compress="false"):
+    def export(self, filepath, compress=False):
```
stormi, line 514: This is a special case. `compress` may be 'false', 'true' or 'zstd', so maybe we shouldn't treat it as a boolean here and keep using a string instead.

tests/uefistored/test_secure_boot.py
```diff
@@ -87,7 +87,7 @@ def setup_and_cleanup(self, imported_sb_vm):
             pytest.skip('only valid for Linux VMs')
 
         self.PK, self.KEK, self.db, self.dbx = generate_keys()
-        vm.param_set('platform', 'secureboot', 'false')
+        vm.param_set('platform', 'secureboot', False)
```
stormi, line 90: Seeing this and the above situation where a given param can be either 'true', 'false' or 'zstd', I'm not so sure anymore about the "all bool" solution.

Also, seeing this line https://github.com/xcp-ng/xcp-ng-tests/blob/b90087589d7e1583095adbef115a52095dd0423d/lib/common.py#L319 where the read value from xe `param-get` is still a string feels like we're not being consistent. Bool as input, string as output...

Same with https://github.com/xcp-ng/xcp-ng-tests/blob/b90087589d7e1583095adbef115a52095dd0423d/lib/common.py#L940 and https://github.com/xcp-ng/xcp-ng-tests/blob/b90087589d7e1583095adbef115a52095dd0423d/lib/common.py#L979

I'm seriously starting to prefer that we'd just check that any param given to `xe` is a string, and let devs be responsible of the params they give to `xe`.

lib/common.py
Wescoeur: IHMO this parameter name `compress` is a source of error when you don't know the API, it really looks like a boolean. A function parameter like `compressMode` is already more precise using values like: "None", "Default" (or something else), "Zstd"...

Wescoeur: Ok, so we can use directly: `gzip`, `zstd` or anything else like `none`: https://github.com/xapi-project/xen-api/blob/550ff27340167239b33aa41b2f517c7771df9213/ocaml/xapi/export.ml#L821-L826

stormi: Perfect

stormi (approved):

stormi:
The commit message should probably be updated to reflect the latest changes

stormi:
So GH still doesn't offer a way to comment directly on commit messages it seems, so I'll use a comment here.

The first word of the commit messages should be capitalized, and the first line should not be too long.

PR #38, opened 2021-06-03
Add basic tests that don't require SSH access to the VM

stormi:
- new test file test_basic_without_ssh.py
- test_cross_pool_live_storage_migration.py improved and renamed to
  test_cross_pool_migration.py


stormi:
I don't know what pycodestyle wants with me. I thought we had disabled the "I want two blank spaces after a function" thing.

benjamreis:
There's plenty of rules with 2 blank spaces after etc... class, method, function each has one rule.
In your case you need to disable E305.

lib/common.py
```diff
@@ -877,6 +900,13 @@ def get_all_efi_bins(self):
 
         return binaries
 
+    def clone(self):
+        name = self.name() + '_clone_for_tests'
+        print("Clone VM")
+        uuid = self.host.xe('vm-clone', {'uuid': self.uuid, 'new-name-label': "\"%s\"" % name})
```
benjamreis, line 906: Perhaps the `xe` method should escape strings to avoid having to do thinkg like this?

conftest.py
```diff
@@ -200,8 +200,9 @@ def vm_refs(request):
 @pytest.fixture(scope="module")
 def imported_vm(host, vm_ref):
     if is_uuid(vm_ref):
-        print(">> Reuse VM %s on host %s" % (vm_ref, host))
         vm = VM(vm_ref, host)
+        name = vm.name()
```
Wescoeur, line 204: Why do you use a temporary variable in this context? :)

stormi: Because the name() method does a ssh call so I wanted to avoid nesting it in another function call.

Wescoeur: Ok, it makes sense. :+1:

tests/misc/test_cross_pool_migration.py
```diff
@@ -0,0 +1,20 @@
+import pytest
+from lib.common import wait_for, wait_for_not
+
+pytestmark = pytest.mark.default_vm('mini-linux-x86_64-bios')
+
+def test_cross_pool_migration(host, hostB1, imported_vm):
+    assert host.pool.uuid != hostB1.pool.uuid
+
+    vm = imported_vm.clone()
```
benjamreis, line 9: Why are you using a clone now?

benjamreis (approved):

tests/misc/test_cross_pool_migration.py
stormi: So that we can leave the original VM on the first pool, without having to move it back. Do you want me to add a comment?

stormi: Note: this is particularly useful when used with `--vm={uuid of an existing VM}`

benjamreis: No need, I was just curious ^^

PR #39, opened 2021-06-16
Add missing config.py file

Wescoeur:

### DIFF
--- a/lib/config.py
+++ b/lib/config.py
@@ -1,0 +1,1 @@
+ignore_ssh_banner = False


lib/config.py
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env python
```
stormi, line 1: Why is there a shebang ? It's not meant to be executed directly, is it?

stormi (approved):

PR #41, opened 2021-06-30
Improve logging in tester

benjamreis:
`--log-cli-level=(debug|info|warning|error|critical)` to change log level (default: info)
`--ssh-output-max-lines=nbr of lines` (default 20) to set a max number of lines when logging return of an ssh command

Use python logging module to report logs to pytest
Add datetime and level to logs
Log command and error when ssh'ed commands fail
Log command and result when debug enabled for all ssh'ed commands


lib/common.py
```diff
@@ -139,6 +140,9 @@ def ssh(hostname_or_ip, cmd, check=True, simple_output=True, suppress_fingerprin
                 raise SSHCommandFailed(255, "SSH Error: %s" % banner_res.stdout.decode(), command)
             stdout = stdout[len(banner_res.stdout.decode()):]
 
+        if res.returncode:
+            logger.error("Got error {} '{}' while running '{}'".format(res.returncode, stdout, command))
+
         if check and res.returncode:
             raise SSHCommandFailed(res.returncode, stdout, command)
```
benjamreis, line 147: I can do this if you prefer?
```py
        if res.returncode:
            logger.error("Got error {} '{}' while running '{}'".format(res.returncode, stdout, command))
            if check:
                raise SSHCommandFailed(res.returncode, stdout, command)
```

benjamreis:
```py
        if res.returncode:
            if check:
                logger.error("Got error {} '{}' while running '{}'".format(res.returncode, stdout, command))
                raise SSHCommandFailed(res.returncode, stdout, command)
            else:
                logger.warning("Got error {} '{}' while running '{}'".format(res.returncode, stdout, command))
```

Something like this?

lib/logger.py
```diff
@@ -0,0 +1,26 @@
...
+    ERROR = "ERROR"
+
+DATE_FORMAT = "%b %d %H:%M:%S"
+
+def _log(level, message, end):
+    time = datetime.now()
+    print("{} {}: {}".format(time.strftime(DATE_FORMAT), level.value, message), end=end)
+
+def debug(message, end='\n'):
+    _log(LogLevel.DEBUG, message, end)
```
Wescoeur, line 17: It would be interesting to add a variable in the project config to specify the level of verbosity. More particularly to hide the debug logs.

lib/common.py
Wescoeur: Maybe. :thinking: @stormi ?

lib/logger.py
benjamreis: Yes I thought about it. I'm looking at how to do it.

Wescoeur: Take a look at: https://github.com/xcp-ng/xcp-ng-tests/blob/master/lib/config.py ;)

Wescoeur: And: https://github.com/xcp-ng/xcp-ng-tests/blob/master/conftest.py#L67

lib/common.py
stormi: Sometimes we run commands that are expected to fail so that would not be a warning, rather a debug.

pytest.ini
```diff
@@ -4,3 +4,7 @@ markers =
     default_vm
     default_vms
     incremental
+log_cli = 1
+log_cli_level = info
+log_cli_format = %(asctime)s %(levelname)s %(message)s
+log_cli_date_format = %b %d %H:%M:%S
```
stormi, line 10: Better add a newline at end of file

tests/misc/test_basic_without_ssh.py
```diff
@@ -1,4 +1,6 @@
+import logging
 import pytest
+
```
stormi, line 3: These header code style changes should maybe belong to another PR?

conftest.py
```diff
@@ -63,9 +64,16 @@ def pytest_addoption(parser):
...
+    parser.addoption(
+        "--ssh-debug-output-max-lines",
+        action="store",
+        default=20,
+        help="Max lines to output in a debug log (0 if no limit)"
+    )
 
 def pytest_configure(config):
     global_config.ignore_ssh_banner = config.getoption('--ignore-ssh-banner')
+    global_config.max_log_lines = int(config.getoption('--ssh-debug-output-max-lines'))
```
stormi, line 76: I think it would be better to make the CLI option and the stored parameter to have the same name. `max_log_lines` is shorter and seems fine.

benjamreis: the name would be confusing imho.

tests/misc/test_basic_without_ssh.py
benjamreis: I added an import as well i don't think making a specific PR for this is relevant since they'd be no functional changes at all so a revert would have no consequences.

conftest.py
stormi: Then isn't it confusing in config.py already?

benjamreis: Probably ^^'
But I think we should find a better name which would be used for both

benjamreis: `ssh-output-max-lines` is more concise since it has nothing to do with debu specifically

stormi: Looks good

tests/misc/test_basic_without_ssh.py
stormi: Ok

stormi (approved):

PR #43, opened 2021-09-28
XAPI plugins tests

benjamreis:

### DIFF
--- a/tests/xapi-plugins/test_lsblk.py
+++ b/tests/xapi-plugins/test_lsblk.py
@@ -1,2 +1,6 @@
+# Requirements:
+# From --hosts parameter:
+# - host(A1): first XCP-ng host > 8.2.
+
 def test_list_block_devices(host):
     host.call_plugin('lsblk.py', 'list_block_devices')
--- a/tests/xapi-plugins/test_lsblk.py
+++ b/tests/xapi-plugins/test_lsblk.py
@@ -1,0 +1,2 @@
+def test_list_block_devices(host):
+    host.call_plugin('lsblk.py', 'list_block_devices')
--- a/tests/xapi-plugins/test_netdata.py
+++ b/tests/xapi-plugins/test_netdata.py
@@ -1,0 +1,40 @@
+import pytest
+
+# Requirements:
+# From --hosts parameter:
+# - host(A1): first XCP-ng host > 8.2.
+# And:
+# - access to XCP-ng RPM repository from hostA1
+
+@pytest.fixture(scope='session')
+def host_without_netdata(host):
+    assert not host.call_plugin('netdata.py', 'is_netdata_installed')
+    yield host
+
+class TestInstall:
+    def test_is_netdata_installed(self, host):
+        host.call_plugin('netdata.py', 'is_netdata_installed')
+
+    def test_install_netdata(self, host_without_netdata):
+        host = host_without_netdata
+        host.yum_save_state()
+        host.call_plugin('netdata.py', 'install_netdata', {
+            "api_key": "dummy_key", "destination": "127.0.0.1:19999"
+        })
+
+        assert host.call_plugin('netdata.py', 'is_netdata_installed')
+
+        host.yum_restore_saved_state()
+
+class TestApiKey:
+    def test_get_netdata_api_key(self, host_without_netdata):
+        host = host_without_netdata
+        host.yum_save_state()
+        host.call_plugin('netdata.py', 'install_netdata', {
+            "api_key": "dummy_key", "destination": "127.0.0.1:19999"
+        })
+
+        res = host.call_plugin('netdata.py', 'get_netdata_api_key')
+        assert res == "dummy_key"
+
+        host.yum_restore_saved_state()
--- a/tests/xapi-plugins/test_raid.py
+++ b/tests/xapi-plugins/test_raid.py
@@ -1,0 +1,33 @@
+import logging
+import pytest
+
+# Requirements:
+# From --hosts parameter:
+# - host(A1): first XCP-ng host > 8.2.
+
+@pytest.fixture(scope='session')
+def host_with_raid(host):
+    dummy_raid = False
+    if not host.file_exists('/dev/md127'):
+        logging.info("> Host has no raids, creating one for tests")
+        dummy_raid = True
+        host.ssh(['dd', 'if=/dev/zero', 'of=raid-0', 'bs=1M', 'count=200'])
+        host.ssh(['cp', 'raid-0', 'raid-1'])
+        host.ssh(['losetup', '/dev/loop0', 'raid-0'])
+        host.ssh(['losetup', '/dev/loop1', 'raid-1'])
+        host.ssh([
+            'mdadm', '--create', '/dev/md127', '--run', '--level=mirror', '--raid-devices=2', '/dev/loop0', '/dev/loop1'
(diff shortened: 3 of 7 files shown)


benjamreis:
For now tests will only test if the plugin does not raise and some basic easy assert.
To be enhanced in a future PR.

tests/xapi-plugins/test_raid.py
```diff
@@ -0,0 +1,2 @@
+def test_check_raid_pool(host):
+    host.call_plugin('raid.py', 'check_raid_pool')
```
stormi, line 2: Doesn't this raise when there's no RAID, currently?

tests/xapi-plugins/test_zfs.py
```diff
@@ -0,0 +1,7 @@
+import json
+
+from tests.storage.zfs.conftest import zpool_vol0, host_with_zfs
+
+def test_list_zfs_pools(host, zpool_vol0):
+    res = host.call_plugin('zfs.py', 'list_zfs_pools')
+    assert json.loads(res).get("vol0") is not None
```
stormi, line 7: I guess this will raise when there's no vol0, so the test won't work in that case?

tests/xapi-plugins/test_updater.py
```diff
@@ -0,0 +1,36 @@
...
+        host.call_plugin('updater.py', 'check_update')
+
+    def test_update(self, host):
+        host.yum_save_state()
+        host.call_plugin('updater.py', 'update')
+
+        res = host.call_plugin('updater.py', 'check_update')
+        assert res == "[]"
+
+        host.yum_restore_saved_state()
```
stormi, line 14: Does yum_restore_saved_state still work when no changes have been made to the host?

tests/xapi-plugins/test_netdata.py
```diff
@@ -0,0 +1,26 @@
+class TestInstall:
+    def test_is_netdata_installed(self, host):
+        host.call_plugin('netdata.py', 'is_netdata_installed')
+
+    def test_install_netdata(self, host):
```
stormi, line 5: Maybe we should assert that netdata is not installed already? Or make that a prerequisite through a fixture?

stormi: And then we could also test what happens if you call install_netdata again when it's already installed. Two situations here by the way:
* netdata installed by install_netdata then try again
* netdata installed manually then try install_netdata (is the api key correctly set?)

tests/xapi-plugins/test_lsblk.py
```diff
@@ -0,0 +1,2 @@
+def test_list_block_devices(host):
```
stormi, line 1: The requirements should specify that you need the lsblk plugin. For convenience, you could check that the plugin exists and skip if it doesn't, as we haven't released it yet.

stormi (commented):
I think each test file should have a comment that says what the requirements are, even if it's a single host with nothing in particular. BTW, since the output has changed for most of them in the master branch of xcp-ng-xapi-plugins, I suppose some of them at least require recent enough plugins. You could probably skip their execution based on the version of the xcp-ng-xapi-plugins RPM.

tests/xapi-plugins/test_zfs.py
benjamreis: fixture `zpool_vol0` creates a `vol0`

stormi: Ah, indeed

tests/xapi-plugins/test_netdata.py
benjamreis: we always assume that no package are installed in other tests (mainly for storage drivers) so i think we should do the same here.
A fixture would be a good way to enforce that.

tests/xapi-plugins/test_updater.py
benjamreis: Good question

tests/xapi-plugins/test_raid.py
benjamreis: Yes, i'll add a fixture that creates the RAID if it doesn't exist and that'll remove it if it has been created for the tests.

tests/xapi-plugins/test_netdata.py
stormi: > we always assume that no package are installed in other tests (mainly for storage drivers) so i think we should do the same here. A fixture would be a good way to enforce that.

In other tests, you don't have a plugin that may try to install things again, nor competition between users who main install it manually and the plugin that may install it, that's why I suggested that, after installing netdata once, we'd test what happens if you tell the plugin to install it again.

benjamreis: I pushed a fixture enforcing netdata is NOT installed on the host. :+1:

tests/xapi-plugins/test_raid.py
stormi: Do I wait for this or do I approve now?

benjamreis: please wait, i'm working on it, I'd like to be sure no test raise unexpectidely before we merge. :)

benjamreis: You can approve now, after checking my fixture :)

tests/xapi-plugins/test_raid.py
```diff
@@ -0,0 +1,36 @@
...
+
+# Requirements:
+# From --hosts parameter:
+# - host(A1): first XCP-ng host > 8.2.
+
+@pytest.fixture(scope='session')
+def host_with_raid(host):
+    dummy_raid = False
+    try:
+        host.call_plugin('raid.py', 'check_raid_pool')
```
stormi, line 12: I would prefer that you'd simply check that `/dev/md127` doesn't exist on the host.

tests/xapi-plugins/test_raid.py
```diff
@@ -0,0 +1,36 @@
...
+        host.ssh(['losetup', '/dev/loop1', 'raid-1'])
+        host.ssh([
+            'mdadm', '--create', '/dev/md127', '--run', '--level=mirror', '--raid-devices=2', '/dev/loop0', '/dev/loop1'
+        ])
+
+    yield host
+
+    if dummy_raid:
+        logging.info("< Destroying raid created for tests")
+        host.ssh(['mdadm', '--stop', '/dev/md127'])
```
stormi, line 29: If this fails, the files won't be deleted

tests/xapi-plugins/test_raid.py
```diff
@@ -0,0 +1,36 @@
...
+        host.ssh(['losetup', '-d', '/dev/loop0'])
+        host.ssh(['losetup', '-d', '/dev/loop1'])
```
stormi, line 31: Same for those

tests/xapi-plugins/test_raid.py
```diff
@@ -0,0 +1,34 @@
...
+
+# Requirements:
+# From --hosts parameter:
+# - host(A1): first XCP-ng host > 8.2.
+
+@pytest.fixture(scope='session')
+def host_with_raid(host):
+    dummy_raid = False
+    if not host.file_exists('/dev/md127'):
+        # TODO: check the error is no raid
```
stormi, line 12: ಠ_ಠ

stormi (approved):

PR #47, opened 2021-10-07
Fix and improve uefistored tests

stormi:
Many fixes and improvements to Guest secure boot tests.

### DIFF
--- a/tests/uefistored/test_secure_boot.py
+++ b/tests/uefistored/test_secure_boot.py
@@ -167,6 +167,16 @@ class TestGuestLinuxUEFISecureBoot:
         vm.param_set('platform', 'secureboot', True)
         boot_and_check_sb_succeeded(vm)
 
+    def test_sb_off_really_means_off(self, uefi_vm):
+        vm = uefi_vm
+        vm.install_uefi_certs([self.PK, self.KEK, self.db])
+        sign_efi_bins(vm, self.db)
+        vm.param_set('platform', 'secureboot', False)
+        vm.start()
+        vm.wait_for_vm_running_and_ssh_up()
+        logging.info("Check that SB is NOT enabled according to the OS.")
+        assert not vm.booted_with_secureboot()
+
 
 @pytest.mark.usefixtures("pool_without_uefi_certs")
 class TestGuestWindowsUEFISecureBoot:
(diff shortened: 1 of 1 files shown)


Wescoeur (approved):

tests/uefistored/conftest.py
```diff
@@ -1,31 +1,41 @@
...
 @pytest.fixture(scope='module')
-def imported_sb_vm(imported_vm):
-    vm = imported_vm
+def pool_without_uefi_certs(host):
+    pool = host.pool
 
-    if not vm.is_uefi:
-        pytest.skip('imported_sb_vm can only be used on UEFI VMs')
+    # Save the certs.
+    # This may fail if the pool certs are not in a consistent state, and prevent the tests from running.
```
beshleman, line 9: I wonder if we should place this in the README too. A gentle reminder about using secureboot-certs to initialize a pool for tests, perhaps?

lib/common.py
```diff
@@ -871,6 +964,95 @@ def clone(self):
...
+                    10
+                )
+                ret = True
+            except TimeOutException as e:
+                logging.debug(e)
+                pass
+        finally:
+            self.host.ssh(['screen', '-S', session, '-X', 'quit'], check=False)
+            self.host.ssh(['rm', '-f', tmp_file], check=False)
+        return ret
```
beshleman, line 1053: This whole section is not as bad as I was expecting given the comments! :+1:

tests/uefistored/conftest.py
stormi: You don't need to initialize the pool with secureboot-certs. The recorded state will just be that no certs are installed, so all certs will be deleted during teardown in this situation.

I've just added a Requirements comment on top of the file that says that the state must be consistent. As there is no way to enter an inconsistent state using secureboot-certs (in theory), I don't think it needs to be in the main README.

lib/common.py
```diff
@@ -111,6 +117,93 @@ def first_shared_sr(self):
...
+                         % (' '.join(saved_certs.keys()) if saved_certs else 'no certs'))
+        else:
+            for tmp_file in saved_certs.values():
+                self.master.ssh(['rm', '-f', tmp_file])
+            raise Exception(
+                (
+                    "Can't save pool UEFI certs. Only %s certs are defined, "
+                    "which wouldn't be restorable as is with secureboot-certs install"
+                )
+                % '&'.join(saved_certs.keys())
```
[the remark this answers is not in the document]
stormi: Done

lib/common.py
```diff
@@ -25,6 +25,9 @@ def vm_image(vm_key):
         url = DEF_VM_URL + url
     return url
 
+class TimeOutException(Exception):
```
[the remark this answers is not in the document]
stormi: Thanks, this is better indeed.

PR #54, opened 2021-11-04
Add tests/system/test_irqbalance.py to check VIF irq balancing

stormi:
Following a regression in a kernel hotfix, check that IRQs associated to
VIFs are not all affected to the same CPU, to avoid any future similar
regression.

### DIFF
--- a/lib/common.py
+++ b/lib/common.py
@@ -1092,6 +1092,25 @@ class VIF:
         self.uuid = uuid
         self.vm = vm
 
+    def param_get(self, param_name, key=None, accept_unknown_key=False):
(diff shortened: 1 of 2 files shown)


stormi:
I ran the tests with a kernel that has the regression and one that hasn't. Results as expected.

Sometimes the tests fail because one of the VMs doesn't start completely. Might be transient errors on the lab but might be something else related to the fact that I start 4 VMs nearly at the same time.

benjamreis (approved):

Wescoeur (approved):

PR #56, opened 2021-12-02
scripts/ansible: support VM updates with ansible

beshleman:
This PR adds support for updating test VMs using Ansible.

Currently, this script does the following:

1. Parses an Ansible playbook for host names
2. Searches for XVAs located at DEF_VM_URL (usually http://pxe/images)
3. Matches the XVAs to the playbook host names
4. For each match:
   1. Create the VM by importing the XVA
    2. Apply the Ansible playbook to the VM
    3. Shutdown the VM
    4. Export the VM to `--export-directory'` argument OR the default `/root/ansible-updates/`
  

Future work:

1. Allow exporting the VM remotely, such as back to `http://pxe/images`
2. Use the Ansible XenServer plugin to create the VMs instead of using the script to create the inventory. Currently, the XenServer plugin doesn't support creating a VM by importing an XVA, but instead only by creating one from a template.

### DIFF
--- a/scripts/ansible/runner.py
+++ b/scripts/ansible/runner.py
@@ -1,0 +1,257 @@
+#!/usr/bin/env python
+
+import argparse
+import atexit
+import logging
+import os
+import sys
+import tempfile
+import __main__
+from signal import SIGABRT, SIGINT, SIGQUIT, SIGTERM, signal
+from subprocess import run
+from typing import List
+
+import requests
+import yaml
+from bs4 import BeautifulSoup
+
+try:
+    from yaml import CDumper as Dumper
+    from yaml import CLoader as Loader
+except ImportError:
+    from yaml import Dumper, Loader
+
+cur = os.path.split(__main__.__file__)[0]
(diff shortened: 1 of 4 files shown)


stormi:
Is it reviewable yet?

scripts/ansible/runner.py
```diff
@@ -0,0 +1,257 @@
...
+            "-x",
+            type=str,
+            help="The XCP-ng hostname or IP address. Defaults to host found in data.py",
+        )
+        parser.add_argument(
+            "--export-directory",
+            "-e",
+            type=str,
+            required=False,
+            default=os.path.join("/", "root", "ansible-updates/"),
```
beshleman, line 168: This might be feasible for XCP-ng at large, but would maybe be best to have this resolve to the user's home.

benjamreis (approved):

stormi (approved):
To be able to make useful comments I'd need to try it first but have no time for it right now so I'm happy to merge it as is. Don't be surprised if in the future I start questionning this or that because I actually tried it :D

PR #57, opened 2021-12-02
Add DB append tests

beshleman:
This PR adds support for testing the append of the UEFI DB variable. This was the cause of the KB update bug in Windows.

The test has only been tested on Alpine and it requires `efi-updatevar`. It will fail without `efi-updatevar`.

### DIFF
--- a/tests/uefistored/test_auth_var.py
+++ b/tests/uefistored/test_auth_var.py
@@ -44,21 +44,25 @@ def test_auth_variable(imported_vm):
         pytest.skip('not valid test for Windows VMs')
 
     vm.start()
-    vm.wait_for_vm_running_and_ssh_up()
 
-    cert = Certificate()
+    try:
+        vm.wait_for_vm_running_and_ssh_up()
 
-    # Set the variable
-    set_and_assert_var(vm, cert, b'I am old news', should_pass=True)
+        cert = Certificate()
 
-    # Set the variable with new data, signed by the same cert
-    set_and_assert_var(vm, cert, b'I am new news', should_pass=True)
+        # Set the variable
+        set_and_assert_var(vm, cert, b'I am old news', should_pass=True)
 
-    # Remove var
-    set_and_assert_var(vm, cert, b'', should_pass=True)
+        # Set the variable with new data, signed by the same cert
+        set_and_assert_var(vm, cert, b'I am new news', should_pass=True)
 
-    # Set the variable with new data, signed by the same cert
-    set_and_assert_var(vm, cert, b'new data', should_pass=True)
+        # Remove var
+        set_and_assert_var(vm, cert, b'', should_pass=True)
 
-    # Set the variable with new data, signed by a different cert
-    set_and_assert_var(vm, Certificate(), b'this should fail', should_pass=False)
+        # Set the variable with new data, signed by the same cert
+        set_and_assert_var(vm, cert, b'new data', should_pass=True)
+
+        # Set the variable with new data, signed by a different cert
+        set_and_assert_var(vm, Certificate(), b'this should fail', should_pass=False)
+    finally:
+        vm.shutdown(verify=True)
--- a/tests/uefistored/test_auth_var.py
+++ b/tests/uefistored/test_auth_var.py
@@ -3,6 +3,8 @@ import pytest
 from lib.commands import SSHCommandFailed
 from lib.efi import (
     Certificate,
+    EFIAuth,
+    EFI_GUID_STRS,
     EFI_GLOBAL_VARIABLE_GUID,
     EFI_GLOBAL_VARIABLE_GUID_STR,
     EFI_AT_ATTRS_BYTES,
@@ -14,7 +16,8 @@ from lib.efi import (
 # From --hosts parameter:
 # - host: XCP-ng host >= 8.2 (+ updates)
 # From --vm parameter
-# - A UEFI VM to import
+# - A Linux UEFI VM to import
+# - The UEFI VM must have efitools installed
 
 def set_and_assert_var(vm, cert, new, should_pass):
     var = 'myvariable'
@@ -38,11 +41,8 @@ def set_and_assert_var(vm, cert, new, should_pass):
         assert not ok, 'This var should not have successfully set'
 
 
-def test_auth_variable(imported_vm):
-    vm = imported_vm
-    if vm.is_windows:
-        pytest.skip('not valid test for Windows VMs')
-
+def test_auth_variable(linux_uefi_vm):
+    vm = linux_uefi_vm
     vm.start()
 
     try:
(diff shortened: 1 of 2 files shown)


tests/uefistored/test_auth_var.py
```diff
@@ -15,6 +17,7 @@
 # - host: XCP-ng host >= 8.2 (+ updates)
 # From --vm parameter
 # - A UEFI VM to import
```
stormi, line 19:
```suggestion
# - A Linux UEFI VM to import
```

tests/uefistored/test_auth_var.py
```diff
@@ -62,3 +65,53 @@ def test_auth_variable(imported_vm):
 
     # Set the variable with new data, signed by a different cert
     set_and_assert_var(vm, Certificate(), b'this should fail', should_pass=False)
+
+    vm.shutdown(verify=True)
```
stormi, line 69: The test this snippet belongs to (github won't let me comment the lines higher) should probably depend on the linux_uefi_vm fixture too, and the `if vm.is_windows` check be removed from it.

tests/uefistored/test_auth_var.py
```diff
@@ -62,3 +65,53 @@ def test_auth_variable(imported_vm):
...
+    vm.ssh(
+        [
+            "efi-updatevar",
+            "-k",
+            vm_kek_key,
+            "-c",
+            vm_db_cert,
+            "-a",
+            "db",
+        ]
```
stormi, line 109: I don't find this much readable for ssh commands, but I won't argue :)

tests/uefistored/test_auth_var.py
```diff
@@ -62,3 +65,53 @@ def test_auth_variable(imported_vm):
 
     # Set the variable with new data, signed by a different cert
     set_and_assert_var(vm, Certificate(), b'this should fail', should_pass=False)
+
+    vm.shutdown(verify=True)
+
+
+def test_db_append(linux_uefi_vm):
+    """Pass if appending the DB succeeds. Otherwise, fail."""
```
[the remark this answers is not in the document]
benjamreis: This rule is not enforced by `pydocstyle`?
Should it?

benjamreis (commented):
Nothing to add to Sam review. Will approve once conversation are resolved. :)

tests/uefistored/test_auth_var.py
Wescoeur: Why not this instead:
```python
vm.ssh([
    "efi-updatevar",
    "-k",
    vm_kek_key,
    "-c",
    vm_db_cert,
    "db"
])
```

A single level of indentation is quite sufficient. :)

Wescoeur: Or one line: `vm.ssh(["efi-updatevar", "-k", vm_kek_key, "-c", vm_db_cert, "db"])`.
The max char limit is not reached.

tests/uefistored/test_auth_var.py
```diff
@@ -3,6 +3,8 @@
 from lib.commands import SSHCommandFailed
 from lib.efi import (
     Certificate,
+    EFIAuth,
```
Wescoeur, line 6: Maybe `EFI_AUTH` to use a unified syntax?

tests/uefistored/test_auth_var.py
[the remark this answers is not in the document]
beshleman: Why change [PEP257](https://www.python.org/dev/peps/pep-0257/#one-line-docstrings)?

For one-line docstrings, it states:

> The docstring is a phrase ending in a period.

Which is why...

```python
def foo():
    """ Do nothing. """
```

Results in:

```bash
$ pydocstyle /tmp/foo.py

/tmp/foo.py:4 in public function `foo`:
        D210: No whitespaces allowed surrounding docstring text
```

tests/uefistored/test_auth_var.py
beshleman: Unfortunately `EFIAuth` is a class and so should probably stay CamelCase, as opposed to the `UPPER_SNAKECASE` variables, which are just "constant" primitives.

I would like to eventually replace the GUIDs with an object that has `as_str()` and `as_bytes()` methods, which will clean up the verbose exports, but that's probably for another PR...

Wescoeur: Oh! I misread the code. :sweat_smile:

tests/uefistored/test_auth_var.py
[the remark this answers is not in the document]
benjamreis: Please rebase your branch on master, this will add new checks regarding code & doc style.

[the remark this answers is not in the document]
stormi: The current convention in our repositories is with whitespaces (I can find exceptions, such as VM.set_efi_var and VM.file_exists). I don't think we realized PEP257 advised otherwise.

If we want to change that, we'll have to do it globally (in a separate PR), and enforce it in our pydocstyle settings. Regardless of my own taste, it's probably better indeed.

tests/uefistored/test_auth_var.py
```diff
@@ -62,3 +65,53 @@ def test_auth_variable(imported_vm):
 
     # Set the variable with new data, signed by a different cert
     set_and_assert_var(vm, Certificate(), b'this should fail', should_pass=False)
+
+    vm.shutdown(verify=True)
+
+
+def test_db_append(linux_uefi_vm):
+    """Pass if appending the DB succeeds. Otherwise, fail."""
+    vm = linux_uefi_vm
```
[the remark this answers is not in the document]
beshleman: I had to add this a little bit lower, because the VM needs to be booted before being able to detect this.

stormi (approved):

benjamreis (approved):
Please rebase on master to add the codestyle checks.

stormi:
Could you rebase on master so that the style checks can run on the PR?

beshleman:
> Please rebase on master to add the codestyle checks.

Done.

Not sure if I should merge or not, so leaving it as is!

benjamreis:
IDK why the doc & code style checks are not run on this PR...
Is there a `main.yaml` & a `format.yaml` in your sources after the rebase? In `.github/workflows`

benjamreis:
Does the commands `pydocstyle` and `pycodestyle` return with no error at the root of the repo on your branch?

benjamreis:
From https://github.com/xcp-ng/xcp-ng-tests/pull/64 the code style is okay.
So this is good to be merged for me.
It seems a simple rebase doesn't trigger the GH actions.

I'll let your merge the Pr when you want :)

beshleman:
> From #64 the code style is okay. So this is good to be merged for me. It seems a simple rebase doesn't trigger the GH actions.
> 
> I'll let your merge the Pr when you want :)

looks good from root:

```
[bobbye@piano xcp-ng-tests]$ pydocstyle
[bobbye@piano xcp-ng-tests]$ pycodestyle
```

Will merge now :)

PR #59, opened 2021-12-07
Run of XAPI's `quicktest`

benjamreis:
See: https://github.com/xapi-project/xen-api/tree/master/ocaml/quicktest

The test fails in case of:
- Quicktest failure
- Leaked XAPI object (missing or bad cleanup)

Otherwise success.

The result is printed for debug mode or thrown in case of error.

### DIFF
--- a/tests/quicktest/test_quicktest.py
+++ b/tests/quicktest/test_quicktest.py
@@ -1,0 +1,12 @@
+import logging
+
+# Requirements:
+# From --host parameter:
+# - A XCP-ng host.
+#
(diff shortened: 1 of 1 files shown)


Wescoeur (approved):

stormi (commented):
LGTM but it would be better if the output were streamed to debug output rather than displayed only when it finishes.

stormi (approved):

benjamreis:
It's complex, I'll try to do it in a separate PR if time allows.
I'll create an issue

PR #62, opened 2021-12-09
Improve EFI module and VM EFI method ergonomics

beshleman:
Some parts of the EFI code are quite cumbersome, so this PR attempts to try and help that.

It depends on #57.

The main refactoring is in refactoring the GUID and is described in more detail in the commit message for `lib/efi: refactor GUID strings and bytes into GUID class`.

The second refactoring is simply eliminating the return of the attributes in `vm.get_efi_var()` since they are never used and are unlikely to be used in the future.

### DIFF
--- a/tests/uefistored/test_auth_var.py
+++ b/tests/uefistored/test_auth_var.py
@@ -4,9 +4,7 @@ from lib.commands import SSHCommandFailed
 from lib.efi import (
     Certificate,
     EFIAuth,
-    EFI_GUID_STRS,
-    EFI_GLOBAL_VARIABLE_GUID,
-    EFI_GLOBAL_VARIABLE_GUID_STR,
+    global_variable_guid,
     EFI_AT_ATTRS_BYTES,
 )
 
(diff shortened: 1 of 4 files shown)


tests/uefistored/test_auth_var.py
```diff
@@ -83,14 +81,16 @@ def test_db_append(linux_uefi_vm):
 
     assert vm.ssh_with_result(["which", "efi-updatevar"]).returncode == 0, "This test requires efi-updatevar"
 
+    old = vm.get_efi_var(db.name, db.guid)
+
+    assert old != b"", "db failed to install"
```
stormi, line 86: I think this assert is always true due to the change you made to the get_efi_var function.

beshleman: If `/sys/firmware/efi/efivars/db-GUID` is empty, this will fire (common.py:936). This happened to me once, but unlikely to happen again. I think had accidentally used the wrong GUID with the old API, if I'm remember correctly. I think it can safely be removed.

stormi (approved):

benjamreis (approved):

PR #68, opened 2022-02-07
Use scp to fix TestProxies tests

Wescoeur:
We can't use a json string argument on a host using host.xe
because we must already escape the simple/double quotes
of the ssh and xe commands. JSON introduces a new level of quotes...
Otherwise we have a bad shell interpretation of the quotes on the server.

Solution: we use scp to copy a script that contains the xe command on the host.

### DIFF
--- a/lib/common.py
+++ b/lib/common.py
@@ -248,7 +248,7 @@ class Host:
             suppress_fingerprint_warnings=suppress_fingerprint_warnings, local_dest=local_dest
         )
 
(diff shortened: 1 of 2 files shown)


stormi:
The commit message is ambiguous: at first I thought you were fixing tests that are using scp.

benjamreis (approved):

stormi (approved):

PR #70, opened 2022-02-28
Fix host evacuation tests

benjamreis:
- SSH doesn't returns a decoded string
- Do not give empty args to 'reconfigure-ip' methods

### DIFF
--- a/tests/migration/test_host_evacuate.py
+++ b/tests/migration/test_host_evacuate.py
@@ -10,7 +10,8 @@ from tests.storage.nfs.conftest import vm_on_nfs_sr, nfs_sr, nfs_device_config
 # From --vm parameter
 # - A VM to import
 # From --second-network parameter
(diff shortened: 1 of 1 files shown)


tests/migration/test_host_evacuate.py
```diff
@@ -10,7 +10,7 @@
 # From --vm parameter
 # - A VM to import
 # From --second-network parameter
-# - second_network: A 2nd network of the pool, NOT the management interface, with PIF plugged on all hosts
+# - second_network: A 2nd physical network of the pool, NOT the management interface, with PIF plugged and configred with an IP on all hosts
```
stormi, line 13:
```suggestion
# - second_network: A 2nd physical network of the pool, NOT the management interface, with PIF plugged and configured with an IP on all hosts
```

stormi (commented):
In commit message: s/SSH doesn't returns/SSH doesn't return/

stormi (approved):

PR #72, opened 2022-03-07
Do not test for raid file as a regular file

benjamreis:
`/dev/md127` is NOT a regular file therefore:
`test -f` will always return `false`

Use `test -e` instead.

Add `regular_file=True` to `file_exists` to enable
the check of non regular file existence.

`test -e` only checks a file existence no matter what type it is.

### DIFF
--- a/lib/common.py
+++ b/lib/common.py
@@ -490,8 +490,9 @@ class Host:
         disks.sort()
         return disks
 
-    def file_exists(self, filepath):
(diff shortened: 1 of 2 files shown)


stormi (approved):

stormi:
> We can discuss to have an Enum of file type to also check the type is correct. `-b` for example in case of `/dev/md127`

We can probably make it evolve in this direction later if the need arises.

Wescoeur (approved):

PR #73, opened 2022-03-18
Split lib/common.py into several files (one per class)

stormi:
The PR also contains a change in the way we initialize hosts, made necessary to solve an import ordering cycle.

PR #75, opened 2022-03-18
Autodetect unused disks for SR tests (sr_disk fixture and such)

stormi:
And more or less related changes.

Depends on #73

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -170,12 +170,11 @@ def local_sr_on_hostB1(hostB1):
 
 @pytest.fixture(scope='session')
 def sr_disk(host):
-    disks = host.disks()
-    # there must be at least 2 disks
-    assert len(disks) > 1, "at least two disks are required on the first host"
-    # Using the second disk for SR
-    disk = disks[1]
-    logging.info(">> a second disk for a local SR is present on hostA1: %s" % disk)
+    logging.info(">> Check for the presence of a free disk device on the master host")
+    disks = host.available_disks()
+    assert len(disks) > 0, "a free disk device is required on the master host"
+    disk = disks[0]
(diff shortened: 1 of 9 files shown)


conftest.py
```diff
@@ -184,15 +184,22 @@ def sr_disk_wiped(host, sr_disk):
     yield sr_disk
 
 @pytest.fixture(scope='session')
-def sr_disk_for_all_hosts(host, sr_disk):
+def sr_disk_for_all_hosts(host):
+    master_disks = host.available_disks()
+    assert len(master_disks) > 0, "a free disk device is required on the master host"
```
benjamreis, line 189: the assert can be put in `available_disks` perhaps with a `raise` flag. To avoir duplicating the assert and error message.

benjamreis (changes requested):
A lot of modifs come from the split of `common.py`, please re ask for a review once, this PR is rebased after the split has been merged.

conftest.py
Wescoeur: I'm not sure it's a good idea: `available_disks` is clear, it returns the available disks, IMHO an assert in this helper is equal to a bad naming.

Wescoeur: Also `available_disks` is used in the for loop after this line without `assert`. ;)

stormi: If there's no consensus, then I guess I won't change it :D

benjamreis: Tht's why I proposed a flag to param it ;)

Wescoeur: Sometimes, it's not a problem to duplicate a line, if it's not abusive. ^^"

stormi:
I ran storage tests (except CephFS, Linstor and MooseFS for whose we need make it easier to run the tests internally) with success (with a minimal VM).

benjamreis (approved):

Wescoeur (approved):

PR #76, opened 2022-03-21
Add --sr-disk parameter and make it mandatory for tests that need it

stormi:
The value given to --sr-disk can be either 'auto' or the name of a
device, such as 'sdb'.

The parameter is now mandatory for any test that requires either the
`sr_disk` or the `sr_disk_for_all_hosts` fixtures.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -71,6 +71,13 @@ def pytest_addoption(parser):
         default=20,
         help="Max lines to output in a ssh log (0 if no limit)"
     )
+    parser.addoption(
+        "--sr-disk",
+        action="append",
+        default=[],
+        help="Name of an available disk (sdb) or partition device (sdb2) to be formatted and used in storage tests. "
+             "Set it to 'auto' to let the fixtures auto-detect available disks."
+    )
 
 def pytest_configure(config):
     global_config.ignore_ssh_banner = config.getoption('--ignore-ssh-banner')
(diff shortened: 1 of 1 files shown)


stormi:
I've tested various combinations of parameters and situations:
* --sr-disk not specified on a test that needs it
* --sr-disk not specified on a test that doesn't need it
* --sr-disk set to auto when there is an available disk on master
* --sr-disk set to auto when there is an available disk on master and other hosts (sr_disk_for_all_hosts fixture)
* --sr-disk set to sda when sda is the disk the system is installed on
* --sr-disk set to a value that doesn't exist
* --sr-disk set to sdb, disk available on master but busy on second host (sr_disk_for_all_hosts fixture)
* --sr-disk set to auto when only master has an available disk but not the second pool host (sr_disk_for_all_hosts fixture)

Storage tests are currently running.

conftest.py
```diff
@@ -71,6 +71,13 @@ def pytest_addoption(parser):
         default=20,
         help="Max lines to output in a ssh log (0 if no limit)"
     )
+    parser.addoption(
+        "--sr-disk",
+        action="append",
+        default=[],
+        help="Name of an available disk (sdb) or partition device (sdb2) to be formatted and used in storage tests. "
+             "Set it to 'auto' to let the fixtures auto-detect available disks."
```
benjamreis, line 79: `auto` as default value?

gduperrey (approved):

conftest.py
stormi: This would make the parameter not mandatory anymore.

benjamreis (approved):

Wescoeur (approved):

PR #79, opened 2022-03-22
LVM SR: retrigger the GC in case of SR destroy failure...

stormi:
... If the failure is due to a not empty SR.

We found out that successively running the test_cold_intrapool_migration and
the test_live_intrapool_migration may leave a VDI chain at teardown,
after the VM destroy, that is not correctly garbage collected before the SR
destroy (the parent VDI remains).

```
Mar 22 19:22:31 r620-q2 SMGC: [29353] SR 4659 ('LVM-local-SR') (6 VDIs in 5 VHD trees):
Mar 22 19:22:31 r620-q2 SMGC: [29353]         *8341e744[VHD](2.000G//2.012G|n)
Mar 22 19:22:31 r620-q2 SMGC: [29353]             *53b46c10[VHD](2.000G//40.000M|n)
Mar 22 19:22:31 r620-q2 SMGC: [29353]         *e7657971[VHD](2.000G//2.012G|n)
Mar 22 19:22:31 r620-q2 SMGC: [29353]         *43b9baf6[VHD](2.000G//8.000M|n)
Mar 22 19:22:31 r620-q2 SMGC: [29353]         *080f8024[VHD](2.000G//8.000M|n)
Mar 22 19:22:31 r620-q2 SMGC: [29353]         *4f3fa3ba[VHD](2.000G//172.000M|n)
Mar 22 19:22:31 r620-q2 SMGC: [29353]
Mar 22 19:22:31 r620-q2 SMGC: [29353] Found 5 VDIs for deletion:
Mar 22 19:22:31 r620-q2 SMGC: [29353]   *53b46c10[VHD](2.000G//40.000M|n)
Mar 22 19:22:31 r620-q2 SMGC: [29353]   *e7657971[VHD](2.000G//2.012G|n)
Mar 22 19:22:31 r620-q2 SMGC: [29353]   *43b9baf6[VHD](2.000G//8.000M|n)
Mar 22 19:22:31 r620-q2 SMGC: [29353]   *080f8024[VHD](2.000G//8.000M|n)
Mar 22 19:22:31 r620-q2 SMGC: [29353]   *4f3fa3ba[VHD](2.000G//172.000M|n)
```

Re-attach the SR and force another garbage collection (that removed the last VDI
in our tests) and then attempt another sr delete.

Here's what the second GC run looks like:

```
Mar 22 19:22:43 r620-q2 SMGC: [30330] SR 4659 ('LVM-local-SR') (1 VDIs in 1 VHD trees):
[...]

PR #80, opened 2022-03-23
Trigger GC run and wait for it to complete before destroying a SR

stormi:
(this supersedes PR #79)

We encountered situations where the VDIs were not garbage collected
correctly when sr destroy is called, despite the fact that the driver's
delete method did trigger a forced GC run.

Based on upstream advice, we run sr scan to trigger a GC run and then
wait for the GC run to complete, before attempting the SR deletion.

See https://github.com/xapi-project/sm/issues/583

**Update: I changed the way to do again, because this made the teardown too long when the GC takes a long time to wake up**

### DIFF
--- a/lib/common.py
+++ b/lib/common.py
@@ -58,3 +58,7 @@ def parse_xe_dict(xe_dict):
         key, value = pair.split(':')
         res[key.strip()] = value.strip()
     return res
+
+def safe_split(text, sep=','):
+    """ A split function that returns an empty list if the input string is empty. """
+    return text.split(sep) if len(text) > 0 else []
--- a/lib/sr.py
+++ b/lib/sr.py
@@ -2,18 +2,20 @@ import logging
 
 import lib.commands as commands
 
-from lib.common import wait_for, wait_for_not
+from lib.common import safe_split, wait_for, wait_for_not
 
 class SR:
     def __init__(self, uuid, pool):
         self.uuid = uuid
         self.pool = pool
+        self._is_shared = None # cached value for is_shared()
+        self._main_host = None # cached value for main_host()
 
     def pbd_uuids(self):
         return self.pool.master.xe('pbd-list', {'sr-uuid': self.uuid}, minimal=True).split(',')
 
     def unplug_pbds(self, force=False):
-        logging.info("Unplug PBDs")
+        logging.info(f"Unplug PBDs for SR {self.uuid}")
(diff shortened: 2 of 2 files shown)


Wescoeur (approved):

gduperrey (approved):

benjamreis (approved):

stormi:
New way of trying to solve the SR destroy problem.

benjamreis (approved):

gduperrey (approved):

lib/host.py
```diff
@@ -62,6 +62,12 @@ def stringify(key, value):
                 simple_output=simple_output
             )
 
+        if split is not None:
+            result_as_list = []
+            if len(result) > 0:
+                result_as_list = result.split(split)
+            return result_as_list
```
Wescoeur, line 69: I'm not sure if this is the right place for this code. This helper can be generic.
I think this new `split` param can be removed,  and this block could be moved in a new function executed after the `xe` call.

lib/sr.py
```diff
@@ -38,12 +40,48 @@ def plug_pbds(self, verify=True):
         if verify:
             wait_for(self.all_pbds_attached, "Wait for PDBs attached")
 
+    def vdi_uuids(self, managed=False):
+        return self.pool.master.xe('vdi-list', {'sr-uuid': self.uuid, 'managed': managed}, minimal=True, split=',')
```
Wescoeur, line 44: So regarding my previous comment, you can have this code instead:
```python
    return non_empty_split(
        self.pool.master.xe('vdi-list', {'sr-uuid': self.uuid, 'managed': managed}, minimal=True)
    )
```

Wescoeur (changes requested):

Wescoeur (approved):

lib/common.py
```diff
@@ -58,3 +58,9 @@ def parse_xe_dict(xe_dict):
         key, value = pair.split(':')
         res[key.strip()] = value.strip()
     return res
+
+def safe_split(text, sep=','):
+    """ A split function that returns an empty list if the input string is empty. """
+    if len(text) == 0:
+        return []
+    return text.split(sep)
```
benjamreis, line 66:
```suggestion
    return return [] if !len(text) else text.split(sep)
```

benjamreis (changes requested):

benjamreis (approved):

PR #85, opened 2022-03-29
Fix SB detection on CentOS 7 VMs

stormi:
Previously, we would call tail -c1 directly, and it worked in almost
all our test VMs, but it turns out CentOS 7 can't handle tail -c1 on
that special file (can't "seek").

So we need to cat and then pipe into tail.

### DIFF
--- a/lib/vm.py
+++ b/lib/vm.py
@@ -425,8 +425,11 @@ class VM(BaseVM):
                 "Got: %s" % output
             )
         else:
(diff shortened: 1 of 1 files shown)


stormi:
Tested on all our UEFI VM XVAs.

benjamreis (approved):

andSmv (approved):

PR #86, opened 2022-03-29
Improve host-evacuate tests

stormi:
* Improve the second_network fixture so that the tests that require it
  fail nicely when it doesn't meet the expected criteria: is present,
  the uuid points at an existing network which has a PIF and an IP address
  on master host, and is not the management network.
* Improve the requirements comment section. In particular, stress out
  the fact that the NFS SR must not depend on the "second network".
* Mark the tests incremental and put a test that requires the
  second_network fixture, so that test execution fails fast when the
  parameter is missing. It's still possible to directly call the only test
  that doesn't require this fixture.
* Add logs.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -49,7 +49,7 @@ def pytest_addoption(parser):
     )
     parser.addoption(
         "--second-network",
-        action="append",
+        action="store",
         default=None,
         help="UUID of second network in the A pool, NOT the management network"
     )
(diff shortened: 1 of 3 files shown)


stormi:
I could not test the last 3 test cases yet because my second network is needed to access the NFS share.

conftest.py
```diff
@@ -314,8 +314,19 @@ def additional_repos(request, hosts):
...
 @pytest.fixture(scope='session')
-def second_network(request):
-    return request.param
+def second_network(request, host):
+    if request.param is None:
+        pytest.fail("This test requires the --second-network parameter!")
+    network_uuid = request.param
+    pif_uuid = host.xe('pif-list', {'host-uuid': host.uuid, 'network-uuid': network_uuid}, minimal=True)
+    if pif_uuid == "":
+        pytest.fail("The provided --second-network UUID doesn't exist or doesn't have a PIF on master host")
```
benjamreis, line 323: The network should have an PIF with IP configured on all pools hosts.

benjamreis (changes requested):

conftest.py
stormi: I know but didn't want to push the tests too far and assumed a normally configured pool would have a PIF and IP for all hosts of the pool if master has.

stormi: The goal was to avoid common failures at fixture initialisation, but I think it's acceptable if in the case when the pool would be heterogeneous the test would fail later.

benjamreis: Fine by me :)

benjamreis (approved):

benjamreis (approved):

lib/vm.py
```diff
@@ -36,7 +36,8 @@ def is_paused(self):
 
     # By xe design on must be an host name-label
     def start(self, on=None):
-        logging.info("Start VM")
+        msg_starts_on = f" (on host {on})" if on is not None else None
```
Wescoeur, line 39:
```python
msg_starts_on = None if on else f" (on host {on})"
```
Maybe more clear?

stormi:
```
msg_starts_on = f" (on host {on})" if on else None
```

Wescoeur (approved):

PR #91, opened 2022-04-04
Refactor uefistored conftest

benjamreis:
New fixture `unix_uefi_vm_and_snapshot`

### DIFF
--- a/tests/uefistored/conftest.py
+++ b/tests/uefistored/conftest.py
@@ -18,10 +18,8 @@ def pool_without_uefi_certs(host):
     pool.restore_uefi_certs()
 
 @pytest.fixture(scope='module')
-def uefi_vm_and_snapshot(imported_vm):
-    vm = imported_vm
-    if not vm.is_uefi:
-        pytest.skip('imported_sb_vm can only be used on UEFI VMs')
+def uefi_vm_and_snapshot(uefi_vm):
+    vm = uefi_vm
 
     # Any VM that has been booted at least once comes with some
     # UEFI variable state, so simply clear the state of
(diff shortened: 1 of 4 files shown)


benjamreis:
Tests ok on linux and Windows

tests/uefistored/conftest.py
```diff
@@ -18,10 +18,16 @@ def pool_without_uefi_certs(host):
...
 
 @pytest.fixture(scope='module')
-def uefi_vm_and_snapshot(imported_vm):
+def uefi_vm(imported_vm):
     vm = imported_vm
     if not vm.is_uefi:
-        pytest.skip('imported_sb_vm can only be used on UEFI VMs')
+        pytest.skip('uefi_vm can only be used on UEFI VMs')
+
+    yield vm
```
stormi, line 26: We can move it to the global conftest.py now.

tests/uefistored/conftest.py
```diff
@@ -31,19 +37,18 @@ def uefi_vm_and_snapshot(imported_vm):
...
-    vm, _ = uefi_vm_and_snapshot
-    yield vm
-
 @pytest.fixture(scope='module')
 def linux_uefi_vm(uefi_vm):
     vm = uefi_vm
     if vm.is_windows:
         pytest.skip("linux_uefi_vm can only be used on Linux VMs")
 
     yield vm
```
stormi, line 50: While we're at refactoring, this fixture is not necessary anymore. It's just a combination of `unix_vm` (which will be available in global conftest once Gaël's PR is merged) and `uefi_vm`, and we can use them instead.

gduperrey (approved):

tests/uefistored/conftest.py
benjamreis: Should it be named `unix_...` as well?

stormi: Probably

stormi (approved):

PR #92, opened 2022-04-05
Modify method to format new hard disk with glusterfs

gduperrey:

### DIFF
--- a/tests/storage/glusterfs/conftest.py
+++ b/tests/storage/glusterfs/conftest.py
@@ -2,6 +2,7 @@ import logging
 import pytest
 
 from lib.commands import SSHCommandFailed
+from lib.common import setup_formatted_and_mounted_disk, teardown_formatted_and_mounted_disk
 
 GLUSTERFS_PORTS = [('24007', 'tcp'), ('49152:49251', 'tcp')]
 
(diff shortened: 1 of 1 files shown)


tests/storage/glusterfs/conftest.py
```diff
@@ -1,7 +1,7 @@
 import logging
 import pytest
-
```
benjamreis, line 3: Do not remove this empty line.

tests/storage/glusterfs/conftest.py
```diff
@@ -88,22 +88,15 @@ def pool_with_glusterfs(host):
...
-        h.ssh(['echo', '%s /mnt/sr_disk xfs defaults 0 0' % device, '>>/etc/fstab'])
-        h.ssh(['mount', '/mnt/sr_disk'])
+        setup_formatted_and_mounted_disk(h, sr_disk, 'xfs', mountpoint)
     yield
     for h in hosts:
         logging.info("<< Restore fstab and unmount /mnt/sr_disk on host %s" % h)
-        h.ssh(['cp', '-f', '/etc/fstab.orig', '/etc/fstab'])
-        h.ssh(['umount', '/mnt/sr_disk'])
-        h.ssh(['rmdir', '/mnt/sr_disk'])
+        teardown_formatted_and_mounted_disk(h, mountpoint)
```
benjamreis, line 99: Would this log be relevant in the called fonctions?

benjamreis (changes requested):

tests/storage/glusterfs/conftest.py
gduperrey: why ?
do we need to have an empty line between import and from lines ?

tests/storage/glusterfs/conftest.py
gduperrey: i don't know, but it doesn't seem to hurt, so i just replace the exact lines and not that one.

tests/storage/glusterfs/conftest.py
benjamreis: IMHO yes, this ease readability.

gduperrey: ok, i will change it. I didn't know, as it wasn't a comment on the iso pr ;)

tests/storage/glusterfs/conftest.py
stormi: Second answer to "why": because it is not related to the objective of your commit. We don't want to mix unrelated changes together. Formatting changes that you want to do to would belong to a separate commit (and a different PR usually), if relevant.

tests/storage/glusterfs/conftest.py
stormi: Second answer to "why": because it is not related to the objective of your commit. We don't want to mix unrelated changes together. Formatting changes that you want to do to would belong to a separate commit (and a different PR usually), if relevant.

tests/storage/glusterfs/conftest.py
benjamreis: I looked at the code. Logs are already in the functions so you can remove the logs as well :)

tests/storage/glusterfs/conftest.py
gduperrey: i don't want to do any formatting change... i just keep writing as i do with the iso pr for that part, so i just delet an empty line when i add an import one. But the explanation from Benjamin is enough. I need to have that empty line, si i will put it back. And i won't change something else in this PR.

benjamreis (approved):
LGTM after rebased into 1 commit.

stormi (approved):

stormi:
Last minute request: in commit message, "hardisk" should be "harddisk" or "hard disk"

Wescoeur (approved):

PR #106, opened 2022-05-05
Make sure `PK.auth` is never missing

benjamreis:
Test that if XAPI clear UEFI certificates `PK.auth` is fetched from a fallback dir

See: https://github.com/xapi-project/xen-api/pull/4696

### DIFF
--- a/tests/uefistored/test_cert_inheritance.py
+++ b/tests/uefistored/test_cert_inheritance.py
@@ -331,7 +331,7 @@ class TestPoolToDiskCertPropagationToAllHosts:
         vm, snapshot = uefi_vm_and_snapshot
         for h in host.pool.hosts:
             logging.info(f"Check host {h} has no certificate on disk except PK.")
-            # Make sure PK has been fetched from fallback dir so none SB UEFI VMs can still be booted
+            # Make sure PK has been fetched from fallback dir so UEFI VMs without certs can still be booted
             assert h.file_exists(f'{CERT_DIR}/PK.auth')
             for key in ['KEK', 'db', 'dbx']:
                 assert not h.file_exists(f'{CERT_DIR}/{key}.auth')
--- a/tests/uefistored/test_cert_inheritance.py
+++ b/tests/uefistored/test_cert_inheritance.py
@@ -169,7 +169,8 @@ class TestPoolToDiskCertInheritanceAtXapiStart:
         install_certs_to_disks(host.pool, disk_auths, ['PK', 'KEK', 'db', 'dbx'])
         host.restart_toolstack(verify=True)
         logging.info('Check that the certs on disk have been erased since there is none in the pool.')
-        for key in ['PK', 'KEK', 'db', 'dbx']:
+        assert host.file_exists(f'{CERT_DIR}/PK.auth')
+        for key in ['KEK', 'db', 'dbx']:
             assert not host.file_exists(f'{CERT_DIR}/{key}.auth')
 
     def test_pool_certs_present_and_some_different_disk_certs_present(self, host):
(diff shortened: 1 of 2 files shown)


tests/uefistored/test_cert_inheritance.py
```diff
@@ -325,15 +325,18 @@ def test_set_pool_certificates_partial(self, host):
...
 
-    def test_clear_certificates_from_pool(self, host):
-        keys = ['PK', 'KEK', 'db', 'dbx']
-        pool_auths = generate_keys(as_dict=True)
-        host.pool.install_custom_uefi_certs([pool_auths[key] for key in keys])
-        host.pool.clear_uefi_certs()
+    @pytest.mark.small_vm
+    def test_clear_certificates_from_pool(self, host, uefi_vm_and_snapshot):
+        vm, snapshot = uefi_vm_and_snapshot
+        revert_vm_state(vm, snapshot)
```
stormi, line 331: Should not be necessary. Other tests already revert the VM state in their teardown. That's the contract.

benjamreis: not in this class, as it's the only test using a vm in this class.

stormi: What I mean is:
* either you're executing just this class and the VM is already clean
* or you executed other tests before but they left the VM clean

Conclusion: the VM is already clean of certificates.

tests/uefistored/test_cert_inheritance.py
```diff
@@ -325,15 +326,17 @@ def test_set_pool_certificates_partial(self, host):
...
             assert not h.file_exists(f'{CERT_DIR}/{missing_key}.auth')
 
-    def test_clear_certificates_from_pool(self, host):
-        keys = ['PK', 'KEK', 'db', 'dbx']
-        pool_auths = generate_keys(as_dict=True)
-        host.pool.install_custom_uefi_certs([pool_auths[key] for key in keys])
-        host.pool.clear_uefi_certs()
+    @pytest.mark.small_vm
+    def test_clear_certificates_from_pool(self, host, uefi_vm):
+        vm = uefi_vm
```
stormi, line 331: Now your test, when executed alone, won't necessarily have a VM clean of certificates. It's the `uefi_vm_and_snapshot` fixture that cleans them in its setup, not the `uefi_vm` fixture.

Plus, using `uefi_vm_and_snapshot` would let you revert the VM at the end of this test, for other tests that may be executed afterwards (pytest's execution order is not always that of the file, or we may add more tests in the future) and that use the same VM.

tests/uefistored/test_cert_inheritance.py
```diff
@@ -325,15 +326,18 @@ def test_set_pool_certificates_partial(self, host):
...
-        host.pool.install_custom_uefi_certs([pool_auths[key] for key in keys])
-        host.pool.clear_uefi_certs()
+    @pytest.mark.small_vm
+    def test_clear_certificates_from_pool(self, host, uefi_vm_and_snapshot):
+        vm, snapshot = uefi_vm_and_snapshot
         for h in host.pool.hosts:
-            logging.info(f"Check host {h} has no certificate on disk.")
-            for key in keys:
+            logging.info(f"Check host {h} has no certificate on disk except PK.")
+            # Make sure PK has been fetched from fallback dir so none SB UEFI VMs can still be booted
```
stormi, line 334:
```suggestion
            # Make sure PK has been fetched from fallback dir so UEFI VMs without certs can still be booted
```

stormi (commented):
Looks good, now we need to wait for the upstream PR.

stormi:
I think we can close this PR. Certificate handling has changed a lot since it was created and we can't clear certificates: only custom ones. There's TestPoolToDiskCertPropagationToAllHosts::test_clear_custom_pool_certificates which checks that the symlink is back in place after a clear.

PR #111, opened 2022-05-10
Fix VDI migration on the right SR

Wescoeur:


gduperrey (approved):

lib/basevm.py
```diff
@@ -63,6 +63,12 @@ def all_vdis_on_host(self, host):
                 return False
         return True
 
+    def all_vdis_on_sr(self, sr):
```
stormi, line 66: Do we still need the all_vdis_on_host method? If yes, should we make it return False for shared SRs?

tests/migration/test_host_evacuate.py
```diff
@@ -20,8 +20,7 @@
 #   This network will be disconnected at some point during the tests
 
 def _host_evacuate_test(source_host, dest_host, network_uuid, vm, expect_error=False, error=""):
-    source_name = source_host.xe('host-param-get', {'uuid': source_host.uuid, 'param-name': 'name-label'})
-    vm.start(on=source_name)
+    vm.start(on=source_host.uuid)
```
stormi, line 23: IIRC there's also a param that must be removed from above the vm.start method definition.

tests/storage/moosefs/conftest.py
```diff
@@ -13,15 +14,24 @@ def host_with_moosefs(host):
...
+def pool_with_moosefs(hostA1):
+    pool = Pool(hostA1.hostname_or_ip)
+    for host in pool.hosts:
+        enable_moosefs(host)
+
+    yield pool
+
+    for host in pool.hosts:
+        disable_moosefs(host)
+
```
stormi, line 34: This fixture will behave poorly if for example the third host has moosefs installed on it for some reason. But it's not worse than in other similar tests, so I think I'll try to fix them all at some point.

stormi: test_reboot should also probably explicitly start the VM on the master host as this is the only host that was rebooted.

stormi (approved):

PR #129, opened 2022-06-27
Fix option accumulation in _ssh()

stormi:
Commit 0bec9b87a4d6df4c7f6dc9d13a7dd82e67dbf7e9 introduced a regression:
as the default value for parameter `options` is mutable, only one
reference is maintained by python, and so every call to _ssh() adds the
same parameters to the list, again and again, until the following
exception is raised: "OSError: [Errno 7] Argument list too long"

Use a copy of the parameter in the function to fix this.


stormi:
I'm waiting for the previous job to end before I can test this.

gduperrey (approved):

lib/commands.py
```diff
@@ -149,19 +150,19 @@ def ssh(hostname_or_ip, cmd, check=True, simple_output=True, suppress_fingerprin
     return result_or_exc
 
 def scp(hostname_or_ip, src, dest, check=True, suppress_fingerprint_warnings=True, local_dest=False):
-    options = '-o "BatchMode yes"'
+    opts = '-o "BatchMode yes"'
```
Wescoeur, line 153: Not directly related to the commit message, no? :)

stormi: Indeed. Commit message now updated.

stormi:
> I'm waiting for the previous job to end before I can test this.

Test OK.

benjamreis (approved):

PR #131, opened 2022-06-30
install_xcpng.py: add the restore option to restore upgraded xcpng

gduperrey:

### DIFF
--- a/README.md
+++ b/README.md
@@ -373,7 +373,7 @@ in data.py.
 
 ## install_xcpng.py
 
-This script installs (or upgrades) XCP-ng in a VM using a PXE server whose configuration can be defined dynamically. Basically, it writes files in a directory named after the MAC address of the VM, on a PXE server that will then build a boot configuration for the given MAC address. This is rather specific to Vates' test lab at the moment. If you are interested in automated installation in general, check https://xcp-ng.org/docs/install.html#automated-install.
+This script installs, upgrades or restores XCP-ng in a VM using a PXE server whose configuration can be defined dynamically. Basically, it writes files in a directory named after the MAC address of the VM, on a PXE server that will then build a boot configuration for the given MAC address. This is rather specific to Vates' test lab at the moment. If you are interested in automated installation in general, check https://xcp-ng.org/docs/install.html#automated-install.
 
 ```
 usage: install_xcpng.py [-h] [--installer INSTALLER] host vm_uuid action xcpng_version
@@ -388,6 +388,12 @@ Upgrade example:
 ```
 python scripts/test_install_xcpng.py 10.0.0.2 f0f5f010-80c6-25ae-44a2-1fb154e32d14 upgrade 8.2.1
 ```
+
+Restore example:
+```
+python scripts/test_install_xcpng.py 10.0.0.2 f0f5f010-80c6-25ae-44a2-1fb154e32d14 restore 8.2.1
+```
+Note: in case of restore, the version must be that of the installer (here 8.2.1), not the version of XCP-ng that will be restored.
 
 The script requires the addressable name or IP of the PXE config server to be defined in `data.py`:
 ```
--- a/scripts/install_xcpng.py
+++ b/scripts/install_xcpng.py
@@ -30,21 +30,25 @@ def pxe_address():
     except ImportError:
         raise Exception('No address for the PXE server found in data.py (`PXE_CONFIG_SERVER`)')
 
-def generate_boot_conf(directory, installer):
+def generate_boot_conf(directory, installer, action):
+    # in case of restore, we disable the text ui from the installer completely,
+    # to workaround a bug that leaves us stuck on a confirmation dialog at the end of the operation.
+    rt = 'rt=1' if action == 'restore' else ''
     with open(f'{directory}/boot.conf', 'w') as bootfile:
         bootfile.write(f"""answerfile=custom
-    installer={installer}
-    is_default=1
-    """)
+installer={installer}
+is_default=1
+{rt}
+""")
 
-def generate_answerfile(directory, installer, hostname_or_ip, type, hdd):
+def generate_answerfile(directory, installer, hostname_or_ip, action, hdd):
     pxe = pxe_address()
     password = host_data(hostname_or_ip)['password']
     cmd = ['openssl', 'passwd', '-6', password]
     res = subprocess.run(cmd, stdout=subprocess.PIPE)
     encrypted_password = res.stdout.decode().strip()
     with open(f'{directory}/answerfile.xml', 'w') as answerfile:
-        if type == 'install':
+        if action == 'install':
             answerfile.write(f"""<?xml version="1.0"?>
 <installation>
     <keymap>fr</keymap>
@@ -59,7 +63,7 @@ def generate_answerfile(directory, installer, hostname_or_ip, type, hdd):
     </script>
 </installation>
         """)
-        elif type == 'upgrade':
+        elif action == 'upgrade':
             answerfile.write(f"""<?xml version="1.0"?>
 <installation mode="upgrade">
     <existing-installation>{hdd}</existing-installation>
@@ -69,8 +73,13 @@ def generate_answerfile(directory, installer, hostname_or_ip, type, hdd):
     </script>
 </installation>
         """)
+        elif action == 'restore':
+            answerfile.write(f"""<?xml version="1.0"?>
+<restore>
+</restore>
+        """)
         else:
-            raise Exception(f"Unknown type `{type}`")
+            raise Exception(f"Unknown action: `{action}`")
 
 def copy_files_to_pxe(mac_address, tmp_local_path):
     assert mac_address
@@ -152,14 +161,17 @@ def main():
     parser = argparse.ArgumentParser()
     parser.add_argument(
         "host",
-        help="hostname or IP address of the host hosting the VM that will be installed/upgraded"
+        help="hostname or IP address of the host hosting the VM that will be installed/upgraded/restored"
     )
-    parser.add_argument("vm_uuid", help="UUID of an existing VM in which XCP-ng will be installed (or upgraded)")
-    parser.add_argument("action", metavar='action', choices=['install', 'upgrade'], help="install or upgrade")
+    parser.add_argument("vm_uuid", help="UUID of an existing VM in which XCP-ng will be installed/upgraded/restored")
+    parser.add_argument(
+        "action", metavar='action', choices=['install', 'upgrade', 'restore'], help="install, upgrade or restore"
+    )
     parser.add_argument(
         "xcpng_version",
         help="target version, used to build the installer URL if none provided via --installer, "
              "and also used to check the system version at the end. Example: 8.2.1"
+             "In case of a restore, specify the version of the installer."
     )
     parser.add_argument("--installer", help="URL of the installer")
     args = parser.parse_args()
@@ -206,7 +218,7 @@ def main():
         logging.info('Generate files: answerfile.xml and boot.conf')
         hdd = 'nvme0n1' if vm.is_uefi else 'sda'
         generate_answerfile(tmp_local_path, installer, args.host, args.action, hdd)
-        generate_boot_conf(tmp_local_path, installer)
+        generate_boot_conf(tmp_local_path, installer, args.action)
         logging.info('Copy files to the pxe server')
         copy_files_to_pxe(mac_address, tmp_local_path)
         atexit.register(lambda: clean_files_on_pxe(mac_address))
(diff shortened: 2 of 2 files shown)


scripts/install_xcpng.py
```diff
@@ -62,24 +62,58 @@ def generate_answerfile(directory, installer, hostname_or_ip, type, hdd):
         elif type == 'upgrade':
             answerfile.write(f"""<?xml version="1.0"?>
 <installation mode="upgrade">
-    <existing-installation>sda</existing-installation>
+    <existing-installation>{hdd}</existing-installation>
```
stormi, line 65: This change is not documented in the commit message and may even deserve a separate bugfix commit.

scripts/install_xcpng.py
```diff
@@ -62,24 +62,58 @@ def generate_answerfile(directory, installer, hostname_or_ip, type, hdd):
...
     <source type="url">{installer}</source>
     <script stage="filesystem-populated" type="url">
         http://{pxe}/configs/presets/scripts/filesystem-populated.py
     </script>
 </installation>
         """)
+        elif type == 'restore':
+            answerfile.write(f"""<?xml version="1.0"?>
+<restore>
+    <existing-installation>{hdd}</existing-installation>
```
stormi, line 75: Does this line do anything concretely? It's specific to the `upgrade` operation so I doubt it. I suppose it's ignored and the restore process auto-detects the existing installation.

scripts/install_xcpng.py
```diff
@@ -62,24 +62,58 @@ def generate_answerfile(directory, installer, hostname_or_ip, type, hdd):
...
+ONTIMEOUT custom
+
+LABEL custom
+    MENU LABEL Custom XCP-ng installation
+    KERNEL mboot.c32
+    APPEND {installer}/boot/xen.gz dom0_max_vcpus=1-16 dom0_mem=max:8192M com1=115200,8n1 console=com1,vga \
+--- {installer}/boot/vmlinuz xencons=hvc console=hvc0 console=tty0 \
+rt_answerfile=http://{pxe}/configs/custom/{mac_address}/answerfile.xml install \
+--- {installer}install.img"""
+            )
```
stormi, line 103: As I told you earlier, using `pxelinux.cfg` was to be done manually, as a way to test the `rt_answerfile` option before. It was not supposed to end up in this script. I also said after the manual tests we would improve `bootconfig.php` so that it would support a `rt_answerfile` option in `boot.conf`.

By the way pxelinux is not used in UEFI mode so the script probably doesn't work with UEFI XCP-ng.

In the future, maybe we'll add support for writing `pxelinux.cfg` and `grub.cfg` files in this script (so let's save this snipped somewhere), but for now `boot.conf` is the way.

scripts/install_xcpng.py
```diff
@@ -232,9 +282,16 @@ def main():
...
+            else:
+                raise Exception(
+                    f"The installed host ({vm_ip_address}) is not in the expected version. Got: {host2.xcp_version}.\n"
+                    f"Expected: {xcp_version}."
+                )
+        elif host2.xcp_version == version.parse(xcp_version) and args.action == 'restore':
             raise Exception(
-                f"The installed host ({vm_ip_address}) is not in the expected version. Got: {host2.xcp_version}.\n"
-                f"Expected: {xcp_version}."""
+                f"The installed host ({vm_ip_address}) is not in a previous version. Got: {host2.xcp_version}.\n"
```
stormi, line 294: It would be better to first test the action and then test the version accordingly:

```
if args.action == 'restore':
    # check that the new version is lower than xcp_version
else
    # check that the version is exactly the expected version
```

benjamreis: Plus instead of imbricating `if` blocks. an assert would do the job here:
```py
if args.action == 'restore':
    assert(new_version < current_version, <error msg to log>)
else
    assert(new_version == current_version, <other error msg to log>)

**Review by benjamreis — CHANGES_REQUESTED** [review]

File: scripts/install_xcpng.py (hunk 4 above)
Comment (stormi) [reply scripts/install_xcpng.py:294]: We're not in the code of a test so semantically speaking I wouldn't use assert to check the result of the operation, even if the result would be globally the same.

Comment (Wescoeur) [reply scripts/install_xcpng.py:294]: Well, an assertion is also used to be sure an algorithm or function is correctly implemented. It's not reserved for testing. ;)

In the present case, I suppose we should use an exception if we can trigger this problem using a "bad" XCP-ng.

Comment (stormi) [reply scripts/install_xcpng.py:294]: Yes, the error would not be related to the implementation of the script, so Exception it is.

File: scripts/install_xcpng.py [hunk 5]
```diff
@@ -62,24 +62,58 @@ def generate_answerfile(directory, installer, hostname_or_ip, type, hdd):
...
+    APPEND {installer}/boot/xen.gz dom0_max_vcpus=1-16 dom0_mem=max:8192M com1=115200,8n1 console=com1,vga \
+--- {installer}/boot/vmlinuz xencons=hvc console=hvc0 console=tty0 \
+rt_answerfile=http://{pxe}/configs/custom/{mac_address}/answerfile.xml install \
+--- {installer}install.img"""
+            )
         else:
             raise Exception(f"Unknown type `{type}`")
 
-def copy_files_to_pxe(mac_address, tmp_local_path):
+def copy_files_to_pxe(mac_address, tmp_local_path, type):
```
Wescoeur, line 107: Maybe `action` instead of `type` to be symetrical with the caller?

scripts/install_xcpng.py
```diff
@@ -62,24 +62,58 @@ def generate_answerfile(directory, installer, hostname_or_ip, type, hdd):
...
+    <script stage="filesystem-populated" type="url">
+        http://{pxe}/configs/presets/scripts/filesystem-populated.py
+    </script>
+</restore>
+        """)
+        else:
+            raise Exception(f"Unknown type `{type}`")
+
+def generate_pxelinux(directory, installer, type, mac_address):
+    pxe = pxe_address()
```
Wescoeur, line 86: The variable is not used.

Wescoeur (approved):

scripts/install_xcpng.py
```diff
@@ -155,7 +167,9 @@ def main():
         help="hostname or IP address of the host hosting the VM that will be installed/upgraded"
     )
     parser.add_argument("vm_uuid", help="UUID of an existing VM in which XCP-ng will be installed (or upgraded)")
-    parser.add_argument("action", metavar='action', choices=['install', 'upgrade'], help="install or upgrade")
+    parser.add_argument(
+        "action", metavar='action', choices=['install', 'upgrade', 'restore'], help="install, upgrade or restore"
+    )
     parser.add_argument(
         "xcpng_version",
         help="target version, used to build the installer URL if none provided via --installer, "
```
stormi, line 175: The help message needs to be completed with what the version means in the case of a restore.

scripts/install_xcpng.py
```diff
@@ -217,7 +231,7 @@ def main():
                 vm.shutdown(force=True, verify=True)
         vm.start()
         # wait a bit to let the PXE server give the boot configuration to the VM, then disable the specific boot config
-        time.sleep(20)
+        time.sleep(60)
```
stormi, line 234: Not documented in the commit message

scripts/install_xcpng.py
```diff
@@ -230,11 +244,17 @@ def main():
         host2 = pool2.master
         host2.inventory = host2._get_xensource_inventory()
         check_mac_address(host2, mac_address)
-        logging.info('New host is started and enabled')
-        if host2.xcp_version != version.parse(xcp_version):
+        logging.info(f'New host is started and enabled in version:{host2.xcp_version}')
+        if args.action == 'restore' and host2.xcp_version < version.parse(xcp_version):
+            logging.info(f'The host {vm_ip_address} is in version : {host2.xcp_version}')
```
stormi, line 249: This doesn't seem useful to me. This repeats what the previous line says already.

scripts/install_xcpng.py
```diff
@@ -230,11 +244,17 @@ def main():
         host2 = pool2.master
         host2.inventory = host2._get_xensource_inventory()
         check_mac_address(host2, mac_address)
-        logging.info('New host is started and enabled')
-        if host2.xcp_version != version.parse(xcp_version):
+        logging.info(f'New host is started and enabled in version:{host2.xcp_version}')
```
stormi, line 247: Looks like I missed an occurrence of "new host" in the previous PR. It's not always "new". Replacing with "target".

```suggestion
        logging.info(f'Target host is started and enabled in version: {host2.xcp_version}')
```

stormi (changes requested):

scripts/install_xcpng.py
```diff
@@ -30,21 +30,26 @@ def pxe_address():
     except ImportError:
         raise Exception('No address for the PXE server found in data.py (`PXE_CONFIG_SERVER`)')
 
-def generate_boot_conf(directory, installer):
+def generate_boot_conf(directory, installer, action):
+    if action == 'restore':
+        rt = 'rt=1'
+    else:
+        rt = ''
```
benjamreis, line 37:
```suggestion
    rt = 'rt=1' if action == 'restore' else ''
```
It's not blocking for the merge but it is more concise and more inline with the python standard coding style.

scripts/install_xcpng.py
```diff
@@ -230,11 +243,15 @@ def main():
...
         host2.inventory = host2._get_xensource_inventory()
         check_mac_address(host2, mac_address)
-        logging.info('New host is started and enabled')
-        if host2.xcp_version != version.parse(xcp_version):
+        logging.info(f'Target host is started and enabled in version:{host2.xcp_version}')
+        if args.action == 'restore' and host2.xcp_version == version.parse(xcp_version):
+            raise Exception(
+                f"The installed host ({vm_ip_address}) is not in a previous version. Got: {host2.xcp_version}.\n"
+            )
+        elif args.action != 'restore' and host2.xcp_version != version.parse(xcp_version):
```
benjamreis, line 251: I'm not sure about this condition in the case `args.action` is `upgrade` but I might misunderstand.

scripts/install_xcpng.py
```diff
@@ -230,11 +240,15 @@ def main():
         host2 = pool2.master
         host2.inventory = host2._get_xensource_inventory()
         check_mac_address(host2, mac_address)
-        logging.info('New host is started and enabled')
-        if host2.xcp_version != version.parse(xcp_version):
+        logging.info(f'Target host is started and enabled in version:{host2.xcp_version}')
+        if args.action == 'restore' and not host2.xcp_version < version.parse(xcp_version):
```
benjamreis, line 244:
```suggestion
        if args.action == 'restore' and host2.xcp_version >= version.parse(xcp_version):
```
Simpler & clearer :)

stormi (commented):
In addition to inline comments, two comments:

* README.txt needs to be updated
* The title of the first commit message is not easily understandable. So I suggest replacing "Correct value hdd for upgrade for a dynamic one" with "Fix HDD value in answerfile for upgrade actions"

And the message body:
```diff
- The value for hdd on upgrade was sda. It could just
work for bios vm. We need a dynamic value so it work
for bios and uefi
- The value for `existing-installation` in the answerfile on upgrade was `sda`. It would only
work for bios VMs. We need a diff dynamic value so it works for bios and uefi.
```

scripts/install_xcpng.py
```diff
@@ -155,11 +162,14 @@ def main():
         help="hostname or IP address of the host hosting the VM that will be installed/upgraded"
```
stormi, line 162:
```suggestion
        help="hostname or IP address of the host hosting the VM that will be installed/upgraded/restored"
```

scripts/install_xcpng.py
```diff
@@ -155,11 +162,14 @@ def main():
...
     )
     parser.add_argument("vm_uuid", help="UUID of an existing VM in which XCP-ng will be installed (or upgraded)")
```
stormi, line 164:
```suggestion
    parser.add_argument("vm_uuid", help="UUID of an existing VM in which XCP-ng will be installed/upgraded/restored")
```

stormi (approved):

README.md
```diff
@@ -389,6 +389,11 @@ Upgrade example:
 python scripts/test_install_xcpng.py 10.0.0.2 f0f5f010-80c6-25ae-44a2-1fb154e32d14 upgrade 8.2.1
 ```
 
+Restore example:
+```
+python scripts/test_install_xcpng.py 10.0.0.2 f0f5f010-80c6-25ae-44a2-1fb154e32d14 restore 8.2.1
+```
+
```
stormi, line 396:
```suggestion
Restore example:
```
python scripts/test_install_xcpng.py 10.0.0.2 f0f5f010-80c6-25ae-44a2-1fb154e32d14 restore 8.2.1
```
Note: in case of restore, the version must be that of the installer (here 8.2.1), not the version of XCP-ng that will be restored.

```

Wescoeur (approved):

PR #132, opened 2022-07-05
Test `is_open` call (on-slave plugin) for many SRs

Wescoeur:

### DIFF
--- a/lib/sr.py
+++ b/lib/sr.py
@@ -2,7 +2,8 @@ import logging
 
 import lib.commands as commands
 
-from lib.common import safe_split, wait_for, wait_for_not
+from lib.common import prefix_object_name, safe_split, wait_for, wait_for_not
+from lib.vdi import VDI
 
 class SR:
     def __init__(self, uuid, pool):
@@ -137,3 +138,11 @@ class SR:
         if self._is_shared is None:
             self._is_shared = self.pool.master.xe('sr-param-get', {'uuid': self.uuid, 'param-name': 'shared'})
         return self._is_shared
+
+    def create_vdi(self, name_label, virtual_size=64):
+        vdi_uuid = self.pool.master.xe('vdi-create', {
+            'name-label': prefix_object_name(name_label),
+            'virtual-size': str(virtual_size),
+            'sr-uuid': self.uuid
+        })
+        return VDI(self, vdi_uuid)
--- a/tests/storage/storage.py
+++ b/tests/storage/storage.py
@@ -1,5 +1,6 @@
 from lib.commands import SSHCommandFailed
-from lib.common import wait_for
+from lib.common import strtobool, wait_for
+from lib.sr import SR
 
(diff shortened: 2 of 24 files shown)


tests/storage/storage.py
```diff
@@ -43,3 +44,31 @@ def live_storage_migration_then_come_back(vm, prov_host, prov_sr, dest_host, des
     wait_for(lambda: vm.all_vdis_on_sr(prov_sr), "Wait for all VDIs back on provenance SR")
     wait_for(lambda: vm.is_running_on_host(prov_host), "Wait for VM to be running on provenance host")
     vm.shutdown(verify=True)
+
+def vdi_is_open(vdi):
```
stormi, line 48: Should be named with an action as it does things.

For example this?
```suggestion
def call_vdi_is_open(vdi):
```

tests/storage/xfs/test_xfs_sr.py
```diff
@@ -39,6 +40,10 @@ def test_create_and_destroy_sr(self, sr_disk, host_with_xfsprogs):
         vm.destroy(verify=True)
         sr.destroy(verify=True)
 
+@pytest.mark.simple_vdi
+def test_xfs_sr_vdi_is_open(vdi_on_xfs_sr):
+    vdi_is_open(vdi_on_xfs_sr)
+
```
stormi, line 46: Why not make it a method of `TestXFSSR`?

tests/storage/storage.py
Wescoeur: I modified the function, now a boolean is returned. So I'm not sure we should rename it.

tests/storage/xfs/test_xfs_sr.py
Wescoeur: Many helpers (Test*SR) use a vm fixture (useless for is_open) like here:
```python
@pytest.mark.usefixtures("nfs_sr", "vm_on_nfs_sr")
class TestNFSSR:
```

I prefer to keep this new test outside these classes.

lib/sr.py
```diff
@@ -130,10 +131,19 @@ def main_host(self):
...
         return self.pool.master.xe('sr-param-get', {'uuid': self.uuid, 'param-name': 'content-type'})
 
     def is_shared(self):
         if self._is_shared is None:
             self._is_shared = self.pool.master.xe('sr-param-get', {'uuid': self.uuid, 'param-name': 'shared'})
         return self._is_shared
+
+    def create_vdi(self, name_label, virtual_size=64):
+        vdi_uuid = self.pool.master.xe('vdi-create', {
+            'name-label': name_label, 'virtual-size': str(virtual_size), 'sr-uuid': self.uuid
```
stormi, line 147: It could be good to add the name prefix we already use for various other objects to the name-label.

gduperrey (approved):

stormi:
We'll wait for the corresponding fix to be released to users before merging this.

stormi (approved):

PR #133, opened 2022-07-27
Execute quicktest for each SM driver

Wescoeur:


gduperrey (approved):

tests/storage/storage.py
```diff
@@ -70,3 +70,13 @@ def get_xapi_session():
         'vdiUuid': vdi.uuid,
         'srRef': master.execute_script(get_sr_ref, shebang='python')
     }))
+
+def run_quicktest_on_sr(sr):
```
stormi, line 74: Maybe this could be a method of SR

stormi:
The latest update candidate for xapi-tests will fix the behaviour of the `-default-sr` parameter of `quicktest`. You can install it on the hosts to test your PR.

tests/storage/storage.py
gthvn1: I confirm that I run it with
```
sr.pool.master.ssh(['/opt/xensource/debug/quicktest', '-sr', sr.uuid])
```
So no need to have `get_default_sr` and `set_default_sr` anymore (at least when `-sr` will be available)

tests/storage/fsp/test_fsp_sr.py
```diff
@@ -3,6 +3,7 @@
 import logging
 
 from .conftest import DIRECTORIES_PATH
+from tests.storage import run_quicktest_on_sr
```
benjamreis, line 6: No longer needed.

benjamreis (changes requested):

stormi:
This PR has had a force push last month, but it's still named "[WIP]", and no re-review was requested. What's the current status?

benjamreis:
IIRC we were waiting for new sources from XAPI to include the quicktest `--sr` option and from smapi to have the proper test filter applied.

stormi:
Maybe I'm wrong, since you are the ones leading this development, but in my view I was waiting for you to tell me: this PR is good, you can review it.

gthvn1:
Ooops sorry yes I back ported the patch to allow user to select the SR using the option. 
As this patch is the work of @Wescoeur I will let him modify the title and remove the WIP and check if everything is OK but for me it is ready for reviewing.

tests/storage/fsp/test_fsp_sr.py
benjamreis: Still relevant

benjamreis (changes requested):
also do we remove the `tests/quicktest` since now it is launched for each driver?

Wescoeur:
>  As this patch is the work of @Wescoeur I will let him modify the title and remove the WIP and check if everything is OK but for me it is ready for reviewing.

Because I don't work on that, you can freely modify whatever you want. :)

benjamreis (approved):

tests/storage/storage.py
benjamreis: IMHO this should stay as test method as it's only used in the `storage` scope.

stormi: A method this simple and whose main input is a SR UUID, I would have put it on the SR class anyway (whose scope is storage anyway), and avoided a lot of imports in test files.

stormi (approved):
Next step, before merging: run the tests several times in a row to try to ensure they are stable.

stormi:
Rebased on master branch

gthvn1 (approved):

PR #134, opened 2022-09-15
Do not attempt to remove same elem several times

benjamreis:
In case of a pool with more than 2 hosts,
a same disk can be tried to be removed from `candidates` more than once.
This results in `list.remove` throwing a `ValueError`.

Instead of `master_disks`, iterate over `candidates` to only attempt to remove elem in it.


conftest.py
```diff
@@ -209,7 +209,7 @@ def sr_disk_for_all_hosts(request, host):
     candidates = list(master_disks)
     for h in host.pool.hosts[1:]:
         other_disks = h.available_disks()
-        for d in master_disks:
+        for d in [x for x in candidates]:
             if d not in other_disks:
                 candidates.remove(d)
                 continue
```
Wescoeur, line 215:
```python
   for h in host.pool.hosts[1:]:
        other_disks = h.available_disks()
        new_candidates = []
        for d in candidates:
            if d not in other_disks:
                continue
            new_candidates.append[d]
        candidates = new_candidates
```

stormi (approved):

gduperrey (approved):

Wescoeur (approved):

PR #136, opened 2022-11-03
Local ISO SR tests: start the VM on the same host as the SR

gduperrey:
As the imported VM needs to be on the host with the local SR, we specify the host where the VM must be running.

### DIFF
--- a/tests/storage/iso/test_local_iso_sr.py
+++ b/tests/storage/iso/test_local_iso_sr.py
@@ -6,7 +6,7 @@ from .conftest import create_local_iso_sr, copy_tools_iso_to_iso_sr, \
 
 # Requirements:
 # From --hosts parameter:
-# - host: a XCP-ng host
+# - host: a XCP-ng host, with the default SR being either a shared SR, or a local SR on the master host
 # From --sr-disk parameter:
 # - an additional unused disk for the SR
 # From --vm parameter:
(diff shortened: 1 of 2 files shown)


tests/storage/iso/test_local_iso_sr.py
```diff
@@ -5,6 +5,7 @@
     check_iso_mount_and_read_from_vm, remove_iso_from_sr
 
 # Requirements:
+# A shared SR or a default SR on the master
 # From --hosts parameter:
 # - host: a XCP-ng host
```
stormi, line 10: It's a bit awkward to mention the shared SR before even mentioning the host/pool.

```suggestion
# From --hosts parameter:
# - host: a XCP-ng host, with the default SR being either a shared SR, or a local SR on the master host
```

tests/storage/iso/test_local_iso_sr.py
```diff
@@ -42,11 +43,13 @@ def test_create_and_destroy_sr(self, host, formatted_and_mounted_ext4_disk):
 @pytest.mark.usefixtures("local_iso_sr")
 class TestLocalISOSR:
 
-    def test_iso_mount_and_read(self, host, local_iso_sr, running_unix_vm):
+    def test_iso_mount_and_read(self, host, local_iso_sr, imported_vm):
```
stormi, line 46: The `unix_vm` fixture would be more appropriate to replace the `running_unix_vm` fixture. Else we lose an important part of the fixture.

tests/storage/iso/test_local_iso_sr_reboot.py
```diff
@@ -4,6 +4,7 @@
 from .conftest import copy_tools_iso_to_iso_sr, check_iso_mount_and_read_from_vm, remove_iso_from_sr
 
 # Requirements:
+# A shared SR or a default SR on the master
```
stormi, line 7: Same comment as above.

stormi (changes requested):

benjamreis (approved):
LGTM after required changes

stormi:
The commit title lacks context. Change to "Local ISO SR tests: start the VM on the same host as the SR"?

stormi (approved):

PR #140, opened 2022-11-30
Repair fsp tests: Install xcp-ng-xapi-storage

Wescoeur:

### DIFF
--- a/lib/host.py
+++ b/lib/host.py
@@ -90,6 +90,29 @@ class Host:
     def param_get(self, param_name, key=None, accept_unknown_key=False):
         return _param_get(self, self.xe_prefix, self.uuid, param_name, key, accept_unknown_key)
 
+    def create_file(self, filename, text):
+        with tempfile.NamedTemporaryFile('w') as file:
+            file.write(text)
+            file.flush()
+            self.scp(file.name, filename)
+
+    def add_xcpng_repo(self, name):
+        major = self.xcp_version.major
+        minor = self.xcp_version.minor
+        version = f"{major}.{minor}"
+        self.create_file(f"/etc/yum.repos.d/xcp-ng-{name}.repo", (
+            f"[xcp-ng-{name}]\n"
+            f"name=XCP-ng {name} Repository\n"
+            f"baseurl=http://mirrors.xcp-ng.org/{major}/{version}/{name}/x86_64/ http://updates.xcp-ng.org/{major}/{version}/{name}/x86_64/\n" # noqa
+            "enabled=1\n"
+            "gpgcheck=1\n"
+            "repo_gpgcheck=1\n"
+            "gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-xcpng\n"
+        ))
+
+    def remove_xcpng_repo(self, name):
+        self.ssh(['rm -f /etc/yum.repos.d/xcp-ng-{}.repo'.format(name)])
+
     def execute_script(self, script_contents, shebang='sh', simple_output=True):
         with tempfile.NamedTemporaryFile('w') as script:
             os.chmod(script.name, 0o775)
--- a/tests/storage/fsp/conftest.py
+++ b/tests/storage/fsp/conftest.py
@@ -1,24 +1,43 @@
 import pytest
 
+FSP_REPO_NAME = 'runx'
+
+FSP_PACKAGES = ['xcp-ng-xapi-storage']
+
 DIRECTORIES_PATH = 'directories'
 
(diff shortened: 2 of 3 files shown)


tests/storage/fsp/conftest.py
```diff
@@ -1,24 +1,55 @@
 import pytest
 
+FSP_REPO_FILE = '/etc/yum.repos.d/xcp-ng-runx.repo'
+
+FSP_REPO_TEXT = """\
+[xcp-ng-runx]
+name=XCP-ng runx Repository
+baseurl=http://mirrors.xcp-ng.org/8/8.2/runx/x86_64/ http://updates.xcp-ng.org/8/8.2/runx/x86_64/
```
stormi, line 8: The URL should depend on the release of XCP-ng of the host being tested.

tests/storage/fsp/conftest.py
```diff
@@ -1,24 +1,55 @@
...
+FSP_PACKAGES = ['xcp-ng-xapi-storage']
+
 DIRECTORIES_PATH = 'directories'
 
+def install_fsp(host):
+    assert not host.file_exists('/usr/bin/runx'), \
+        "runx should not be installed on the host before all tests"
+    host.create_file(FSP_REPO_FILE, FSP_REPO_TEXT)
+    host.yum_save_state()
+    host.yum_install(FSP_PACKAGES, enablerepo='epel')
```
stormi, line 24: 1. Enabling epel is probably not necessary if we don't install runx.
2. this install will work, provided the release of the xcp-ng-xapi-storage in the runx repo is always higher than that in xcp-ng-base and xcp-ng-updates repos. If you want to ensure this works (provided we don't install xcp-ng-xapi-storage on hosts by default someday), you should disable the other repos before installing.

tests/storage/fsp/test_fsp_sr.py
```diff
@@ -36,17 +36,17 @@ def test_create_and_destroy_sr_non_existing_path(self, host):
 
 @pytest.mark.usefixtures("host_at_least_8_3")
```
stormi, line 37: We can remove this restriction now that the runx repo is added to make it work.

tests/storage/fsp/conftest.py
```diff
@@ -1,24 +1,55 @@
...
+"""
+
+FSP_PACKAGES = ['xcp-ng-xapi-storage']
+
 DIRECTORIES_PATH = 'directories'
 
+def install_fsp(host):
+    assert not host.file_exists('/usr/bin/runx'), \
+        "runx should not be installed on the host before all tests"
+    host.create_file(FSP_REPO_FILE, FSP_REPO_TEXT)
```
stormi, line 22: In addition to `create_file`, a Host method to add an optional xcp-ng yum repository (runx, linstor, ...) would be nice.

Example signature :

```
    def add_xcpng_repo(name):
        # creates the repo file for repo `name` from XCP-ng mirror
```

lib/host.py
```diff
@@ -85,6 +85,31 @@ def stringify(key, value):
...
+        with tempfile.NamedTemporaryFile('w') as file:
+            file.write(text)
+            file.flush()
+            self.scp(file.name, filename)
+
+    def add_xcpng_repo(self, name):
+        content = """\
+[xcp-ng-{}]
+name=XCP-ng {} Repository
+baseurl=http://mirrors.xcp-ng.org/8/{}/{}/x86_64/ http://updates.xcp-ng.org/8/{}/{}/x86_64/
```
stormi, line 98: It would not be a lot more work to replace "8" with the first number from the release.

lib/host.py
```diff
@@ -85,6 +85,32 @@ def stringify(key, value):
...
+gpgcheck=1
+repo_gpgcheck=1
+gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-xcpng
+"""
+
+        major = self.xcp_version.major
+        self.create_file(
+            '/etc/yum.repos.d/xcp-ng-{}.repo'.format(name),
+            content.format(name, name, major, self.xcp_version, name, major, self.xcp_version, name)
+        )
```
stormi, line 109: Not a blocking comment, but I would find it much more readable with a f-string.

Wescoeur: Right. I forget it's not python2. :)

stormi (approved):
Approving, to the condition that it was tested on both 8.2 and 8.3.

stormi:
Is it just missing a review from @MatiasVara?

MatiasVara (approved):

PR #141, opened 2022-12-06
Test fileserver HTTPS redirection

benjamreis:

### DIFF
--- a/tests/misc/test_file_server.py
+++ b/tests/misc/test_file_server.py
@@ -1,0 +1,19 @@
+import subprocess
+
+# These tests are meant to test an host fileserver behavior.
+#
+# Requirements:
+# - an XCP-ng host >= 8.2 with latest updates.
+#   The host must be configured with `website-https-only` set to true (which is the default config).
+
+def test_fileserver_redirect_https(host):
+    path = "/path/to/dir/file.txt"
+    process = subprocess.Popen(
+        ["curl", "-i", "http://" + host.hostname_or_ip + path],
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE
+    )
+    stdout, _ = process.communicate()
+    lines = stdout.decode().splitlines()
+    assert lines[0].strip() == "HTTP/1.1 301 Moved Permanently"
+    assert lines[2].strip() == "location:https://" + host.hostname_or_ip + path


tests/misc/test_basic_without_ssh.py
```diff
@@ -72,6 +73,18 @@ def test_checkpoint(self, imported_vm):
         finally:
             snapshot.destroy(verify=True)
 
+    def test_fileserver_redirect_https(self, host):
```
stormi, line 76: This file is not the best place to add this test: it's a collection of tests that run with *a VM* (without using SSH).

We'll need a new test file in misc.

tests/misc/test_basic_without_ssh.py
```diff
@@ -72,6 +73,18 @@ def test_checkpoint(self, imported_vm):
...
+        path = "/path/to/dir/file.txt"
+        process = subprocess.Popen(
+            ["curl", "-i", "http://" + host.hostname_or_ip + path],
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE
+        )
+        stdout, stderr = process.communicate()
+        lines = stdout.decode().split('\r\n')
```
stormi, line 84: I think this is equivalent in this context and more elegant: https://docs.python.org/3/library/stdtypes.html#str.splitlines

```suggestion
        lines = stdout.decode().splitlines()
```

tests/misc/test_basic_without_ssh.py
```diff
@@ -1,5 +1,6 @@
 import logging
 import pytest
+import subprocess
```
stormi, line 3: To remove now

tests/misc/test_file_server.py
```diff
@@ -0,0 +1,13 @@
+import subprocess
+
```
stormi, line 2: Even if the requirements are simple, please add a "requirements" comment on top.

tests/misc/test_file_server.py
```diff
@@ -0,0 +1,19 @@
...
+# These tests are meant to test an host fileserver behavior.
+#
+# Requirements:
+# - a host XCP-ng host >= 8.2 with latest updates.
```
stormi, line 6:
```suggestion
# - an XCP-ng host >= 8.2 with latest updates.
```

stormi (approved):
Looks good. You can squash your commits. We'll wait before the merge, as it would break the tests for the current state of XCP-ng 8.2.

stormi:
The change was pushed to XCP-ng 8.2 updates, so let's merge.

PR #142, opened 2023-02-03
Avoid host crashes when running XTF's XSA 304 POC

stormi:
We want other tests to run even if a host hardware is vulnerable to XSA-304, so we change Xen's `ept` parameter "live" before running XTF.

UNTESTED: CC @gduperrey.


gduperrey (approved):
tests ok on two different hosts.

PR #144, opened 2023-02-10
Add TLS verification tests

stormi:


lib/common.py
```diff
@@ -152,4 +152,4 @@ def _param_get(host, xe_prefix, uuid, param_name, key=None, accept_unknown_key=F
             value = None
         else:
             raise
-    return value
```
benjamreis: Is this related to this PR?

jobs.py
```diff
@@ -24,6 +24,7 @@
             "tests/misc",
             "tests/migration",
             "tests/snapshot",
+            "tests/xapi",
             "tests/xapi-plugins",
```
benjamreis, line 28: Is it related to this PR?

stormi: Yes. Every new test must belong to a job if we want them to be tested regularly.

gduperrey (approved):

Wescoeur (approved):

PR #145, opened 2023-02-23
Test `VIF`'s allowed IP

benjamreis:
- Skip tests on XCP-ng hosts with a SDN controller configured
- Test allowed IP range with CIDR for XCP-ng >= 8.3

### DIFF
--- a/lib/basevm.py
+++ b/lib/basevm.py
@@ -2,7 +2,7 @@ import logging
 
 import lib.commands as commands
 
-from lib.common import _param_get
+from lib.common import _param_get, _param_remove, _param_set
 from lib.sr import SR
 
 class BaseVM:
@@ -15,29 +15,13 @@ class BaseVM:
         self.host = host
 
     def param_get(self, param_name, key=None, accept_unknown_key=False):
-        return _param_get(self.host, self.xe_prefix, self.uuid, param_name, key, accept_unknown_key)
+        return _param_get(self.host, BaseVM.xe_prefix, self.uuid, param_name, key, accept_unknown_key)
 
     def param_set(self, param_name, key, value):
-        args = {'uuid': self.uuid}
+        _param_set(self.host, BaseVM.xe_prefix, self.uuid, param_name, value, key)
 
-        if key is not None:
-            param_name = '{}:{}'.format(param_name, key)
-
-        args[param_name] = value
-
-        return self.host.xe('vm-param-set', args)
-
-    def param_remove(self, param_name, key=None, accept_unknown_key=False):
-        args = {'uuid': self.uuid, 'param-name': param_name}
-        if key is not None:
-            args['param-key'] = key
-        try:
-            self.host.xe('vm-param-remove', args)
-        except commands.SSHCommandFailed as e:
-            if key and accept_unknown_key and e.stdout == "Error: Key %s not found in map" % key:
-                pass
-            else:
-                raise
+    def param_remove(self, param_name, key, accept_unknown_key=False):
+        _param_remove(self.host, BaseVM.xe_prefix, self.uuid, param_name, key, accept_unknown_key)
 
     def name(self):
         return self.param_get('name-label')
--- a/lib/common.py
+++ b/lib/common.py
@@ -153,3 +153,35 @@ def _param_get(host, xe_prefix, uuid, param_name, key=None, accept_unknown_key=F
         else:
             raise
     return value
+
+def _param_set(host, xe_prefix, uuid, param_name, value, key=None):
+    """ Common implementation for param_set. """
+    args = {'uuid': uuid}
+
+    if key is not None:
+        param_name = '{}:{}'.format(param_name, key)
+
+    args[param_name] = value
+
+    host.xe(f'{xe_prefix}-param-set', args)
+
+def _param_add(host, xe_prefix, uuid, param_name, value, key=None):
+    """ Common implementation for param_add. """
+    param_key = f'{key}={value}' if key is not None else value
+    args = {'uuid': uuid, 'param-name': param_name, 'param-key': param_key}
+
+    host.xe(f'{xe_prefix}-param-add', args)
+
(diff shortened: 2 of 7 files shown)


lib/basevm.py
```diff
@@ -15,29 +15,13 @@ def __init__(self, uuid, host):
         self.host = host
 
     def param_get(self, param_name, key=None, accept_unknown_key=False):
-        return _param_get(self.host, self.xe_prefix, self.uuid, param_name, key, accept_unknown_key)
+        return _param_get(self.host, BaseVM.xe_prefix, self.uuid, param_name, key, accept_unknown_key)
```
stormi, line 18: I think using `type(self)` instead of `BaseVM` would make the code more generic and flexible. However I'm not sure if there are drawbacks to this approach.

lib/common.py
```diff
@@ -153,3 +153,37 @@ def _param_get(host, xe_prefix, uuid, param_name, key=None, accept_unknown_key=F
...
+    except commands.SSHCommandFailed as e:
+        if accept_unknown_key and e.stdout == "Error: Key %s not found in map" % key:
+            pass
+        else:
+            raise
+
+def _param_clear(host, xe_prefix, uuid, param_name):
+    """ Common implementation for param_clear. """
+    args = {'uuid': uuid, 'param-name': param_name}
+    host.xe(f'{xe_prefix}-param-clear', args)
```
stormi, line 189: I don't see this function being used.

benjamreis: incoming

lib/basevm.py
benjamreis: AFAIK this is common practice + easier to read IMHO.

lib/common.py
stormi: Ideally the function would either be introduced when needed, or added just before it's needed in a separate commit, or least be mentioned in the commit message. But you already resolved the conversation and I don't want to argue more on this topic.

stormi (approved):

benjamreis:
Tested and successfull on both XCP-ng 8.2 and 8.3

tests/network/test_vif_allowed_ip.py
```diff
@@ -0,0 +1,87 @@
...
+
+        vif.param_add(f"ipv{ip_family}-allowed", ip)
+        assert ip_responsive(ip)
+
+        vif.param_remove(f"ipv{ip_family}-allowed", ip)
+        assert not ip_responsive(ip)
+
+        vif.param_set("locking-mode", "unlocked")
+
+    def test_unallowed_cidr(self, host_at_least_8_3, running_vm):
```
stormi, line 51: The host_at_least_8_3 fixture is not used in the function, and marking it explicitly with pytest makes it easier to spot. This would be consistent with other tests.

Alternatively, we could move these two tests to a separate class and mark the whole class with this fixture, rather than the test functions individually.

```suggestion
    @pytest.mark.usefixtures("host_at_least_8_3")
    def test_unallowed_cidr(self, running_vm):
```

tests/network/test_vif_allowed_ip.py
```diff
@@ -0,0 +1,87 @@
...
+        assert ip_responsive(ip)
+
+        vif.param_add(f"ipv{ip_family}-allowed", dummy_cidr)
+        vif.param_set("locking-mode", "locked")
+        assert not ip_responsive(ip)
+
+        vif.param_clear(f"ipv{ip_family}-allowed")
+        vif.param_set("locking-mode", "unlocked")
+
+    def test_allowed_cidr(self, host_at_least_8_3, running_vm):
```
stormi, line 68: Idem

tests/network/test_vif_allowed_ip.py
benjamreis: i'll split in 2 classes :)

stormi (commented):
Looks good, except the remaining FIXMEs

lib/basevm.py
ydirson, line 18: What's the benefit in hardcoding the class name here and in other places ?  It is likely to cause unwanted issues or at least extra work on class rename, copy-paste, etc

tests/network/conftest.py
```diff
@@ -0,0 +1,7 @@
+import pytest
+
+@pytest.fixture(scope='package')
+def host_no_sdn_controller(host):
+    """ An XCP-ng with no SDN controller. """
+    if host.xe('sdn-controller-list', minimal=True) != '':
+        pytest.skip(f"This test requires an XCP-ng with no SDN controller")
```
ydirson, line 7: formulation makes me uneasy - wouldn't something like "conflicts with the presence of a SDN controller" be correct ?

lib/common.py
```diff
@@ -153,3 +153,37 @@ def _param_get(host, xe_prefix, uuid, param_name, key=None, accept_unknown_key=F
...
+def _param_remove(host, xe_prefix, uuid, param_name, key, accept_unknown_key=False):
+    """ Common implementation for param_remove. """
+    args = {'uuid': uuid, 'param-name': param_name, 'param-key': key}
+    try:
+        host.xe(f'{xe_prefix}-param-remove', args)
+    except commands.SSHCommandFailed as e:
+        if accept_unknown_key and e.stdout == "Error: Key %s not found in map" % key:
+            pass
+        else:
+            raise
```
Wescoeur, line 184: Or:
```python
if not accept_unknown_key or e.stdout != "Error: Key %s not found in map" % key:
    raise
```

tests/network/conftest.py
Wescoeur: Why not:

```python
if host.xe('sdn-controller-list', minimal=True)
```

instead of:

```python
if host.xe('sdn-controller-list', minimal=True) != ''
```

Wescoeur (approved):

benjamreis:
Waiting for XAPI `23.6.0` release to merge.

stormi:
XAPI was updated to `23.24.0`

PR #148, opened 2023-04-17
Add --json to xo-cli calls when needed

bleader:
Originally xo-cli output was in json, then it moved to more human
readable output by default. In v0.17.0 --json option allows users to
choose between the more readable and the json output. Versions in
between are not compatible with xcp-ng-tests at all, therefore, we
now need at least v0.17.0.

### DIFF
--- a/lib/xo.py
+++ b/lib/xo.py
@@ -1,17 +1,23 @@
 import json
 import subprocess
 
-def xo_cli(action, args={}, check=True, simple_output=True):
+def xo_cli(action, args={}, check=True, simple_output=True, use_json=False):
+    run_array = ['xo-cli', action]
+    if use_json:
+        run_array += ['--json']
+    run_array += ["%s=%s" % (key, value) for key, value in args.items()]
     res = subprocess.run(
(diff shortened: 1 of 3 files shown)


stormi (approved):
Approved, but before merging we'll run the full test suite with this change.

stormi:
As discussed just now: we also need to update the README to mention the version requirement.

bleader:
Added a mention of >= 0.17.0 for `xo-cli` let me know if you think we need more.

lib/xo.py
```diff
@@ -1,15 +1,21 @@
...
+        run_array,
         stdout=subprocess.PIPE,
         stderr=subprocess.STDOUT,
         check=check
     )
-    if simple_output:
+    if simple_output and not use_json:
         return res.stdout.decode().strip()
+    elif use_json:
+        return json.loads(res.stdout)
```
stormi, line 18: To me, json.loads should be called on the decoded output, and not in the case when simple_output is False.

Untested:
```suggestion
    if simple_output:
        output = res.stdout.decode().strip()
        if use_json:
            return json.loads(output)
        return return output       
```

stormi (approved):

PR #149, opened 2023-06-01
Add a new test to verify a package update

Wescoeur:


tests/xapi-plugins/plugin_updater/test_updater.py
```diff
@@ -19,6 +19,23 @@ def test_update(self, host):
 
         host.yum_restore_saved_state()
 
+    def test_package_update(self, host):
+        host.yum_save_state()
```
stormi, line 23: There's the `host_with_saved_yum_state` fixture for this.

tests/xapi-plugins/plugin_updater/test_updater.py
```diff
@@ -19,6 +19,23 @@ def test_update(self, host):
...
+
+        packages = host.get_available_package_versions('dummypkg')
+        assert len(packages) >= 2
+        assert packages[0] < packages[1]
```
stormi, line 27: Comparing versions as a simple string comparison is not reliable at all, and in the case of RPMs EVR comparison it's even less reliable.

Reliably comparing the versions of two RPMs is a rather complicated task, which is usually delegated to `python-rpm`, but if we want to avoid such a dependency for xcp-ng-tests,  we should instead find a way to avoid relying on it (unless we really want to add a dependency to python-rpm in our tests).

If we consider that dummypkg-1.0-1 and dummypkg-1.0-2 must always be present in the repositories, then you install the first one with `yum install "dummypkg < 1.0-2"`. 

And create a separate test for testing the installation of a package through the updater.

tests/xapi-plugins/plugin_updater/test_updater.py
```diff
@@ -19,6 +19,23 @@ def test_update(self, host):
...
+
+        try:
+            assert not host.is_package_installed(packages[0])
+            host.call_plugin('updater.py', 'install', {'packages': packages[0]})
+            assert host.is_package_installed(packages[0])
+
+            host.call_plugin('updater.py', 'update', {'packages': 'dummypkg'})
+            assert host.is_package_installed(packages[1])
+        finally:
```
benjamreis, line 36: Perhaps also test the inferior version is not still installed?

tests/xapi-plugins/plugin_updater/test_updater.py
Wescoeur: Or I can assert like that:
```python
assert packages[0] == 'dummypkg-0:1.0-1.xcpng8.2.noarch'
assert packages[1] == 'dummypkg-0:1.0-2.xcpng8.2.noarch'
```
I'm not sure I need to add another test or use an additional dependency.
`dummypkg` is dedicated to testing, we use it for that. If we assume we have two versions of this package, then that's enough.

benjamreis (approved):

benjamreis (approved):

tests/xapi-plugins/plugin_updater/test_updater.py
stormi: And now your test will fail in XCP-ng 8.3

Wescoeur: I can also use `host.xcp_version`...

stormi:
```
assert packages[0].startswith('dummypkg-0:1.0-1.xcpng')
assert packages[1].startswith('dummypkg-0:1.0-2.xcpng')
```

Should work, but I can't promise the order of packages in the output of repoquery is always fixed...

Wescoeur: Yes I know, but it's also unlikely that the repoquery output isn't deterministic.

ydirson: (alphabetically) sort the list before checking it?

Wescoeur: For the moment I will not sort the list. I assume repoquery returns valid output for our tests.
if one day it breaks, I will modify it. :p

PR #152, opened 2023-06-19
Test the presence of HTTP Strict Transport Security

gthvn1:


gthvn1:
It requires a patch that I didn't push as PR yet. I just create it on my repo: https://github.com/gthvn1/xen-api/pull/1

benjamreis (approved):
LGTM We'll wait for the patch to be included in XAPI to merge the test.

stormi (commented):
We said HSTS should probably be opt-in, because of the self signed certs you get first when installing. Has it been assessed it's not necessary anymore?

tests/misc/test_file_server.py
```diff
@@ -17,3 +17,14 @@ def test_fileserver_redirect_https(host):
...
+def test_fileserver_hsts(host):
+    process = subprocess.Popen(
+        ["curl", "-XGET", "-i", "https://" + host.hostname_or_ip],
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE
+    )
+    stdout, _ = process.communicate()
+    lines = stdout.decode().splitlines()
+    assert lines[0].strip() == "HTTP/1.1 200 OK"
+    assert lines[3].strip() == "Strict-Transport-Security:max-age=63072000; includeSubDomains; preload"
```
stormi, line 30: I'm not sure we want `includeSubDomains` and even less `preload`.

gthvn1: Agree with preload. I'm not sure either.  But if we want to enforce the usage of HTTPS I think that `includeSubDomains` makes sense.

gthvn1:
> We said HSTS should probably be opt-in, because of the self signed certs you get first when installing. Has it been assessed it's not necessary anymore?

By opt-in you mean adding an option when XAPI starts to enable the support of HTSTS (and by default it is disabled)?

tests/misc/test_file_server.py
benjamreis: Actually we can't presume of how a XCP-ng user will use the subdomains so better be safe and only use the age.

gthvn1:
So I modified the test. Now that we have the option the test starts by checking that HSTS is disabled (that is the default behavior), then it enables it and checks that header is now present. Before checking the assert we restore the default parameters that is HSTS disabled.

gthvn1:
In fact I disabled the restoration of the default setup because systemctl is complaining that I call the restart of the service too often in a short period of time. So maybe we can modify this behavior but otherwise I think it is ok to keep the HSTS enabled after the test no?

lib/host.py
```diff
@@ -450,3 +450,11 @@ def varstore_dir(self):
             return "/var/lib/uefistored"
         else:
             return "/var/lib/varstored"
+
+    def enable_hsts_header(self):
+        self.ssh([f'echo "add_hsts_response_header = true" > {XAPI_CONF_DIR}/00-XCP-ng-tests-enable-hsts-header.conf'])
```
benjamreis, line 455:
```suggestion
        self.ssh(['echo',  '"add_hsts_response_header = true"',  '>',  f'{XAPI_CONF_DIR}/00-XCP-ng-tests-enable-hsts-header.conf'])
```

tests/misc/test_file_server.py
```diff
@@ -17,3 +17,37 @@ def test_fileserver_redirect_https(host):
...
+            ["curl", "-XGET", "-k", "-I", "https://" + host.hostname_or_ip],
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE
+        )
+        stdout, _ = process.communicate()
+        return stdout.decode().splitlines()
+
+    # By default HSTS header is not set by XAPI server.
+    for line in get_header():
+        assert line != "Strict-Transport-Security:max-age=63072000"
```
benjamreis, line 33: `"Strict-Transport-Security:max-age=63072000"` could me put in variable since used twice.

tests/misc/test_file_server.py
```diff
@@ -17,3 +17,37 @@ def test_fileserver_redirect_https(host):
...
+    def get_header():
+        process = subprocess.Popen(
+            ["curl", "-XGET", "-k", "-I", "https://" + host.hostname_or_ip],
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE
+        )
+        stdout, _ = process.communicate()
+        return stdout.decode().splitlines()
+
+    # By default HSTS header is not set by XAPI server.
```
benjamreis, line 31: Maybe we can assert it's not to be sure the default behavior is respected.

tests/misc/test_file_server.py
```diff
@@ -17,3 +17,37 @@ def test_fileserver_redirect_https(host):
...
+    # Before doing the assert we restore the default state because if the assert
+    # is triggered the HSTS will be enabled.
+    # BUT THERE IS AN ISSUE WHIT THAT:
+    #   lib.commands.SSHCommandFailed: SSH command (xe-toolstack-restart) failed
+    #   with return code 1: Executing xe-toolstack-restart Job for varstored-guard.service
+    #   failed because start of the service was attempted too often.
+    # So disabled it for now
+    # host.disable_hsts_header()
+
+    assert hsts_enabled
```
benjamreis, line 53: disable HSTS header at the end.

conftest.py
```diff
@@ -150,6 +150,13 @@ def host_less_than_8_3(host):
     if not host.xcp_version < version.parse(version_str):
         pytest.skip(f"This test requires an XCP-ng < {version_str} host")
 
+@pytest.fixture(scope='session')
+def host_with_hsts(host):
+    host.enable_hsts_header()
+    yield host
+    host.disable_hsts_header()
+
+
```
benjamreis, line 159:
```suggestion

```

stormi:
Was this closed on purpose?

gthvn1:
Ooops no :-(
I will redo it

gthvn1:
branch on my repo was deleted without remembering that there is a PR in progress (this one :) ). So I created a new one that is the same here #164 
Sorry for the inconvenience and thx to @stormi to spot it :pray:

PR #153, opened 2023-07-17
Add a new test to check that extras pkgs can be installed

gthvn1:
This test gets a list of packages from
https://reports.xcp-ng.org/<xcp-ng-version>/extra_installable.txt and check that all packages can be installed. It saves the yum state before installing the package and restore it after.

### DIFF
--- a/tests/packages/extra/conftest.py
+++ b/tests/packages/extra/conftest.py
@@ -1,0 +1,12 @@
+import logging
+import pytest
+import urllib.request
+
+@pytest.fixture(scope="session")
+def extra_pkgs(host):
+    version = host.xcp_version
+    url = f"https://reports.xcp-ng.org/{version}/extra_installable.txt"
+
+    logging.info(f"Getting extra packages from {url}")
+    response = urllib.request.urlopen(url)
+    return response.read().decode('utf-8').splitlines()
--- a/tests/packages/extra/test_extra_group_pkgs.py
+++ b/tests/packages/extra/test_extra_group_pkgs.py
@@ -1,0 +1,11 @@
+# Explicitly import package-scoped fixtures (see explanation in pkgfixtures.py)
+from pkgfixtures import host_with_saved_yum_state
+
+def test_extra_group_packages_url_resolved(host, extra_pkgs):
+    for p in extra_pkgs:
+        host.ssh(['yumdownloader', '--resolve', '--urls', p])
+        
+def test_extra_group_packages_can_be_installed(host_with_saved_yum_state, extra_pkgs):
+    # Just try to install all packages together. Installing them one by one
+    # takes too much time due to the generation of the initrd.
+    host_with_saved_yum_state.yum_install(extra_pkgs)
--- a/lib/host.py
+++ b/lib/host.py
@@ -275,7 +275,7 @@ class Host:
     def yum_install(self, packages, enablerepo=None):
         logging.info('Install packages: %s on host %s' % (' '.join(packages), self))
         enablerepo_cmd = ['--enablerepo=%s' % enablerepo] if enablerepo is not None else []
-        return self.ssh(['yum', 'install', '-y'] + enablerepo_cmd + packages)
+        return self.ssh(['yum', 'install', '--setopt=skip_missing_names_on_install=False', '-y'] + enablerepo_cmd + packages)
 
     def yum_remove(self, packages):
         logging.info('Remove packages: %s from host %s' % (' '.join(packages), self))


gthvn1:
When tested I found that restoring the yum state after each install takes several minutes... I don't know if it is an issue with my env (my xcp-ng host is a VM) but test is really slow. Shouldn't we save the state at the beginning and restore it once at the end?

tests/misc/test_extra_group_pkgs_can_be_installed.py
```diff
@@ -0,0 +1,44 @@
...
+        ret = run(
+            ["wget", "-O-", "-q", url],
+            capture_output=True,
+            check=True)
+    except CalledProcessError:
+        logging.error(f"failed to get extra installable list from {url}")
+        assert False
+
+    pkgs = ret.stdout.decode("utf-8").splitlines()
+    logging.info(f"{pkgs}")
```
gthvn1, line 41: can probably be removed...

stormi:
> When tested I found that restoring the yum state after each install takes several minutes... I don't know if it is an issue with my env (my xcp-ng host is a VM) but test is really slow. Shouldn't we save the state at the beginning and restore it once at the end?

This is caused by the re-gen of initrd after a driver RPM is installed or removed. It is a post-transaction operation which happens once no matter the number of driver RPMs are installed. So there's a big wait twice in the current test implementation: once after the installation, once when coming back to the saved state (the rollback operation runs `yum remove` for added packages).

So the fastest test would be one which installs all packages at once and they restores the state, rather than installing them one by one. The output may be a bit messy when it fails, but it may be an acceptable trade-off.

tests/misc/test_extra_group_pkgs_can_be_installed.py
```diff
@@ -0,0 +1,44 @@
+import logging
+import pytest
+
+from lib.commands import SSHCommandFailed
+from packaging import version
+from subprocess import run, CalledProcessError
```
stormi, line 6: In this project we tend to `import subprocess` as a whole.

tests/misc/test_extra_group_pkgs_can_be_installed.py
```diff
@@ -0,0 +1,44 @@
...
+    return failed
+
+def test_extra_group_packages_can_be_installed(host):
+    v = version.parse(host.inventory['PRODUCT_VERSION'])
+    url = f"https://reports.xcp-ng.org/{v}/extra_installable.txt"
+    logging.info(f"getting extra packages from {url}")
+
+    try:
+        ret = run(
+            ["wget", "-O-", "-q", url],
```
stormi, line 33: As this is executed on the client side, we could simply use python to download the file. Using `urllib.request.urlopen` for example (https://stackoverflow.com/questions/7243750/download-file-from-web-in-python-3)

tests/misc/test_extra_group_pkgs_can_be_installed.py
```diff
@@ -0,0 +1,44 @@
...
+            check=True)
+    except CalledProcessError:
+        logging.error(f"failed to get extra installable list from {url}")
+        assert False
+
+    pkgs = ret.stdout.decode("utf-8").splitlines()
+    logging.info(f"{pkgs}")
+    failures= [p for p in pkgs if install_failed(p, host)]
+
+    assert failures == []
```
stormi, line 44: You don't need to store failures and assert at the end. Instead, just let the yum command raise an error, and pytest will convert this in a test failure. Which means of course that you must ensure the teardown part is always executed to restore the previous state, but this is easily done through pytest's `fixtures`.

We would lose some information, as not all packages would be tested, but we usually stop at the first error when we run our testsuites anyway, because a failed test may leave the system under test in a state which is not suitable for the next tests, despite our efforts at restoring a good state. Or an early test failure may reveal that we launched tests on a system that is not in an appropriate state.

tests/misc/test_extra_group_pkgs_can_be_installed.py
```diff
@@ -0,0 +1,44 @@
...
+    try:
+        h.yum_install([pkg_name])
+    except SSHCommandFailed:
+        failed = True
+
+    h.yum_restore_saved_state()
+    return failed
+
+def test_extra_group_packages_can_be_installed(host):
+    v = version.parse(host.inventory['PRODUCT_VERSION'])
```
stormi, line 27: I don't think it's necessary to parse the version into a version object, and it's bit risky to then rely on its conversion back to string to generate the URL.

You can probably just keep the original string all along.

tests/misc/test_extra_group_pkgs_can_be_installed.py
gthvn1: I think that with a list of packages that failed we are able to test all packages and the ouptut will be easy to read. Let the yum command raises an error will be less readable I think. But both are ok for me.

tests/misc/test_extra_group_pkgs_can_be_installed.py
benjamreis: But the URL is not the same for 8.2 and 8.3.

benjamreis: you can ge the version from `host.xcp_version` though.
So either read the string from the inventory and don't parse it. Or use the one stored in the host object and stringify it.
I would simply read it and not use `version`.

tests/misc/test_extra_group_pkgs_can_be_installed.py
```diff
@@ -0,0 +1,22 @@
...
+    try:
+        response = urllib.request.urlopen(url)
+        pkgs = response.read().decode('utf-8').splitlines()
+        # Just try to install the package. If yum fails then it will be
+        # caught by pytest.
+        for p in pkgs:
+            host.yum_install([p])
+    except HTTPError:
+        logging.error(f"failed to get extra packages list from {url}")
+        assert False
```
benjamreis, line 22:
```suggestion
        assert False

```

gthvn1: fixed

stormi:
> LGTM please add the change of behavior of `yum_install` in the commit message as well.

It would be even better to make this change in a first dedicated commit.

tests/packages/extra/conftest.py
```diff
@@ -0,0 +1,12 @@
+import logging
+import pytest
+import urllib.request
+
+@pytest.fixture(scope="session")
+def extra_pkgs(host):
+    version = host.xcp_version
+    url = f"https://reports.xcp-ng.org/{version}/extra_installable.txt"
+
+    logging.info(f"getting extra packages from {url}")
```
stormi, line 10: Most our logs begin with a capital letter in this project.

```suggestion
    logging.info(f"Getting extra packages from {url}")
```

tests/packages/extra/test_extra_group_pkgs.py
```diff
@@ -0,0 +1,12 @@
+from lib.commands import SSHCommandFailed
+from pkgfixtures import host_with_saved_yum_state
```
stormi, line 2: Adding the usual comment for package-scoped fixtures, since this explicit import should not be necessary if we weren't workarounding a pytest issue.

```suggestion

# Explicitly import package-scoped fixtures (see explanation in pkgfixtures.py)
from pkgfixtures import host_with_saved_yum_state
```

tests/packages/extra/test_extra_group_pkgs.py
```diff
@@ -0,0 +1,12 @@
...
+
+def test_extra_group_packages_url_resolved(host, extra_pkgs):
+    for p in extra_pkgs:
+        host.ssh(['yumdownloader', '--resolve', '--urls', p])
+        
+def test_extra_group_packages_can_be_installed(host_with_saved_yum_state, extra_pkgs):
+    # Just try to install all packages together. Installing them one by one
+    # takes too much time due to the generation of the initrd.
+    # If yum fails then it will be caught by pytest.
```
stormi, line 11: This is a tacit contract with pytest, so we don't really need to mention it. It is true for almost anything we do in our tests.

```suggestion
```

stormi (changes requested):
Looks good to me. Just a few minor changes requested to improve the consistency with the rest of the tests.

tests/packages/extra/test_extra_group_pkgs.py
gthvn1: makes sense :+1:

tests/packages/extra/test_extra_group_pkgs.py
```diff
@@ -0,0 +1,13 @@
+from lib.commands import SSHCommandFailed
```
gthvn1, line 1: oops can be removed now

PR #155, opened 2023-07-18
check extras pkgs

gthvn1:


gthvn1:
I closed https://github.com/xcp-ng/xcp-ng-tests/pull/153 and create the same PR (this one) but in the xcp-ng repo directly in order to trigger new checks. So former discussions can be read on https://github.com/xcp-ng/xcp-ng-tests/pull/153

tests/packages/extra/test_extra_group_pkgs.py
```diff
@@ -0,0 +1,11 @@
+# Explicitly import package-scoped fixtures (see explanation in pkgfixtures.py)
+from pkgfixtures import host_with_saved_yum_state
+
```
stormi, line 3: Another unwritten rule is that a test must contain a "requirements" comment
```suggestion

# Requirements:
# From --hosts parameter:
# - host(A1): any master host of a pool, with access to XCP-ng RPM repositories and reports.xcp-ng.org.

```

stormi (approved):

gduperrey:
There is a problem when we run this on 8.2.1. The right url use the version 8.2 and not 8.2.1.
We need to adapt the script to generate the url or to have a link in http(s) to reach the right url.

gduperrey:
8.2 :white_check_mark:

PR #158, opened 2023-07-28
Ignore stdout/stderr messages before "yum history list"

Wescoeur:
Yum stderr logs as "Warning: RPMDB modified outside of yum." are appended to the SSH output which completely breaks the history parsing. To solve this problem, we need to skip all log lines before transactions.

### DIFF
--- a/lib/host.py
+++ b/lib/host.py
@@ -272,10 +272,19 @@ class Host:
             history_str = self.ssh(['yum', 'history', 'list', '--noplugins'])
 
         history = history_str.splitlines()
+        line_index = None
+        for i in range(len(history)):
+            if history[i].startswith('--------'):
+                line_index = i
+                break
+
+        if line_index is None:
+            raise Exception('Unable to get yum transactions')
+
         try:
-            return int(history[2].split()[0])
+            return int(history[line_index + 1].split()[0])
         except ValueError:
-            raise Exception('Unable to parse correctly last yum history tid. Output\n:' + history_str)
+            raise Exception('Unable to parse correctly last yum history tid. Output:\n' + history_str)
 
     def yum_install(self, packages, enablerepo=None):
         logging.info('Install packages: %s on host %s' % (' '.join(packages), self))


ydirson:
Maybe explain why those warnings are here ?

benjamreis (approved):

gduperrey (approved):

Wescoeur:
This warn is an example. It's not the first time that we have a problem like this with a yum command, this fix is here to handle all cases, not just this one.

But for the context of this example, you can get this trace when you install an RPM without using a yum command.

ydirson:
> But for the context of this example, you can get this trace when you install an RPM without using a yum command.

This precise case would possibly be better fixed by using `yum` to install a local RPM file, since that works too.

I'm a bit wary of ignoring input more largely, as this can help us miss some real issues.

benjamreis:
The example is a benine warning so ignoring it is the correct way to handle.
Also having a specific behavior for every warning would be too complex.
The error is still logged as before so in case a non benine error happened (which hasn't for now) the test would fail and we'd see it in the logs.

ydirson:
This specific warning is benign and we know how to avoid it.

My point is that this patch ignores *all warnings* -- how can we be sure that all the warnings we can get will be benign as well?
I'm still not convinced.

Wescoeur:
I'm not convinced to abort when there is a warn. A warn is not an error. From my POV, If we can install packages without issue just after this savepoint, it's ok.

Also: the warns were totally ignored before my previous commit and the yum rollback command crashed because we didn't have a valid tid. And of course we didn't have the source of the problem at all. Now we have a log + always a valid tid.

ydirson:
A warning for a human operator is not an error, but there is a reason why compiler flags like `-Werror` exist: they are here to warn the dev that something suspect is happening - as in the case described above, even though the impact of that particular case may be benign.

A warning is something like "this is not necessarily an error, I'm not sure what you're doing, please be more specific or fix the problem if it was not intended". Else there would be no warning in the first place.

benjamreis:
Until now all warning has been benign.
If something else comes up we'll treat the specific case.

ydirson:
> Until now all warning has been benign.

"jusqu'ici tout va bien"

> If something else comes up we'll treat the specific case.

If something else comes up, my understanding is it be silently dropped and we won't see we have another specific case to handle - or did I miss something?

benjamreis:
tested and validated.

PR #160, opened 2023-08-23
vTPM basic tests

tescande:
This PR adds pytest script to create a vtpm device attached to a VM, execute some basic TPM tests, and destroy the vtpm device.

### DIFF
--- a/lib/vm.py
+++ b/lib/vm.py
@@ -411,6 +411,19 @@ class VM(BaseVM):
 
         return binaries
 
+    def get_vtpm_uuid(self):
+        return self.host.xe('vtpm-list', {'vm': self.uuid}, minimal=True)
+
+    def create_vtpm(self):
+        logging.info("Creating vTPM for vm %s" % self.uuid)
+        return self.host.xe('vtpm-create', {'vm-uuid': self.uuid})
+
+    def destroy_vtpm(self):
+        vtpm_uuid = self.get_vtpm_uuid()
+        assert vtpm_uuid, "A vTPM must be present"
+        logging.info("Destroying vTPM %s" % vtpm_uuid)
+        return self.host.xe('vtpm-destroy', {'uuid': vtpm_uuid}, force=True)
+
     def clone(self):
         name = self.name() + '_clone_for_tests'
         logging.info("Clone VM")
--- a/lib/vm.py
+++ b/lib/vm.py
@@ -44,11 +44,21 @@ class VM(BaseVM):
             args['on'] = on
         return self.host.xe('vm-start', args)
 
-    def shutdown(self, force=False, verify=False):
+    def shutdown(self, force=False, verify=False, force_if_fails=False):
+        assert not (force and force_if_fails), "force and force_if_fails cannot be both True"
         logging.info("Shutdown VM" + (" (force)" if force else ""))
-        ret = self.host.xe('vm-shutdown', {'uuid': self.uuid, 'force': force})
-        if verify:
-            wait_for(self.is_halted, "Wait for VM halted")
+
+        try:
+            ret = self.host.xe('vm-shutdown', {'uuid': self.uuid, 'force': force})
+            if verify:
+                wait_for(self.is_halted, "Wait for VM halted")
+        except Exception as e:
+            if force_if_fails:
+                logging.warning("Shutdown failed: %s" % e)
+                ret = self.shutdown(force=True, verify=verify)
+            else:
+                raise
+
         return ret
 
     def reboot(self, force=False, verify=False):
--- a/tests/vtpm/conftest.py
+++ b/tests/vtpm/conftest.py
@@ -1,0 +1,62 @@
+import pytest
+import logging
+
+from lib.common import PackageManagerEnum
+
+@pytest.fixture(scope='module')
+def halted_uefi_unix_vm(uefi_vm, unix_vm):
+    assert uefi_vm.is_halted(), "The VM must be halted for these tests"
+    yield uefi_vm
+
+@pytest.fixture(scope='module')
+def snapshotted_halted_uefi_unix_vm(halted_uefi_unix_vm):
+    vm = halted_uefi_unix_vm
+    snapshot = vm.snapshot()
+
+    yield vm
+
+    try:
+        snapshot.revert()
+    finally:
+        snapshot.destroy()
+
+@pytest.fixture(scope='module')
+def unix_vm_with_vtpm(snapshotted_halted_uefi_unix_vm):
+    vm = snapshotted_halted_uefi_unix_vm
+
+    vm.create_vtpm()
+    yield vm
+    # Tear down
+    vm.destroy_vtpm()
+
+@pytest.fixture(scope='module')
+def started_unix_vm_with_vtpm(unix_vm_with_vtpm):
+    vm = unix_vm_with_vtpm
+
+    vm.start()
+    try:
+        vm.wait_for_os_booted()
+    except Exception:
+        vm.shutdown(force=True, verify=True)
+        raise
+
+    yield vm
+    # Tear down
+    vm.shutdown(verify=True, force_if_fails=True)
+
+@pytest.fixture(scope='module')
+def unix_vm_with_tpm2_tools(started_unix_vm_with_vtpm):
+    vm = started_unix_vm_with_vtpm
+
+    pkg_mgr = vm.detect_package_manager()
+    if pkg_mgr == PackageManagerEnum.APT_GET:
+        pkg_mgr = 'apt-get'
+    elif pkg_mgr == PackageManagerEnum.RPM:
+        pkg_mgr = 'yum'
+    else:
+        pytest.fail("Unsupported package manager for this test. Cannot install tpm2-tools")
+
+    logging.info("Installing tpm2-tools package using '%s'" % pkg_mgr)
+    vm.ssh([pkg_mgr, 'install', '-y', 'tpm2-tools'])
+
+    yield vm
--- a/tests/vtpm/test_vtpm_basic_operations.py
+++ b/tests/vtpm/test_vtpm_basic_operations.py
@@ -1,0 +1,82 @@
+import logging
+import pytest
+import lib.commands as commands
+
+# These tests are basic tests for vTPM devices.
+#   - Create / Destroy a vTPM device on a VM
+#   - Do some basic encryption tests using tpm2-tools package
+#
+# Requirements:
+# - an XCP-ng host >= 8.3
+# - a RPM-based or DEB-based UEFI VM with `tpm2-tools` installable
+#   from default repositories
+
+vtpm_signing_test_script = """#!/bin/env bash
+
+set -ex
+
+tpm2_selftest --fulltest
+
+tpm2_getrandom 32 > /dev/null
+
+TMPDIR=`mktemp -d`
+
+# Create an Endorsement primary key
+tpm2_createprimary --hierarchy e --key-context ${TMPDIR}/primary.ctx > /dev/null
+
+# Create key objects
+tpm2_create --key-algorithm rsa --public ${TMPDIR}/rsa.pub --private ${TMPDIR}/rsa.priv --parent-context \
+            ${TMPDIR}/primary.ctx > /dev/null
+
+# Load keys into the TPM
+tpm2_load --parent-context ${TMPDIR}/primary.ctx --public ${TMPDIR}/rsa.pub --private ${TMPDIR}/rsa.priv \
+          --key-context ${TMPDIR}/rsa.ctx > /dev/null
+
+# Delete loaded key files
+rm -f ${TMPDIR}/rsa.pub ${TMPDIR}/rsa.priv
+
+# Message to sign
+echo 'XCP-ng Rulez' > ${TMPDIR}/message.dat
+
+# Sign the message
+tpm2_sign --key-context ${TMPDIR}/rsa.ctx --hash-algorithm sha256 --signature ${TMPDIR}/sig.rssa \
+          ${TMPDIR}/message.dat > /dev/null
+
+# Verify signature
+tpm2_verifysignature --key-context ${TMPDIR}/rsa.ctx --hash-algorithm sha256 --message ${TMPDIR}/message.dat \
+                     --signature ${TMPDIR}/sig.rssa > /dev/null
+
+# Verify with another message
+echo "XCP-ng Still Rulez" > ${TMPDIR}/message.dat
+
+# Verify signature !!!!! THIS MUST FAIL !!!!!
+if tpm2_verifysignature --key-context ${TMPDIR}/rsa.ctx --hash-algorithm sha256 --message ${TMPDIR}/message.dat \
+                        --signature ${TMPDIR}/sig.rssa > /dev/null 2>&1; then
+    echo "Should not succeed"
+    exit 1
+fi
+
+rm -rf ${TMPDIR}
+"""
+
+@pytest.mark.small_vm
+@pytest.mark.usefixtures("host_at_least_8_3")
+def test_create_and_destroy_vtpm(halted_uefi_unix_vm):
+    vm = halted_uefi_unix_vm
+    assert not vm.get_vtpm_uuid(), "there must be no vTPM before we create it"
+    vtpm_uuid = vm.create_vtpm()
+    assert vtpm_uuid
+    assert vm.get_vtpm_uuid(), "a vTPM must be present after creation"
+    logging.info("vTPM created with uuid: %s" % vtpm_uuid)
+    vm.destroy_vtpm()
+    assert not vm.get_vtpm_uuid(), "there must be no vTPM after we deleted it"
+
+@pytest.mark.small_vm
+@pytest.mark.usefixtures("host_at_least_8_3")
+def test_vtpm(unix_vm_with_tpm2_tools):
+    global vtpm_signing_test_script
+    vm = unix_vm_with_tpm2_tools
+
+    # Basic TPM2 tests with tpm2-tools
+    logging.info("Running TPM2 test script on the VM")
+    vm.execute_script(vtpm_signing_test_script)
--- a/jobs.py
+++ b/jobs.py
@@ -221,6 +221,20 @@ JOBS = {
         "params": {},
         "paths": ["tests/xen"],
     },
+    "vtpm": {
+        "description": "Testing vTPM functionalities",
+        "requirements": [
+            "A XCP-ng host >= 8.3 and a Unix RPM-based or DEB-based UEFI VM with "
+            "tpm2-tools installable from default repositories.",
+        ],
+        "nb_pools": 1,
+        "params": {
+            # The test also works on CentOS, for example, but in this job definition
+            # we settle for a debian VM
+            "--vm": "single/debian_uefi_vm",
+        },
+        "paths": ["tests/vtpm"],
+    },
     "flaky": {
         "description": "tests that usually pass, but sometimes fail unexpectedly",
         "requirements": [
--- a/lib/host.py
+++ b/lib/host.py
@@ -67,15 +67,17 @@ class Host:
             suppress_fingerprint_warnings=suppress_fingerprint_warnings, local_dest=local_dest
         )
 
-    def xe(self, action, args={}, check=True, simple_output=True, minimal=False):
+    def xe(self, action, args={}, check=True, simple_output=True, minimal=False, force=False):
         maybe_param_minimal = ['--minimal'] if minimal else []
+        maybe_param_force = ['--force'] if force else []
 
         def stringify(key, value):
             if isinstance(value, bool):
                 return "{}={}".format(key, to_xapi_bool(value))
             return "{}={}".format(key, shlex.quote(value))
 
-        command = ['xe', action] + maybe_param_minimal + [stringify(key, value) for key, value in args.items()]
+        command = ['xe', action] + maybe_param_minimal + maybe_param_force + \
+                  [stringify(key, value) for key, value in args.items()]
         result = self.ssh(
             command,
             check=check,
--- a/vm_data.py-dist
+++ b/vm_data.py-dist
@@ -13,6 +13,8 @@ VMS = {
         "small_vm_efitools": "",
         # "small" Windows VM (UEFI)
         "small_vm_windows": "",
+        # Debian VM (UEFI, no GUI)
+        "debian_uefi_vm": "",
     },
     "multi": {
         # all VMs we want to run "multi_vms" tests on


tests/misc/test_vtpm_basic_operations.py
```diff
@@ -0,0 +1,112 @@
+import logging
+import pytest
+import lib.commands as commands
+
```
stormi, line 4: Every test must have a "Requirements" comment on top listing the requirements of the test (one host, several, if a VM, what kind of VM, etc.).

tests/misc/test_vtpm_basic_operations.py
```diff
@@ -0,0 +1,117 @@
...
+if tpm2_verifysignature --key-context ${TMPDIR}/rsa.ctx --hash-algorithm sha256 --message ${TMPDIR}/message.dat \
+                        --signature ${TMPDIR}/sig.rssa > /dev/null 2>&1; then
+    echo "Should not succeed"
+    exit 1
+fi
+
+rm -rf ${TMPDIR}
+"""
+
+def _vtpm_get_for_vm(host, vm_uuid):
```
stormi, line 55: This could/should be a method on the VM class.

tests/misc/test_vtpm_basic_operations.py
```diff
@@ -0,0 +1,117 @@
...
+    try:
+        res = host.xe('vtpm-list', {'vm': vm_uuid})
+        return res.splitlines()[0].split(':')[1].strip()
+    except IndexError:
+        return None
+
+def _vtpm_create(host, vm_uuid):
```
stormi, line 62: This could/should be a method on the VM class.

tescande:
Not added to `jobs.py` yet on purpose

tests/misc/test_vtpm_basic_operations.py
```diff
@@ -0,0 +1,118 @@
...
+    if not vm.is_halted():
+        force = vm.is_paused() or vm.is_suspended()
+        vm.shutdown(verify=True, force=force)
+
+    vtpm_uuid = _vtpm_get_for_vm(host, vm.uuid)
+    # Fail if vtpm-destroy fails since it's not possible to attach more than
+    # one vtpm to a vm
+    _vtpm_destroy(host, vtpm_uuid)
+
+@pytest.mark.big_vm
```
stormi, line 86: This is probably not the marker you want. As described in pytest.ini, big_vm means "I want this test to be run with as big a VM as possible".

tests/misc/test_vtpm_basic_operations.py
```diff
@@ -0,0 +1,118 @@
...
+@pytest.mark.usefixtures("host_at_least_8_3")
+def test_vtpm(host, uefi_vm):
```
stormi, line 88: I think this tests does too much for a single test. 

I'd separate this in two tests:
* test_create_and_destroy, where you just create a vTPM then destroy it. So that these operations are tested on their own.
* test_vtpm_in_vm which requires a new unix_vm_with_vtpm fixture (which in turn would require the existing unix_vm and uefi_vm fixtures, it's a cascade), the fixture handling the creation and destruction of the vTPM outside the test code itself.

stormi (changes requested):

tests/misc/test_vtpm_basic_operations.py
tescande: Done

tests/misc/test_vtpm_basic_operations.py
tescande: Done

tests/misc/test_vtpm_basic_operations.py
tescande: Done

tests/misc/test_vtpm_basic_operations.py
```diff
@@ -0,0 +1,117 @@
...
+        res = host.xe('vtpm-list', {'vm': vm_uuid})
+        return res.splitlines()[0].split(':')[1].strip()
+    except IndexError:
+        return None
+
+def _vtpm_create(host, vm_uuid):
+    logging.info("Creating vtpm for vm %s" % vm_uuid)
+    return host.xe('vtpm-create', {'vm-uuid': vm_uuid})
+
+def _vtpm_destroy(host, vtpm_uuid):
```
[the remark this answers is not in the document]
tescande: It is not possible to create a vTPM device without associating it to a VM, so VM class makes more sense imo.

tests/misc/test_vtpm_basic_operations.py
```diff
@@ -0,0 +1,118 @@
...
+
+    # vm must be shutdown to attach a vtpm
+    if not vm.is_halted():
+        force = vm.is_paused() or vm.is_suspended()
+        vm.shutdown(verify=True, force=force)
+
+    vtpm_uuid = _vtpm_get_for_vm(host, vm.uuid)
+    # Fail if vtpm-destroy fails since it's not possible to attach more than
+    # one vtpm to a vm
+    _vtpm_destroy(host, vtpm_uuid)
```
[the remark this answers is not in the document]
tescande: Done. A new fixture has been added for that purpose: `unix_vm_with_vtpm`

tests/misc/test_vtpm_basic_operations.py
tescande: Done

tests/misc/test_vtpm_basic_operations.py
tescande: Changed it to `small_vm` has agreed together

lib/vm.py
```diff
@@ -401,6 +401,26 @@ def get_all_efi_bins(self):
...
+
+    def create_vtpm(self):
+        logging.info("Creating vtpm for vm %s" % self.uuid)
+        return self.host.xe('vtpm-create', {'vm-uuid': self.uuid})
+
+    def destroy_vtpm(self):
+        vtpm_uuid = self.get_vtpm()
+        if vtpm_uuid is not None:
+            logging.info("Destroying vtpm %s" % vtpm_uuid)
+            self.host.xe('vtpm-destroy', {'uuid': vtpm_uuid}, force=True)
```
benjamreis, line 422: Why is `--force` needed to destroy a vTPM?

tests/vtpm/test_vtpm_basic_operations.py
```diff
@@ -0,0 +1,107 @@
...
+import pytest
+import lib.commands as commands
+
+# These test are basic tests for vTPM devices.
+#   - Create / Destroy a vTPM device against a VM
+#   - Do some basic encryption tests using tpm2-tools package
+#
+# Requirements:
+# - an XCP-ng host >= 8.3
+# - a Debian based VM
```
benjamreis, line 11:
```suggestion
# - a Debian based UEFI VM
```

tests/vtpm/test_vtpm_basic_operations.py
```diff
@@ -0,0 +1,107 @@
...
+    echo "Should not succeed"
+    exit 1
+fi
+
+rm -rf ${TMPDIR}
+"""
+
+@pytest.mark.small_vm
+@pytest.mark.usefixtures("host_at_least_8_3")
+def test_create_and_destroy_vtpm(host, uefi_unix_halted_vm):
```
benjamreis, line 67: `host` isn't used in the test and is alreay required by `host_at_least_8_3` above.

tests/vtpm/test_vtpm_basic_operations.py
```diff
@@ -0,0 +1,107 @@
...
+    pkg_mgr = vm.detect_package_manager()
+    if pkg_mgr != PackageManagerEnum.APT_GET:
+        logging.error("This test requires a Debian based VM")
+        assert False
+
+    # Basic TPM2 tests with tpm2-tools
+    try:
+        logging.info("Running TPM2 test script on the VM")
+        out = vm.execute_script(vtpm_signing_test_script)
+        logging.info("*****\n%s\n*****" % out)
```
benjamreis, line 104: Do we want this to be printed no matter what? or do we want to have it only at debug level?

benjamreis (changes requested):

lib/vm.py
tescande: This is required by the `vtpm-destroy` command because of possible data loss...

tests/vtpm/test_vtpm_basic_operations.py
tescande: Yes, debug level is enough. Fixed

tests/vtpm/test_vtpm_basic_operations.py
```diff
@@ -0,0 +1,107 @@
...
+    # unix_vm_with_vtpm fixture starts the VM and stops it at teardown
+    # No need to call vm.start() here
+
+    # Also used to retreive the VM IP address
+    vm.wait_for_os_booted()
+
+    pkg_mgr = vm.detect_package_manager()
+    if pkg_mgr != PackageManagerEnum.APT_GET:
+        logging.error("This test requires a Debian based VM")
+        assert False
```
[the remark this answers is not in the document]
tescande: That's a question I asked myself indeed... I chose to check that in the test itself since the test also controls the test script executed on the VM. If the script evolves to support other distros it will be easier (and more clear) to modify this check here rather that in the fixture. My 2 cts...

tests/vtpm/test_vtpm_basic_operations.py
```diff
@@ -0,0 +1,107 @@
...
+
+    assert vtpm_uuid is not None
+
+    logging.info("vtpm created with uuid: %s" % vtpm_uuid)
+
+    vm.destroy_vtpm()
+
+@pytest.mark.small_vm
+@pytest.mark.usefixtures("host_at_least_8_3")
+def test_vtpm(host, unix_vm_with_vtpm):
```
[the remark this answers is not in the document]
tescande: Fixed

tests/vtpm/test_vtpm_basic_operations.py
tescande: Fixed

tests/vtpm/test_vtpm_basic_operations.py
tescande: Fixed

tests/vtpm/test_vtpm_basic_operations.py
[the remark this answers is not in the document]
benjamreis: Fine by me

benjamreis (approved):

tests/vtpm/test_vtpm_basic_operations.py
```diff
@@ -0,0 +1,105 @@
...
+@pytest.mark.small_vm
+@pytest.mark.usefixtures("host_at_least_8_3")
+def test_vtpm(unix_vm_with_vtpm):
+    global vtpm_signing_test_script
+    vm = unix_vm_with_vtpm
+
+    # unix_vm_with_vtpm fixture starts the VM and stops it at teardown
+    # No need to call vm.start() here
+
+    # Also used to retreive the VM IP address
```
stormi, line 92:
```suggestion
    # Also used to retrieve the VM IP address
```

tests/vtpm/test_vtpm_basic_operations.py
```diff
@@ -0,0 +1,105 @@
...
+    vm.wait_for_os_booted()
```
stormi, line 93: This is the responsibility of the fixture which prepares the VM for the test and starts the VM, not a responsibility of the test itself.

tests/vtpm/test_vtpm_basic_operations.py
```diff
@@ -0,0 +1,105 @@
...
+
+    pkg_mgr = vm.detect_package_manager()
+    if pkg_mgr != PackageManagerEnum.APT_GET:
+        pytest.fail("This test requires a Debian based VM")
```
stormi, line 97: This is still test setup, not actual test code, so should be in a fixture, e.g. `running_debian_vm_with_vtpm`

tests/vtpm/test_vtpm_basic_operations.py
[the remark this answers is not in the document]
stormi: Looks like I made the same comment :o).

tests/vtpm/test_vtpm_basic_operations.py
stormi: By the way, this would fit perfectly in a fixture which does both:
* check the package manager
* install the TPM2 tools in the VM

tests/vtpm/test_vtpm_basic_operations.py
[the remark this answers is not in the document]
stormi: > Looks like I made the same comment :o).

But I have more arguments. Detecting the distro and installing the tpm2 tools are definitely setup material, which we avoid doing in test code. There are many ways to write setup/teardown code in pytest, but fixtures are generally the most convenient. It's not a problem if your fixture is only used by one test. You can define it just before the test rather than in conftest.py if you think it makes it more readable.

tests/vtpm/test_vtpm_basic_operations.py
stormi: I made the same comment. Was the change really pushed?

stormi:
By the way, the changes to `jobs.py` (discussed earlier) and `vm-data.py-dist` (discussed earlier too, but I forgot to mention the `-dist` file is the source for our `vm-data.py` files) still have to be updated with a new vtpm job.

lib/vm.py
```diff
@@ -401,6 +401,26 @@ def get_all_efi_bins(self):
 
         return binaries
 
+    def get_vtpm(self):
```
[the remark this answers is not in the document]
tescande: Correct, will fix it

conftest.py
```diff
@@ -277,6 +277,28 @@ def running_vm(imported_vm):
...
+    if vm.is_paused():
+        logging.info("VM is paused. Unpause it")
+        vm.unpause()
+    elif vm.is_suspended():
+        logging.info("VM is suspended. Resume it")
+        vm.resume()
+
+    if vm.is_running():
+        vm.wait_for_os_booted()
+        vm.shutdown(verify=True)
```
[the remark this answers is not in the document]
tescande: Works for me

lib/vm.py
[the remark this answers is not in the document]
tescande: Agree, much better

tests/vtpm/test_vtpm_basic_operations.py
tescande: Ok, will be moved to `unix_vm_with_vtpm` fixture if that's ok for you

tescande:
> By the way, `jobs.py` (discussed earlier) and `vm-data.py-dist` (discussed earlier too, but I forgot to mention the `-dist` file is the source for our `vm-data.py` files) still have to be updated with a new `vtpm` job.

I'll add `jobs.py` for the sake of check jobs but the UEFI Debian 12 image has to added to the single VM list in `vm_data.py`.

stormi:
>

stormi:
> I'll add `jobs.py` for the sake of check jobs but the UEFI Debian 12 image has to added to the single VM list in `vm_data.py`.

Not for the sake of check jobs. For your new test to be usable in our CI at all, because all test suites are started by `./jobs.py`. And don't forget `vm_data-dist.py` which is the reference file for building your own `vm-data.py`.

Then yes, we'll have to update our internal CI `vm-data.py`, of course, or the job will fail.

lib/vm.py
```diff
@@ -401,6 +401,26 @@ def get_all_efi_bins(self):
...
+        if not self.is_uefi:
+            return None
+
+        try:
+            res = self.host.xe('vtpm-list', {'vm': self.uuid})
+            return res.splitlines()[0].split(':')[1].strip()
+        except IndexError:
+            return None
```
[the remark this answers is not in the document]
tescande: Agree. Fixed

tests/vtpm/conftest.py
```diff
@@ -0,0 +1,28 @@
+import pytest
+
+@pytest.fixture(scope='module')
+def uefi_unix_halted_vm(uefi_vm, unix_vm, halted_vm):
+    yield uefi_vm
```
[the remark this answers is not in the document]
tescande: Agree, `halted_vm` fixture removed

tests/vtpm/test_vtpm_basic_operations.py
tescande: Fixed

tests/vtpm/test_vtpm_basic_operations.py
tescande: Moved to a new fixture called `unix_vm_with_tpm2_tools` that does a snapshot, starts the VM, and install `tpm2-tools`

tests/vtpm/test_vtpm_basic_operations.py
tescande: Actually created a new fixture called `unix_vm_with_tpm2_tools`

tests/vtpm/test_vtpm_basic_operations.py
```diff
@@ -0,0 +1,105 @@
...
+
+    # Basic TPM2 tests with tpm2-tools
+    try:
+        logging.info("Running TPM2 test script on the VM")
+        out = vm.execute_script(vtpm_signing_test_script)
+        logging.debug("*****\n%s\n*****" % out)
+    except commands.SSHCommandFailed as e:
+        pytest.fail("%s" % str(e))
```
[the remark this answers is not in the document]
tescande: Fair enough

tests/vtpm/test_vtpm_basic_operations.py
```diff
@@ -0,0 +1,105 @@
...
+    vm.wait_for_os_booted()
+
+    pkg_mgr = vm.detect_package_manager()
+    if pkg_mgr != PackageManagerEnum.APT_GET:
+        pytest.fail("This test requires a Debian based VM")
+
+    # Basic TPM2 tests with tpm2-tools
+    try:
+        logging.info("Running TPM2 test script on the VM")
+        out = vm.execute_script(vtpm_signing_test_script)
```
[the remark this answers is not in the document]
tescande: Fixed by moving the package installation to the fixture called `unix_vm_with_tpm2_tools`

tests/vtpm/test_vtpm_basic_operations.py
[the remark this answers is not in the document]
tescande: Moved the package manager test into the new fixture `unix_vm_with_tpm2_tools`. It also tries to install tps2-tools on RedHat based distros since tests show that this is working on a centos 8...

tests/vtpm/conftest.py
```diff
@@ -0,0 +1,28 @@
...
+
+@pytest.fixture(scope='module')
+def unix_vm_with_vtpm(uefi_unix_halted_vm):
+    vm = uefi_unix_halted_vm
+    vtpm_uuid = None
+
+    if vm.get_vtpm() is None:
+        vtpm_uuid = vm.create_vtpm()
+
+    snapshot = vm.snapshot()
```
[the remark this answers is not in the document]
tescande: It doesn't really matter since destroying a snapshot doesn't destroy the associated vtpm that would have been created after the snapshot. You have to destroy the vtpm anyway.

And it's even worth because if you do:
```
snapshot = vm.snapshot()
vm.create_vtpm()
<...>
snapshot.destroy()
```
Then the vtpm will still exist and be attached to the VM but won't be usable by the VM anymore...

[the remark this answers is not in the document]
tescande: And yes, there's a missing `snapshot.revert()`

tests/vtpm/conftest.py
```diff
@@ -0,0 +1,28 @@
...
+
+    vm.start()
+
+    yield vm
+
+    # Tear down
+    if not vm.is_halted():
+        vm.shutdown(verify=True)
```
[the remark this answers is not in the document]
tescande: Works for me

tests/vtpm/conftest.py
```diff
@@ -0,0 +1,28 @@
...
+
+    snapshot.destroy()
```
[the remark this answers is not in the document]
tescande: Yeah, right... Sorry for that

tests/vtpm/conftest.py
```diff
@@ -0,0 +1,28 @@
+import pytest
+
+@pytest.fixture(scope='module')
+def uefi_unix_halted_vm(uefi_vm, unix_vm, halted_vm):
+    yield uefi_vm
+
+@pytest.fixture(scope='module')
+def unix_vm_with_vtpm(uefi_unix_halted_vm):
```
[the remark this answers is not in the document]
tescande: I did it another way :)

There's the fixture `uefi_unix_halted_vm` which is used by the `unix_vm_with_vtpm` fixture responsible for creating / destroying the vtpm, and finally the `unix_vm_with_tpm2_tools` that does the snapshot and installs tpm2-tools.

It makes more sense to me that the one installing tpm2-tools is the one that takes care of the snapshot...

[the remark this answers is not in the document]
stormi: It's still a fixture which does too much and thus is vulnerable to failures: 
* if `vm.start()` fails, or we have a timeout when waiting for the OS to boot and IP address to be available, the setup of the fixture fails and the teardown is not executed: revert is not called and the snapshot is not destroyed. 
* If the package manager is not a supported one, the test fails and the teardown is not executed: the VM is not shutdown, snapshot is not reverted
* Same if a network or mirror error causes the installation of tpm2-tools to fail
* During teardown, if vm shutdown fails (sometimes a VM can't shut down cleanly), the snapshot is not reverted nor destroyed.

The principles of nested fixtures I tried hard to put forward (an example in mattermost + my preview review comments) were not respected here.

stormi (changes requested):

stormi:
> And don't forget `vm_data-dist.py` which is the reference file for building your own `vm-data.py`.

This.

tests/vtpm/conftest.py
```diff
@@ -0,0 +1,62 @@
...
+@pytest.fixture(scope='module')
+def started_unix_vm_with_vtpm(unix_vm_with_vtpm):
+    vm = unix_vm_with_vtpm
+
+    try:
+        vm.start()
+        vm.wait_for_os_booted()
+    except Exception:
+        vm.shutdown(force=True, verify=True)
+        raise Exception(f"Failed to start VM")
```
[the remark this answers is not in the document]
tescande: Moving `wm.start()` out of the try/except block seems enough as `vm.shutdown()` would be executed only if `vm.start() succeeds and `vm.wait_for_os_booted()` fails.

The reason I used `raise` is because while testing this code, nothing was happening upon return from `vm.shutdown()`. Maybe a `pytest.fail()` would leave the original Exception message in log ?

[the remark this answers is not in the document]
stormi: We could try (to be tested)

stormi (approved):
I think we're good!

stormi:
I transmitted the informations to Gaël for him to add this new job to our CI.

tescande:
\o/

PR #161, opened 2023-08-23
Use random hostname starting with xcp-ng-

gthvn1:
This allows to have different hostname when installing new VM.


stormi:
This looks good. One possible improvements would be making it an option of the script, for users who want to control the hostname, and only falling back to random name when no hostname is specified.

gthvn1:
@stormi , @Wescoeur I added an option to specify the VM hostname if the user wants to. When nothing is passed the hostname is generated using the pattern "xcp-ng-XXXXX" where X is a lowercase character choosen randomly. I just tested both use cases.

scripts/install_xcpng.py
```diff
@@ -174,6 +181,11 @@ def main():
              "In case of a restore, specify the version of the installer."
     )
     parser.add_argument("--installer", help="URL of the installer")
+    parser.add_argument(
+        "--vm_hostname",
```
stormi, line 185: Underscores in parameter names are not really common.  This would be `--vm-hostname` , but at some point in the future we might extend the script to handle physical hosts, and so `vm` would become outdated. Maybe `--set-hostname`, or just `--hostname` (I know there's a little risk of confusion with the host parameter, but with the help text I think it's good enough)

gthvn1: Yes I was concerned by the confusion between `hostname` and `host` (hostname_or_ip).
What about `--xcp-ng-hostname`  ?

stormi: It's equally confusing, because the other hostname is also a XCP-ng hostname :D. I thought of `--target-hostname` but it may be a bit long.

gthvn1: Let's try target hostname. Too long is better than too confusing :)

stormi: Works for me. And with argparse you can easily add a `-h` short version of the parameter.

stormi: Or  `-t` if `-h` conflicts with the help :grimacing:

stormi (approved):

scripts/install_xcpng.py
```diff
@@ -41,12 +43,16 @@ def generate_boot_conf(directory, installer, action):
...
-def generate_answerfile(directory, installer, hostname_or_ip, action, hdd):
+def generate_answerfile(directory, installer, hostname_or_ip, target_hostname, action, hdd):
     pxe = pxe_address()
     password = host_data(hostname_or_ip)['password']
     cmd = ['openssl', 'passwd', '-6', password]
     res = subprocess.run(cmd, stdout=subprocess.PIPE)
     encrypted_password = res.stdout.decode().strip()
+    if target_hostname is None:
+        target_hostname = "xcp-ng-" + "".join(
+            random.choice(string.ascii_lowercase) for i in range(5)
```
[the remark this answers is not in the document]
gthvn1: Nice. I will do that.

Wescoeur (approved):

scripts/install_xcpng.py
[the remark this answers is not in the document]
gthvn1: Cool. But in my hurry I merged and forgot to push the new random.choices that I prepared...:-(.
 Sorry Yann ;)

PR #165, opened 2023-09-12
Add a test for plugin smartctl

gthvn1:
This test checks that the plugin returns a JSON string as expected.

### DIFF
--- a/tests/xapi-plugins/plugin_smartctl/test_smartctl.py
+++ b/tests/xapi-plugins/plugin_smartctl/test_smartctl.py
@@ -1,0 +1,21 @@
+import json
+import pytest
+
+# Requirements:
+# From --hosts parameter:
+# - host(A1): first XCP-ng host >= 8.3.
+
+def _call_plugin(host, fn):
+    ret = host.call_plugin("smartctl.py", fn)
+    try:
+        json.loads(ret)
+    except ValueError:
+        pytest.fail("JSON string was expected but ValueError was raised")
+
+@pytest.mark.usefixtures("host_at_least_8_3")
+def test_smartctl_information(host):
+    _call_plugin(host, "information")
+
+@pytest.mark.usefixtures("host_at_least_8_3")
+def test_smartctl_health(host):
+    _call_plugin(host, "health")


gthvn1:
It is not ready to merge. We are waiting for [xcp-ng-xapi-plugins PR37](https://github.com/xcp-ng/xcp-ng-xapi-plugins/pull/37)

tests/xapi-plugins/plugin_smartctl/test_smartctl.py
```diff
@@ -0,0 +1,18 @@
+# Requirements:
+# From --hosts parameter:
+# - host(A1): first XCP-ng host > 8.2.
+
+import json
+
+def _call_plugin(host, fn):
+    ret = host.call_plugin("smartctl.py", fn)
+    try:
+        json.loads(ret)
```
gthvn1, line 10: can probably just do: `assert json.loads(ret)`

tests/xapi-plugins/plugin_smartctl/test_smartctl.py
```diff
@@ -0,0 +1,18 @@
...
+    except ValueError:
+        assert False, "JSON string was expected but ValueError was raised"
```
benjamreis, line 12: We prefer to use `pystest.fails`

benjamreis (changes requested):

tests/xapi-plugins/plugin_smartctl/test_smartctl.py
gthvn1: or use pytest.fails :+1:

benjamreis (approved):

tests/xapi-plugins/plugin_smartctl/test_smartctl.py
```diff
@@ -0,0 +1,19 @@
+# Requirements:
+# From --hosts parameter:
+# - host(A1): first XCP-ng host > 8.2.
```
stormi, line 3: 8.2.1 > 8.2, so this doesn't work. >= 8.3 would be more accurate.

stormi, line 3: Putthing the "Requirements" comment before the imports may be a good idea, but it's not consistent with the rest of the test scripts, so such a file layout change should not happen in this PR.

gthvn1: done

gthvn1: done

tests/xapi-plugins/plugin_smartctl/test_smartctl.py
gthvn1: Yes I triggered the error to check that it is working as expected :+1:

tests/xapi-plugins/plugin_smartctl/test_smartctl.py
```diff
@@ -0,0 +1,19 @@
+import json
+import pytest
+
+# Requirements:
+# From --hosts parameter:
+# - host(A1): first XCP-ng host >= 8.3.
```
stormi, line 6: I should have seen it earlier. @benjamreis should have seen it earlier too: you need to decorate your tests (or a test class containing the tests) with:

`@pytest.mark.usefixtures("host_at_least_8_3")`

This will skip the tests on a host which is < 8.3

stormi (approved):

gthvn1:
Just need to wait that smartctl plugin is available in 8.3. Otherwise test is ready I think.

PR #166, opened 2023-09-12
Ignore GPG check for install or upgrade in the script

gduperrey:
As a temporary fix, we added an option to the script install_xcpng.py to ignore the gpg check in the install and upgrade mode. It will last until we have a signed netinstall iso to create the repo on the PXE.

### DIFF
--- a/scripts/install_xcpng.py
+++ b/scripts/install_xcpng.py
@@ -43,7 +43,7 @@ is_default=1
 {rt}
 """)
 
-def generate_answerfile(directory, installer, hostname_or_ip, target_hostname, action, hdd):
+def generate_answerfile(directory, installer, hostname_or_ip, target_hostname, action, hdd, netinstall_gpg_check):
     pxe = pxe_address()
     password = host_data(hostname_or_ip)['password']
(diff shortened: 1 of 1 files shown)


stormi:
To me, this should be an optional parameter to the script, not something that applies to all installs.

Moreover, the option is specific to XCP-ng 8.2 (and earlier) and doesn't exist in XCP-ng 8.3, IIRC (there are different options), so this should be made clear in the commit message and the script's new option.

scripts/install_xcpng.py
```diff
@@ -43,7 +43,7 @@ def generate_boot_conf(directory, installer, action):
 {rt}
 """)
 
-def generate_answerfile(directory, installer, hostname_or_ip, target_hostname, action, hdd):
+def generate_answerfile(directory, installer, hostname_or_ip, target_hostname, action, hdd, ngc):
```
stormi, line 46: I would prefer if we named the variable `netinstall_gpg_check` everywhere. Sure, its a bit long, and you may have to split some lines to avoid pycodestyle complaints, but it's more readable.

stormi (approved):

PR #169, opened 2023-09-27
Only test HSTS on 8.3 and above

gthvn1:
HSTS is available only for host with version greater or equal to 8.3.

### DIFF
--- a/tests/misc/test_file_server.py
+++ b/tests/misc/test_file_server.py
@@ -1,9 +1,10 @@
+import pytest
 import subprocess
 
 # These tests are meant to test an host fileserver behavior.
 #
 # Requirements:
-# - an XCP-ng host >= 8.2 with latest updates.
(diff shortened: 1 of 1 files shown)


tests/misc/test_file_server.py
```diff
@@ -1,10 +1,12 @@
+import pytest
 import subprocess
 
 # These tests are meant to test an host fileserver behavior.
 #
 # Requirements:
 # - an XCP-ng host >= 8.2 with latest updates.
 #   The host must be configured with `website-https-only` set to true (which is the default config).
+# - an XCP-ng host >= 8.3 for testing HSTS
```
stormi, line 9: I'd group the added line with line 7. For example:

```suggestion
# - an XCP-ng host >= 8.2 with latest updates (and >= 8.3 for the HSTS test).
#   The host must be configured with `website-https-only` set to true (which is the default config).
```

benjamreis (approved):

stormi (approved):

PR #170, opened 2023-09-28
Adapt tests for IPv6 hosts

benjamreis:
- add brackets around IP in URI when necessary
- ignore tests not suporting IPv6

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -8,6 +8,7 @@ import lib.config as global_config
 
 from lib.common import wait_for, vm_image, is_uuid
 from lib.common import setup_formatted_and_mounted_disk, teardown_formatted_and_mounted_disk
+from lib.netutil import is_ipv6
 from lib.pool import Pool
 from lib.vm import VM
 from lib.xo import xo_cli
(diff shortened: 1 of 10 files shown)


conftest.py
```diff
@@ -176,6 +177,17 @@ def xfail_on_xcpng_8_3(host, request):
     if host.xcp_version >= version.parse("8.3"):
         request.node.add_marker(pytest.mark.xfail)
 
+@pytest.fixture(scope='session')
+def host_no_ipv6(host):
+    if is_ipv6(host.hostname_or_ip):
+        pytest.skip(f"This test requires an IPv4 XCP-ng")
+
+@pytest.fixture(scope='function')
+def xfail_on_xcpng_ipv6(host, request):
```
stormi, line 186: We don't need this fixture anymore, right?

tests/storage/moosefs/test_moosefs_sr.py
```diff
@@ -31,6 +31,7 @@ def test_create_moosefs_sr_without_mfsmount(self, host, moosefs_device_config):
             sr.destroy()
             assert False, "MooseFS SR creation should failed!"
 
+    @pytest.mark.usefixtures("host_no_ipv6")
```
stormi, line 34: Everytime we use this fixture, I think there should be a comment explaining why the test has to be skipped in IPv6 mode.

conftest.py
benjamreis: Right, I thought about remove and then thought it might be handy in the future.
I can remove it if you'd like.

stormi:
LGTM :+1:

stormi (approved):

PR #172, opened 2023-09-28
Wait for 10 minutes before the 5th sr-destroy attempt

stormi:
We already have code which retries a failed sr-destroy, when the cause is remaining hidden VDIs which haven't been garbage-collected yet, and it's enough in most cases, but not always.

Sometimes, the force-gc that is triggered by SM before the delete operation is not enough. VDIs that look like they should be removed remain, and are removed only a bit later, when the normal garbage collector runs.

This may be related to the fact that the forced GC operation doesn't coalesce VDIs, or to a special case where a recently coalesced VDI is temporarily protected against deletion.

We add a 10 minutes wait before the 5th sr-destroy attempt, to give the normal garbage collector enough time to run (it should run 5 minutes after the last operation that creates a need for the GC to run).

### DIFF
--- a/lib/sr.py
+++ b/lib/sr.py
@@ -1,4 +1,5 @@
 import logging
+import time
 
 import lib.commands as commands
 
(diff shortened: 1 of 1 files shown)


stormi:
The conditions for entering the new code path are not that easy to reproduce, so let's be extra careful in the review.

benjamreis (approved):

gthvn1 (approved):

PR #176, opened 2023-10-25
Adapt UEFI tests to latest 8.3

stormi:


stormi:
@benjamreis @ydirson Everything but the last commit is ready for review.

Context: we had already adapted somehow the tests to XCP-ng 8.3 (with a few tests known to fail and marked as `xfail` for us to solve later / pending the then upcoming XAPI changes) and to the use of varstored instead of uefistored, but this was before things changed again (for the better) in XAPI and varstored.

I can provide more details if you want to better understand/check each test.

stormi:
I strongly suggest to review commit per commit. I've tried not to mix changing code with moving code or renaming files in the same commit, to make it easier to review.

tests/uefi_sb/test_uefistored_sb.py
```diff
@@ -210,71 +209,5 @@ def setup_and_cleanup(self, uefi_vm_and_snapshot):
...
-            ok = True
-            try:
-                vm.set_efi_var(auth.name, auth.guid,
-                               EFI_AT_ATTRS_BYTES, auth.auth_data)
-            except SSHCommandFailed:
-                ok = False
-
-            if (should_succeed and not ok) or (ok and not should_succeed):
-                raise AssertionError('Failed to set {} {}'.format(i, auth.name))
+        from .utils import _test_key_exchanges
```
ydirson, line 219: It is generally considered bad practice to import anywhere else than at top of file

lib/pool.py
```diff
@@ -97,6 +99,7 @@ def first_shared_sr(self):
         return None
 
     def save_uefi_certs(self):
+        assert self.master.xcp_version < version.parse("8.3"), "function only relevant on XCP-ng 8.2"
```
ydirson, line 102: Messages showing the reasons would be more useful.  Maybe "8.3+ only saves custom certificates, not all of them" here?

lib/pool.py
```diff
@@ -156,12 +160,18 @@ def restore_uefi_certs(self):
...
 
     def clear_uefi_certs(self):
+        assert self.master.xcp_version < version.parse("8.3"), "function only relevant on XCP-ng 8.2"
         logging.info('Clearing pool UEFI certificates in XAPI and on hosts disks')
         self.master.ssh(['secureboot-certs', 'clear'])
         # remove files on each host
         for host in self.hosts:
             host.ssh(['rm', '-f', f'{host.varstore_dir()}/*'])
 
+    def clear_custom_uefi_certs(self):
```
ydirson, line 170: Adding docstrings to those funcs could help understand the difference with `clear_uefi_certs`

tests/uefi_sb/test_cert_inheritance.py
```diff
@@ -265,7 +265,7 @@ def keys_auths_for_joined_host(self, host, hostB1):
         from packaging import version
         version_str = "8.3"
         if not hostB1.xcp_version >= version.parse(version_str):
-            pytest.skip(f"This test requires a second XCP-ng pool with version >= {version_str}")
+            pytest.fail(f"This test requires a second XCP-ng pool with version >= {version_str}")
```
ydirson: `pytest.fail` says a correct test detected an error in the product. To show it as an error in the test it should raise an exception instead.

lib/efi.py
```diff
@@ -442,6 +443,8 @@ def esl_from_auth_bytes(auth: bytes) -> bytes:
     """
     return auth[auth.index(EFI_CERT_X509_GUID):]
 
+def get_md5sum_from_auth(auth):
+    return hashlib.md5(esl_from_auth_file(auth)).hexdigest()
```
ydirson, line 447: This (and a few more funcs) are reported as moved code by `git show --color-moved --color-moved-ws=allow-indentation-change`, it could be more clear to separate code reorg from new tests

ydirson (commented):
other comment not possible to put in proper place otherwise: colons at end of commit summary make it look bad in e.g. `git log --oneline` and `gitk`

jobs.py
```diff
@@ -128,7 +128,7 @@
         "params": {
             "--vm": "single/small_vm_efitools",
         },
-        "paths": ["tests/uefi_sb/test_auth_var.py", "tests/uefi_sb/test_secure_boot.py"],
+        "paths": ["tests/uefi_sb/test_auth_var.py", "tests/uefi_sb/test_uefistored_sb.py"],
```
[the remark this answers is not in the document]
benjamreis: Agreed, uefistored == 8.2 and varstored == 8.3+ is not obvious at all.

stormi:
> other comment not possible to put in proper place otherwise: colons at end of commit summary make it look bad in e.g. `git log --oneline` and `gitk`

Indeed. This was not intentional.

stormi:
@ydirson @benjamreis all comments resolved.

tests/uefi_sb/conftest.py
```diff
@@ -1,8 +1,11 @@
 import logging
 import pytest
 
-@pytest.fixture(scope='module')
+from packaging import version
+
+@pytest.fixture(scope='module') # FIXME: why not package scope?
 def pool_without_uefi_certs(host):
+    assert host.xcp_version < version.parse("8.3"), "fixture only relevant on XCP-ng 8.2"
```
benjamreis, line 8: Why not use `host_less_than_8_3`?

stormi: I haven't though of it, to be honest :)

However, there's a difference:
- `host_less_than_8_3` is designed so that we are not punished if we call a set of tests containing some that are not usable: the incompatible tests are just skipped.
- here I want that any developer who tries to use this fixture in a test that is going to run on XCP-ng 8.3+ gets an immediate error.

ydirson (approved):
1.

> Drop TestPoolToDiskCertInheritanceAtXapiStart. There is no such propagation anymore.

So we don't need this test in 8.2 any more either?

2. "move functions out of test_cert_inheritance.py" is still more involved to follow than it could be, but LGTM nevertheless, barring comparisons that constant scrolling back and forth made tedious (would make it easier: separate commits for func renames, scope changes getting rid of `self`, plus a preliminary "move code" so the latter does not move lines)

stormi:
> > Drop TestPoolToDiskCertInheritanceAtXapiStart. There is no such propagation anymore.
> 
> So we don't need this test in 8.2 any more either?

We have never needed this test in 8.2, actually. It was made to test XCP-ng 8.3, before the behaviour there changed again. I'll update the commit message.

lib/pool.py
```diff
@@ -97,6 +99,24 @@ def first_shared_sr(self):
         return None
 
     def save_uefi_certs(self):
+        """
+        Save UEFI certificates in order to restore them later. XCP-ng 8.2 only.
+
+        This method was developed for XCP-ng 8.2, because many secureboot tests were dependent
+        on the initial state of the pool certificates, due to how certificates propagate.
+        Also, there were no certificates installed by default (except PK) on XCP-ng 8.2, and
+        we tried to be gentle and restore the initial state after the tests.
```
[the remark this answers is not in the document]
stormi: Done

PR #180, opened 2023-11-15
Test linstor diskless resource

benjamreis:
Make sure a resource remains diskless upon VM lifecycle

New optional arg to VM.vdi_uuids(): sr_uuid to get
vdis on a specific SR

### DIFF
--- a/lib/basevm.py
+++ b/lib/basevm.py
@@ -27,12 +27,20 @@ class BaseVM:
     def name(self):
         return self.param_get('name-label')
 
-    def vdi_uuids(self):
+    def vdi_uuids(self, sr_uuid=None):
         output = self._disk_list()
         vdis = []
         for line in output.splitlines():
             vdis.append(line.split(',')[0])
-        return vdis
+
+        if sr_uuid is None:
+            return vdis
+
+        vdis_on_sr = []
+        for vdi in vdis:
+            if self.get_vdi_sr_uuid(vdi) == sr_uuid:
+                vdis_on_sr.append(vdi)
+        return vdis_on_sr
 
     def destroy_vdi(self, vdi_uuid):
         self.host.xe('vdi-destroy', {'uuid': vdi_uuid})
--- a/tests/storage/linstor/test_linstor_diskless_resources.py
+++ b/tests/storage/linstor/test_linstor_diskless_resources.py
@@ -1,0 +1,62 @@
+import pytest
+
+# Requirements:
+# From --hosts parameter:
+# - host(A1): first XCP-ng host >= 8.2 with an additional unused disk for the SR.
+# From --vm parameter
+# - A VM to import to the LINSTOR SR
+# And:
+# - access to XCP-ng RPM repository from hostA1
+
+def _get_diskful_hosts(host, controller_option, volume_name):
+    # Find host where volume is diskless
+    # | {volume_name} | {host} | 7017 | Unused | Ok    |   UpToDate | 2023-10-24 18:52:05 |
+    lines = host.ssh([
+        "linstor", controller_option, "resource", "list",
+        "|", "grep", volume_name, "|", "grep", "UpToDate"
+    ]).splitlines()
+    diskfuls = []
+    for line in lines:
+        hostname = line.split('|')[2].strip()
+        diskfuls += hostname
+    return diskfuls
+
+def _ensure_resource_remain_diskless(host, controller_option, volume_name, diskless):
+    diskfuls = _get_diskful_hosts(host, controller_option, volume_name)
+    for diskless_host in diskless:
+        assert diskless_host.param_get("name-label").lower() not in diskfuls
+
+class TestLinstorDisklessResource:
+    @pytest.mark.small_vm
+    def test_diskless_kept(self, host, linstor_sr, vm_on_linstor_sr):
+        vm = vm_on_linstor_sr
+        vdi_uuids = vm.vdi_uuids(sr_uuid=linstor_sr.uuid)
+        vdi_uuid = vdi_uuids[0]
+        assert vdi_uuid is not None
+
+        controller_option = "--controllers="
+        for member in host.pool.hosts:
+            controller_option += f"{member.hostname_or_ip},"
+
+        # Get volume name from VDI uuid
+        # "xcp/volume/{vdi_uuid}/volume-name": "{volume_name}"
+        output = host.ssh([
+            "linstor-kv-tool", "--dump-volumes", "-g", "xcp-sr-linstor_group_thin_device",
+            "|", "grep", "volume-name", "|", "grep", vdi_uuid
+        ])
+        volume_name = output.split(': ')[1].split('"')[1]
+
+        diskfuls = _get_diskful_hosts(host, controller_option, volume_name)
+        diskless = []
+        for member in host.pool.hosts:
+            if member.param_get("name-label").lower() not in diskfuls:
+                diskless += [member]
+        assert diskless
+
+        # Start VM on host with diskless resource
+        vm.start(on=diskless[0].uuid)
+        vm.wait_for_os_booted()
+        _ensure_resource_remain_diskless(host, controller_option, volume_name, diskless)
+
+        vm.shutdown(verify=True)
+        _ensure_resource_remain_diskless(host, controller_option, volume_name, diskless)


tests/storage/linstor/test_linstor_diskless_resources.py
```diff
@@ -0,0 +1,70 @@
...
+            "linstor-kv-tool", "--dump-volumes", "-u", "localhost", "-g",
+            "xcp-sr-linstor_group_thin_device", "|", "grep", "volume-name", "|",
+            "grep", f"{vdi_uuid}"
+        ])
+        volume_name = res.split(': ')[1].split('"')[1]
+
+        # Find host where volume is diskless
+        # ┊ {volume_name} ┊ {host} ┊ 7017 ┊ Unused ┊ Ok    ┊   Diskless ┊ 2023-10-24 18:52:05 ┊
+        res = host.ssh([
+            "linstor", "resource", "list", "|", "grep", f"{volume_name}",
```
stormi, line 35:
```suggestion
            "linstor", "resource", "list", "|", "grep", volume_name,
```

tests/storage/linstor/test_linstor_diskless_resources.py
```diff
@@ -0,0 +1,70 @@
...
+                vdi_uuid = v
+                break
+        assert vdi_uuid is not None
+
+        # Get volume name from VDI uuid
+        # "xcp/volume/{vdi_uuid}/volume-name": "{volume_name}"
+        res = host.ssh([
+            "linstor-kv-tool", "--dump-volumes", "-u", "localhost", "-g",
+            "xcp-sr-linstor_group_thin_device", "|", "grep", "volume-name", "|",
+            "grep", f"{vdi_uuid}"
```
stormi, line 28:
```suggestion
            "grep", vdi_uuid
```

tests/storage/linstor/test_linstor_diskless_resources.py
```diff
@@ -0,0 +1,70 @@
...
+        hostname = res.split('┊')[2].strip()
+
+        # Start VM on host with diskless resource
+        vm.start(on=hostname)
+        vm.wait_for_os_booted()
+
+        # Ensure resource remains diskless
+        res = host.ssh([
+            "linstor", "resource", "list", "|", "grep", f"{volume_name}",
+            "|", "grep", "Diskless", "|", "grep", f"{hostname}"
```
stormi, line 47:
```suggestion
            "linstor", "resource", "list", "|", "grep", volume_name,
            "|", "grep", "Diskless", "|", "grep", hostname
```

tests/storage/linstor/test_linstor_diskless_resources.py
```diff
@@ -0,0 +1,70 @@
...
+        ]).splitlines()
```
stormi, line 48: Why do you split the lines if all you need is testing for the string to be nonempty?

tests/storage/linstor/test_linstor_diskless_resources.py
```diff
@@ -0,0 +1,70 @@
...
+        assert len(res) > 0
+
+        vm.shutdown(verify=True)
+
+        # Ensure resource remains diskless
+        res = host.ssh([
+            "linstor", "resource", "list", "|", "grep", f"{volume_name}",
+            "|", "grep", "Diskless", "|", "grep", f"{hostname}"
```
stormi, line 56:
```suggestion
            "linstor", "resource", "list", "|", "grep", volume_name,
            "|", "grep", "Diskless", "|", "grep", hostname
```

tests/storage/linstor/test_linstor_diskless_resources.py
```diff
@@ -0,0 +1,70 @@
...
+        res = host.ssh([
+            "linstor-kv-tool", "--dump-volumes", "-u", "localhost", "-g",
+            "xcp-sr-linstor_group_thin_device", "|", "grep", "volume-name", "|",
+            "grep", f"{vdi_uuid}"
+        ])
+        volume_name = res.split(': ')[1].split('"')[1]
+
+        # Find host where volume is diskless
+        # ┊ {volume_name} ┊ {host} ┊ 7017 ┊ Unused ┊ Ok    ┊   Diskless ┊ 2023-10-24 18:52:05 ┊
+        res = host.ssh([
```
stormi, line 34: In other tests and methods, we mostly named `res` a SSHResult object. When the result is a string or a list, I think we should use other names. `output` is a good name for a string. `lines` may be a good name for a list of lines. This comment applies to the other occurrences of res in this file.

tests/storage/linstor/test_linstor_diskless_resources.py
```diff
@@ -0,0 +1,70 @@
...
+
+        # Start VM on host with diskless resource
+        vm.start(on=hostname)
+        vm.wait_for_os_booted()
+
+        # Ensure resource remains diskless
+        res = host.ssh([
+            "linstor", "resource", "list", "|", "grep", f"{volume_name}",
+            "|", "grep", "Diskless", "|", "grep", f"{hostname}"
+        ]).splitlines()
```
stormi, line 67: Same comment as above. I don't think the call to splitlines is necessary.

tests/storage/linstor/test_linstor_diskless_resources.py
```diff
@@ -0,0 +1,70 @@
...
+        assert len(res) > 0
+
+        vm.shutdown(verify=True)
```
stormi, line 70: I think this should be in a teardown function or in the teardown part of a fixture. Otherwise, any failure somewhere in the test may leave the VM running. I know it will be destroyed afterwards, but there's no guarantee that we don't need the VM for another test later (stopping on the first error is not mandatory to execute our tests, it's just the default behaviour).

tests/storage/linstor/test_linstor_diskless_resources.py
```diff
@@ -0,0 +1,70 @@
...
+    @pytest.mark.small_vm
+    def test_diskless_kept(self, host, linstor_sr, vm_on_linstor_sr):
+        vm = vm_on_linstor_sr
+        vdi_uuids = vm.vdi_uuids()
+        vdi_uuid = None
+        for v in vdi_uuids:
+            if vm.get_vdi_sr_uuid(v) == linstor_sr.uuid:
+                vdi_uuid = v
+                break
+        assert vdi_uuid is not None
```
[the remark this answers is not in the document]
benjamreis: What if the VM has vdi on other SR?

tests/storage/linstor/test_linstor_diskless_resources.py
benjamreis: Right, but this applies to a lot of our tests with a VM, all of them shut it down at the end no?

stormi: True. Not all of them, but many do indeed.

tests/storage/linstor/test_linstor_diskless_resources.py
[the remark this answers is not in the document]
stormi: My bad. I misread what `all_vdis_on_sr` does. I suggest instead to add an optional `sr_uuid` - or `sr` - parameter to `BaseVM.vdi_uuids()`

tests/storage/linstor/test_linstor_diskless_resources.py
```diff
@@ -0,0 +1,70 @@
...
+        ])
+        volume_name = output.split(': ')[1].split('"')[1]
+
+        # Find host where volume is diskless
+        # ┊ {volume_name} ┊ {host} ┊ 7017 ┊ Unused ┊ Ok    ┊   Diskless ┊ 2023-10-24 18:52:05 ┊
+        output = host.ssh([
+            "linstor", "resource", "list", "|", "grep", volume_name,
+            "|", "grep", "Diskless"
+        ]).splitlines()[0]
+        hostname = output.split('┊')[2].strip()
```
stormi, line 38: `output` is a good name of a string, but here I would name it `lines`, as suggested previously.

tests/storage/linstor/test_linstor_diskless_resources.py
```diff
@@ -0,0 +1,70 @@
...
+        output = host.ssh([
+            "linstor", "resource", "list", "|", "grep", volume_name,
+            "|", "grep", "Diskless", "|", "grep", hostname
+        ])
+        assert output != ""
+
+        vm.shutdown(verify=True)
+
+        # Ensure resource remains diskless
+        output = host.ssh([
```
stormi, line 54: `output` is not used here

tests/storage/linstor/test_linstor_diskless_resources.py
```diff
@@ -0,0 +1,78 @@
...
+        lines = host.ssh([
+            "linstor", controller_option, "resource", "list",
+            "|", "grep", volume_name, "|", "grep", "UpToDate"
+        ]).splitlines()
+        diskfulls = []
+        for line in lines:
+            hostname = line.split('|')[2].strip()
+            diskfulls += hostname
+
+        diskless = None
```
stormi, line 46: `diskfull_hosts` and `diskless_host` would make it clearer that these variables contain either lists of Host objects, or one Host object. This is just a suggestion.

tests/storage/linstor/test_linstor_diskless_resources.py
```diff
@@ -0,0 +1,78 @@
...
+        for h in host.pool.hosts:
+            if h.param_get("name-label").lower() not in diskfulls:
+                diskless = h
```
stormi, line 49: No `break`? There can be only one host where the volume is diskless? If so, should you get all such hosts and assert there's only one?

benjamreis:
needs to wait for https://github.com/xcp-ng/sm/pull/48

Nambrok (approved):

lib/basevm.py
```diff
@@ -26,12 +26,20 @@ def param_remove(self, param_name, key, accept_unknown_key=False):
...
         vdis = []
         for line in output.splitlines():
             vdis.append(line.split(',')[0])
-        return vdis
+
+        if sr_uuid is None:
+            return vdis
+
+        vdis_on_sr = []
+        for v in vdis:
```
Wescoeur, line 39: Why not `vdi` directly? Simple letters are not easy to grep in a script.

tests/storage/linstor/test_linstor_diskless_resources.py
```diff
@@ -0,0 +1,74 @@
...
+        ])
+        volume_name = output.split(': ')[1].split('"')[1]
+
+        # Find host where volume is diskless
+        # | {volume_name} | {host} | 7017 | Unused | Ok    |   UpToDate | 2023-10-24 18:52:05 |
+        lines = host.ssh([
+            "linstor", controller_option, "resource", "list",
+            "|", "grep", volume_name, "|", "grep", "UpToDate"
+        ]).splitlines()
+        diskfulls = []
```
Wescoeur, line 37: `s/diskfulls/diskfuls/`

tests/storage/linstor/test_linstor_diskless_resources.py
```diff
@@ -0,0 +1,74 @@
...
+class TestLinstorDisklessResource:
+    @pytest.mark.small_vm
+    def test_diskless_kept(self, host, linstor_sr, vm_on_linstor_sr):
+        vm = vm_on_linstor_sr
+        vdi_uuids = vm.vdi_uuids(sr_uuid=linstor_sr.uuid)
+        vdi_uuid = vdi_uuids[0]
+        assert vdi_uuid is not None
+
+        controller_option = "--controllers="
+        for h in host.pool.hosts:
```
Wescoeur, line 20: `host` instead?

tests/storage/linstor/test_linstor_diskless_resources.py
```diff
@@ -0,0 +1,74 @@
...
+
+        # Get volume name from VDI uuid
+        # "xcp/volume/{vdi_uuid}/volume-name": "{volume_name}"
+        output = host.ssh([
+            "linstor-kv-tool", "--dump-volumes", "-g", "xcp-sr-linstor_group_thin_device",
+            "|", "grep", "volume-name", "|", "grep", vdi_uuid
+        ])
+        volume_name = output.split(': ')[1].split('"')[1]
+
+        # Find host where volume is diskless
```
Wescoeur, line 31: We can have many diskless, so >= 1 host(s). We must probably use an array of diskless.

tests/storage/linstor/test_linstor_diskless_resources.py
```diff
@@ -0,0 +1,74 @@
...
+        # Ensure resource remains diskless
+        lines = host.ssh([
+            "linstor", controller_option, "resource", "list",
+            "|", "grep", volume_name, "|", "grep", "UpToDate"
+        ]).splitlines()
+        diskfulls = []
+        for line in lines:
+            hostname = line.split('|')[2].strip()
+            diskfulls += hostname
+        assert diskless.param_get("name-label").lower() not in diskfulls
```
Wescoeur, line 61: Use a for loop here on a diskless array. ;)

Wescoeur (changes requested):

benjamreis:
Comments addressed - Manual run succesful!

Wescoeur (approved):

PR #181, opened 2023-12-05
Add a filesystem diff script

tescande:
This script checks for file system tree differences between 2 XCP-ng hosts. The comparison can be made against a reference installation (let say a fresh ISO install) and another XCP-ng of the same version but obtained by upgrading an existing installation, or to check differences between ISO and Net installations.

Supported options are:

  --reference-host REF_HOST, -r REF_HOST
         The XCP-ng host used as reference

  --test-host TEST_HOST, -t TEST_HOST
         The XCP-ng host to be tested after install or upgrade

  --save-reference SAVE_REF, -s SAVE_REF
         Save filesystem information of the reference host to a file

  --load-reference LOAD_REF, -l LOAD_REF
         Load reference filesystem information from a file

Option -r HOST is used to pass the hostname or ip address of the XCP-ng host used as reference.

With -s FILENAME, the reference host filesyistem information is saved to the file FILENAME passed as parameter.

Instead of -r HOST, one can use the reference file created with -s by using the option -f FILENAME.

The host to be tested is specified with the option -t HOST.

### DIFF
--- a/scripts/xcpng-fs-diff.py
+++ b/scripts/xcpng-fs-diff.py
@@ -37,6 +37,8 @@ import argparse
 import sys
 import subprocess
 import json
+import tempfile
+import os
 from fnmatch import fnmatch
 from enum import Enum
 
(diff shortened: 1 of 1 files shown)


stormi:
I don't have time to review this but I dont want to hold it, so let's merged with the existing two approvals.

PR #182, opened 2023-12-08
WIP test_host_evacuate: check for IP in xenstore first

ydirson:
This can show when after migration an IP address has been published to Xenstore, but XAPI misses it nevertheless.

TODO: check impact on other code paths, notably VM migration. Especially those code paths that do not update VM.host, which will need the same kind of hack we had to do here.


tests/migration/test_host_evacuate.py
```diff
@@ -34,6 +34,7 @@ def _host_evacuate_test(source_host, dest_host, network_uuid, vm, expect_error=F
         source_host.xe('host-evacuate', args)
         wait_for(lambda: vm.all_vdis_on_host(dest_host), "Wait for all VDIs on destination host")
         wait_for(lambda: vm.is_running_on_host(dest_host), "Wait for VM to be running on destination host")
+        vm.host = dest_host
```
gthvn1, line 37: it is not related to wait for xenstore VM IP right? It looks like a bug fix when evacuation is done on the destination host currently self.host is not updated no? If it is the case I think that it is a fix that should be done in its own PR.

ydirson: It is indeed required so the upcoming check for VM IP runs `xenstore` commands on the destination host, although it is a workaround that can likely be used for other things. Could be moved in a separate commit for more clarity.

gthvn1: ok I see :+1:

benjamreis: perhaps a vm.resident_on would be better.

lib/vm.py
```diff
@@ -123,6 +136,7 @@ def wait_for_os_booted(self):
         # waiting for the IP:
         # - allows to make sure the OS actually started (on VMs that have the management agent)
         # - allows to store the IP for future use in the VM object
+        wait_for(self.try_get_ip_xenstore, "Wait for Xenstore VM IP")
         wait_for(self.try_get_and_store_ip, "Wait for VM IP")
```
benjamreis, line 140: Does this solve the issue? IIUC the IP does appear in the xenstore but not seen by XAPI so the 2nd wait would fail anyway right?

stormi: The goal is not to solve the issue, but to make it more visible when it happens.

tests/migration/test_host_evacuate.py
stormi: Agreed with @benjamreis here: as I said earlier, vm.host really should be vm.pool and will probably be in the future. It is not where the VM is running, it's the pool master of the pool to which the VM belongs, and is badly named.

You need to use the `VM.get_residence_host` method in `VM.try_get_ip_xenstore`.

stormi (changes requested):
Marking "changes requested".

tests/migration/test_host_evacuate.py
ydirson: so we're indeed evacuating the host into another pool?

gthvn1: > Agreed with @benjamreis here: as I said earlier, vm.host really should be vm.pool and will probably be in the future. It is not where the VM is running, it's the pool master of the pool to which the VM belongs, and is badly named.

And in the test we are sure that we migrate a VM from a slave to the master (or another slave) during the evacuate? Because if the evacuation is done from the master than we need to update vm.host in this case. Or we are sure that it cannot happens.

gthvn1: Oh wait, in fact no because evacuate from master doesn't mean that we need to change the master... maybe.

stormi: What I mean is VM.host is not where the VM is currently executed. There is no guarantee about this, and it usually just is the master host, regardless of where the VM is (note also that the concept of residence for a VM on a host doesn't work for a VM which is turned off and is on a shared SR. When you start it from XAPI, any host of the pool can become temporarily where it resides). And thus we shouldn't change the value in the test, and use `VM.get_residence_host` in order to find which host's xenstore we need to check.

> so we're indeed evacuating the host into another pool?

No, host evacuate is intra-pool and doesn't involve storage migration.

lib/vm.py
```diff
@@ -71,6 +71,19 @@ def reboot(self, force=False, verify=False):
...
+        domid = self.param_get("dom-id")
+        result = self.host.ssh(["xenstore", "read", f"/local/domain/{domid}/attr/vif/0/ipv4/0"],
+                               check=False, simple_output=False)
+
+        # An IP that starts with 169.254. is not a real routable IP.
+        # VMs may return such an IP before they get an actual one from DHCP.
+        if result.returncode != 0 or result.stdout.startswith('169.254.'):
+            return False
+        else:
+            logging.info("Xenstore VM IP: %s" % result.stdout)
```
stormi, line 84: stdout must be stripped, because it ends with an newline character.

stormi:
I see there were pushes on this PR. Do you want a re-review?

stormi (commented):
It looks useful, but I think we'll need to take IPv6 into account now, before merging.
A VM could be IPv6-only.

benjamreis:
> It looks useful, but I think we'll need to take IPv6 into account now, before merging. A VM could be IPv6-only.

IPv6 can be found in xenstore under `0/ipv6/`, `1/ipv6/`... and shouldn't start with `fe80:` as they're link local addresses that are not reachable from the host.

PR #183, opened 2023-12-12
vtpm: Fix installation of tpm2-tools package using apt-get

tescande:
apt-get needs to update its package index files before installing tpm2-tools as it might try to install outdated dependencies. This patch fixes this issue by first running 'apt-get update' before 'apt-get install'.


stormi (approved):
I'm not 100% for embedding two commands in the same SSH with `&&`, but strong opposition either.

gduperrey (approved):
Runs fine :)

PR #186, opened 2024-01-15
run on ISO SR as well

benjamreis:
Waiting for upstream fix to run on ISO SR


benjamreis:
Replaced by https://github.com/xcp-ng/xcp-ng-tests/pull/231

PR #187, opened 2024-01-19
Only connect XO for migration tests

stormi:
Maintaining a connection to XO adds causes for test failure, so let's not connect it when not strictly required, that is, in the only place where we only use it currectly: live migration tests.


benjamreis (commented):
LGTM, let's put it to the test :)

stormi:
Tested on `main` and `storage-reboots` jobs: looks good.

benjamreis (approved):

PR #191, opened 2024-01-22
Move linstor tests to their own jobs

stormi:


jobs.py
```diff
@@ -120,6 +121,68 @@
...
+            "A second pool with at least 1 host and a SR to receive VMs.",
+            "A small VM that can be imported on the SRs.",
+        ],
+        "nb_pools": 2,
+        "params": {
+            "--vm": "single/small_vm",
+            "--sr-disk": "auto",
+        },
+        "paths": ["tests/storage/linstor"],
+        "markers": "",
```
benjamreis, line 156: is the empty field necessary?

benjamreis: Why `markers` and not `name_filter`

Just asking its non blocking the merge IMHO

stormi: Because I considered the use of `markers` to be common and "right", and the use of `name_filter` more of a hack to be used sparingly.

benjamreis (approved):

gduperrey (approved):

Wescoeur (approved):

PR #192, opened 2024-01-22
Avoid having two ISO SRs at the same time during tests

stormi:
In ISO SR tests, there are package-scoped fixtures for SRs, which is what we usually do in other storage tests. But here, in the same package, we test ISO SR, NFS ISO SR, and CIFS ISO SR.

As a result we start creating a new SR when the previous one hasn't been teared down yet.

By giving these fixtures the `module` scope, we ensure teardown happens at the end of each test file. The counterpart is that we may create and destroy the same SR twice, successively, because there are two test files per SR type. However, since the second file is each time dedicated to tests with reboot, and that we usually run these in a separate job, this doesn't add any overhead to the current test jobs.

### DIFF
--- a/tests/storage/iso/conftest.py
+++ b/tests/storage/iso/conftest.py
@@ -13,7 +13,7 @@ def create_local_iso_sr(host, location):
     }
     return host.sr_create('iso', "ISO-local-SR-test", device_config, verify=True)
 
-@pytest.fixture(scope='package')
+@pytest.fixture(scope='module')
 def local_iso_sr(host, formatted_and_mounted_ext4_disk):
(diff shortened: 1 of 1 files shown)


stormi:
To reviewers: it's not so easy to describe the issue, so if you want I can demonstrate it after the daily meeting today.

benjamreis (approved):

gduperrey (approved):

gduperrey:
I just ran the test on 8.2 :white_check_mark:

stormi:
@gthvn1 too late :)

gthvn1:
who is silent consents :)

PR #195, opened 2024-02-06
xcp-ng-diff.py: Add support for packages diff

tescande:


scripts/xcpng-fs-diff.py
```diff
@@ -228,9 +244,14 @@ def compare_data(ref, test, show_diff):
                 continue
 
             if ref_data[dtype][file] != test_data[dtype][file]:
-                print("{} differs: {}".format(dtype, file))
-                if show_diff:
-                    remote_diff(ref_host, test_host, file)
+                print("{} differs: {}".format(dtype, file), end='')
+                if dtype == 'package':
+                    print(" (ref={}, test={})".format(ref_data[dtype][file],
```
stormi, line 213: We tend to prefer f-strings rather than .format (I should have mentioned it in a previous review)

scripts/xcpng-fs-diff.py
```diff
@@ -228,9 +244,14 @@ def compare_data(ref, test, show_diff):
...
+                                                      test_data[dtype][file]))
+                else:
+                    print("")
+                    if show_diff:
+                        remote_diff(ref_host, test_host, file)
```
stormi, line 218: If show_diff is false, you just print a newline? I haven't checked the rest of the script so maybe it makes sense.

scripts/xcpng-fs-diff.py
tescande: OK, let me add a commit to change all `print` in the script

stormi (approved):

ydirson (approved):
At first I was not sure why you chose "data" over "files", to me that still looked like "comparing files".  Maybe expand the note, telling that you're renaming functions that will take a larger role than just dealing with files.

benjamreis (approved):

tescande:
> At first I was not sure why you chose "data" over "files", to me that still looked like "comparing files". Maybe expand the note, telling that you're renaming functions that will take a larger role than just dealing with files.

Sorry @ydirson, completely missed your comment, it was not intentionally ignored. I merged it directly as I saw everybody's approval.

PR #196, opened 2024-02-14
Fix evacuation with network tests

benjamreis:
- fail 'second_network' fixture if network is *NOT* present on master
- the error condition for evacuate are configured on the receiving host


benjamreis:
Tested and validated on IPv4&6

stormi (approved):

PR #197, opened 2024-02-16
Add support to filter VMs based on host version

tescande:


jobs.py
```diff
@@ -332,29 +332,41 @@
...
 
+    host_version = None
+    if hosts is not None:
+        try:
+            host = hosts.split(',')[0]
+            cmd = ["ssh", host, "lsb_release", "-sr"]
+            res = subprocess.run(cmd, capture_output=True, text=True)
+            if res.returncode == 0:
+                host_version = res.stdout.strip()
+        except Exception as e:
```
benjamreis, line 367: We assume all hosts provided have same versions, is it wanted?

tescande: Gael and Sam will confirm but yes, all hosts are supposed to run the same XCP-ng version.

vm_data.py-dist
```diff
@@ -27,3 +53,42 @@ VMS = {
...
+            res["single"][key] = xva
+
+    for key, vms in VMS["multi"].items():
+        res["multi"][key] = list()
+        for vm in vms:
+            xva = prepend_xva_location(vm, host_version)
+            if xva is not None:
+                res["multi"][key].append(xva)
+
+    return res
```
stormi, line 94: I'd rather have all this code directly in jobs.py, so that vm_data.py remains just about data.

vm_data.py-dist
```diff
@@ -1,7 +1,33 @@
 # All values below are left empty on purpose, for you to fill them with your VM references.
 # The VM references will be passed over by jobs.py to pytest through pytest's --vm parameter.
-# Consult README.md for accepted values.
-# Example value: "http://somewhere/images/small_vm.xva"
+
+# The base URL for your VM references must be specified as:
+XVA_LOCATION="http://somewhere.domain/"
```
stormi, line 5: I now remember why `XVA_LOCATION` was not defined in `vm_data.py-dist`: because I wanted to give the liberty to just make the file a simple dict, with all URLs repeated if needed, and maybe using XVAs downloaded from various places.

It's not a strong requirement, though, so if it makes it easier for us, I'm OK with enforcing the use of XVA_LOCATION.

tescande: So now XVA_LOCATION is used only for XVAs that are not URLs

stormi: I would have let XVA_LOCATION be specific to our instance of vm_data.py, and only define a dict in the `vm_data.py-dist`, but I'm not opposed to your change either.

stormi: Oh, actually, there is an issue with this change. Values can also be VM UUIDs.

stormi: So, in my opinion, it's best not to add any logic to `vm_data.py-dist`. `jobs.py` must expect a dict of URLs or UUIDs, and that's it. Defining a prefix for URLs is up to the user who makes their own `vm_data.py`.

tescande: As discussed, our version of `vm_data.py` needs some code to prepend `XVA_LOCATION` to all VM entries. And now, host version is matched using a regular expression,

jobs.py
```diff
@@ -347,14 +364,45 @@ def get_vm_or_vms_refs(handle):
         print("You need to update your local vm_data.py.")
         print("You may also bypass this error by providing your own --vm parameter(s).")
         sys.exit(1)
-    return VMS[category][key]
+
+    if type(VMS[category][key]) is list:
+        # Multi VMs
```
stormi, line 369: "Multi VMs" could be confused with the "multi" jobs. 

Suggestion:
```suggestion
        # Version-dependent VMs
```

stormi (approved):
LGTM apart from my last minor comment

jobs.py
tescande: Not really. `type(VMS[category][key]) is list` means `multi` category. Otherwise it's `single`. The version dependent VMS are filtered by `filter_vm()`.

PR #200, opened 2024-03-01
Don't skip empty lines...

gthvn1:
For some tricky cases we cannot skip empty line. For example when getting a key from varstored with ssh we can have within the key a sequence of two 0x0a bytes. The issue is that readline will split it with an empty line. If we skip it we basically remove one 0x0a from the key and then of course the checksum will failed.

### DIFF
--- a/lib/commands.py
+++ b/lib/commands.py
@@ -110,9 +110,8 @@ def _ssh(hostname_or_ip, cmd, check=True, simple_output=True, suppress_fingerpri
     stdout = []
     for line in iter(process.stdout.readline, b''):
         readable_line = line.decode(errors='replace').strip()
-        if readable_line != '':
-            stdout.append(line)
-            OUPUT_LOGGER.debug(readable_line)
+        stdout.append(line)
+        OUPUT_LOGGER.debug(readable_line)
     _, stderr = process.communicate()
     res = subprocess.CompletedProcess(ssh_cmd, process.returncode, b''.join(stdout), stderr)
 


lib/commands.py
```diff
@@ -110,9 +110,8 @@ def _ssh(hostname_or_ip, cmd, check=True, simple_output=True, suppress_fingerpri
     stdout = []
     for line in iter(process.stdout.readline, b''):
         readable_line = line.decode(errors='replace').strip()
-        if readable_line != '':
-            stdout.append(line)
-            OUPUT_LOGGER.debug(readable_line)
+        stdout.append(line)
+        OUPUT_LOGGER.debug(readable_line)
```
stormi, line 114: Maybe we'll want to skip the empty line for the logger, while appending it to stdout.

If we started skipping empty line, this is probably because it made the logs less readable in some cases. Or maybe there was a more important reason? Should we go back on this decision? We'll need to run various tests with debug output enabled to determine this.

gthvn1:
FYI test ran 200 times in a row this afternoon. Now as you said Sam we need to be sure that not skipping the line is ok.

stormi:
One test ran 200 times, but this change affects all tests, that was what I meant.

gthvn1:
> One test ran 200 times, but this change affects all tests, that was what I meant.

Yes sure. It was just to say that at least it fixes the issue I was looking for

lib/commands.py
benjamreis: I think the `OUTPUT_LOGGER` line should be put back in the if as indeed there's no point in logging empty lines.
As for the stdout, I guess we need to run various tests indeed to be sure we didn't break other tests not expecting empty lines in the result.

gthvn1: Except to log the "reality". For example in the case of the key from varstored you may ask "ok there is something missing". I mean if it doesn't pollute logs too much I prefer log it but I'm ok to put it back in the IF since your are 2 against 1 in favor to not log it :)

benjamreis: We need to see what the logs looks like as you propose, if its too much spacing or not.

stormi:
I may have another change tomorrow that will require running all tests. We'll probably combine both PRs in a temporary branch so that we can test both at the same time.

lib/commands.py
stormi: I ran many tests, and haven't seen adverse effects.

stormi (approved):

benjamreis (approved):

PR #201, opened 2024-03-01
zfs: don't name our pool "volume"

ydirson:


gthvn1 (approved):

stormi (approved):
The commit message doesn't say why this is needed, but I assume it better matches ZFS terminology.

ydirson:
> The commit message doesn't say why this is needed, but I assume it better matches ZFS terminology.

Exactly, volumes are a completely different thing than pools - the zfs driver does not use them that I know of, zfs-ng maps them to VDI.

PR #202, opened 2024-03-01
Tests for the SMAPIv3 ZFS-ng driver

ydirson:
Those tests are essentially adapted from those existing for SMAPIv1 SR drivers.

Note:
- this PR is stacked over #217
- those tests depend on the zfs-vol RPM from xcp-ng-rpms/xcp-ng-xapi-storage#3

### DIFF
--- a/tests/storage/zfsvol/conftest.py
+++ b/tests/storage/zfsvol/conftest.py
@@ -1,0 +1,35 @@
+import logging
+import pytest
+
+# Explicitly import package-scoped fixtures (see explanation in pkgfixtures.py)
+from pkgfixtures import host_with_saved_yum_state, sr_disk_wiped
+
+@pytest.fixture(scope='package')
+def host_with_zfsvol(host_with_saved_yum_state):
+    host = host_with_saved_yum_state
+    host.yum_install(['xcp-ng-xapi-storage-volume-zfsvol'])
+    yield host
+
+@pytest.fixture(scope='package')
+def zfsvol_sr(host, sr_disk_wiped, host_with_zfsvol):
+    """ A ZFS Volume SR on first host. """
+    device = '/dev/' + sr_disk_wiped
+    sr = host.sr_create('zfs-vol', "ZFS-local-SR-test", {'device': device})
+    yield sr
+    # teardown violently - we don't want to require manual recovery when a test fails
+    sr.forget()
+    host.ssh(["wipefs", "-a", device])
+
+@pytest.fixture(scope='module')
+def vdi_on_zfsvol_sr(zfsvol_sr):
+    vdi = zfsvol_sr.create_vdi('ZFS-local-VDI-test')
+    yield vdi
+    vdi.destroy()
+
+@pytest.fixture(scope='module')
+def vm_on_zfsvol_sr(host, zfsvol_sr, vm_ref):
+    vm = host.import_vm(vm_ref, sr_uuid=zfsvol_sr.uuid)
+    yield vm
+    # teardown
+    logging.info("<< Destroy VM")
+    vm.destroy(verify=True)
--- a/tests/storage/zfsvol/test_zfsvol_sr.py
+++ b/tests/storage/zfsvol/test_zfsvol_sr.py
@@ -1,0 +1,73 @@
+import logging
+import time
+import pytest
+
+from lib.commands import SSHCommandFailed
+from lib.common import wait_for, vm_image
+from tests.storage import vdi_is_open
+
+# Requirements:
+# - one XCP-ng host >= 8.3 with an additional unused disk for the SR
+# - access to XCP-ng RPM repository from the host
+
+pytestmark = pytest.mark.usefixtures("host_at_least_8_3")
+
+class TestZfsvolSRCreateDestroy:
+    """
+    Tests that do not use fixtures that setup the SR or import VMs,
+    because they precisely need to test SR creation and destruction,
+    and VM import.
+    """
+
+    def test_create_and_destroy_sr(self, host, sr_disk_wiped):
+        # Create and destroy tested in the same test to leave the host as unchanged as possible
+        sr = host.sr_create('zfs-vol', "ZFS-local-SR-test", {'device': '/dev/' + sr_disk_wiped}, verify=True)
+        # import a VM in order to detect vm import issues here rather than in the vm_on_xfs_fixture used in
+        # the next tests, because errors in fixtures break teardown
+        vm = host.import_vm(vm_image('mini-linux-x86_64-bios'), sr_uuid=sr.uuid)
+        vm.destroy(verify=True)
+        sr.destroy(verify=True)
+
+@pytest.mark.usefixtures("zfsvol_sr")
+class TestZfsvolVm:
+
+    @pytest.mark.xfail
+    @pytest.mark.quicktest
+    def test_quicktest(self, zfsvol_sr):
+        zfsvol_sr.run_quicktest()
+
+    @pytest.mark.small_vm # run with a small VM to test the features
+    @pytest.mark.big_vm # and ideally with a big VM to test it scales
+    def test_start_and_shutdown_VM(self, vm_on_zfsvol_sr):
+        vm = vm_on_zfsvol_sr
+        vm.start()
+        vm.wait_for_os_booted()
+        vm.shutdown(verify=True)
+
+    @pytest.mark.xfail # needs support for destroying snapshots
+    @pytest.mark.small_vm
+    @pytest.mark.big_vm
+    def test_snapshot(self, vm_on_zfsvol_sr):
+        vm = vm_on_zfsvol_sr
+        vm.start()
+        try:
+            vm.wait_for_os_booted()
+            vm.test_snapshot_on_running_vm()
+        finally:
+            vm.shutdown(verify=True)
+
+    # *** tests with reboots (longer tests).
+
+    @pytest.mark.reboot
+    @pytest.mark.small_vm
+    def test_reboot(self, vm_on_zfsvol_sr, host, zfsvol_sr):
+        sr = zfsvol_sr
+        vm = vm_on_zfsvol_sr
+        host.reboot(verify=True)
+        wait_for(sr.all_pbds_attached, "Wait for PBD attached")
+        # start the VM as a way to check that the underlying SR is operational
+        vm.start()
+        vm.wait_for_os_booted()
+        vm.shutdown(verify=True)
+
+    # *** End of tests with reboots
--- a/tests/storage/zfsvol/test_zfsvol_sr_crosspool_migration.py
+++ b/tests/storage/zfsvol/test_zfsvol_sr_crosspool_migration.py
@@ -1,0 +1,26 @@
+import pytest
+from tests.storage import cold_migration_then_come_back, live_storage_migration_then_come_back
+
+# Requirements:
+# From --hosts parameter:
+# - host(A1): first XCP-ng host >= 8.2 with an additional unused disk for the SR.
+# - hostB1: Master of a second pool. Any local SR.
+# From --vm parameter
+# - A VM to import to the ZFS-ng SR
+# And:
+# - access to XCP-ng RPM repository from hostA1
+
+@pytest.mark.xfail
+@pytest.mark.small_vm # run with a small VM to test the features
+@pytest.mark.big_vm # and ideally with a big VM to test it scales
+@pytest.mark.usefixtures("hostB1", "local_sr_on_hostB1")
+class Test:
+    def test_cold_crosspool_migration(self, host, hostB1, vm_on_zfsvol_sr,
+                                      zfsvol_sr, local_sr_on_hostB1):
+        cold_migration_then_come_back(
+            vm_on_zfsvol_sr, host, zfsvol_sr, hostB1, local_sr_on_hostB1)
+
+    def test_live_crosspool_migration(self, host, hostB1, vm_on_zfsvol_sr,
+                                      zfsvol_sr, local_sr_on_hostB1):
+        live_storage_migration_then_come_back(
+            vm_on_zfsvol_sr, host, zfsvol_sr, hostB1, local_sr_on_hostB1)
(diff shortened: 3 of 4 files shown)


stormi:
Looks like github doesn't send notifications when one marks a PR as ready for review :/

stormi (commented):
We'll need the tests to install the required driver before we can merge it and have it run among the other tests, otherwise existing jobs will fail. Or we can explicitly exclude them in jobs.py but will then need a reminder that we need to enable them when the rest is ready.

tests/storage/zfs_ng/test_zfs_sr.py
```diff
@@ -0,0 +1,69 @@
+import logging
+import time
+import pytest
+
+from lib.commands import SSHCommandFailed
+from lib.common import wait_for, vm_image
+from tests.storage import vdi_is_open
+
+# Requirements:
+# - one XCP-ng host >= 8.2 with an additional unused disk for the SR
```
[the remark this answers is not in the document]
stormi: And subsequently the whole test module would need to require the `host_at_least_8_3` fixture.

tests/storage/zfs_ng/test_zfs_sr.py
stormi, line 1: Since the driver name is  `zfs-ng`, I'd name the file `test_zfs_ng_sr.py`.

tests/storage/zfs_ng/test_zfs_sr.py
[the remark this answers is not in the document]
ydirson: The idea is to backport to 8.2 later (and indeed it mostly already works on 8.2)

ydirson:
> We'll need the tests to install the required driver before we can merge it and have it run among the other tests, otherwise existing jobs will fail.

Yes, but we're not at the point of building a package yet

ydirson:
Note last push does not include all the original commits, essentially because the current driver version has had snapshot destroy disabled.  Those included were reworked (or marked xfail) to take this into account.

tests/storage/zfsvol/conftest.py
```diff
@@ -0,0 +1,35 @@
...
+    host.ssh(['modprobe', 'zfs'])
+    yield host
+
+@pytest.fixture(scope='package')
+def zfsvol_sr(host, sr_disk_wiped, host_with_zfsvol):
+    """ A ZFS Volume SR on first host. """
+    sr = host.sr_create('zfs-vol', "ZFS-local-SR-test", {'device': '/dev/' + sr_disk_wiped})
+    yield sr
+    # teardown violently - we don't want to require manual recovery when a test fails
+    sr.forget()
```
gthvn1, line 20: I think that it won't work bcause unplug PBDs will probably failed. At least most of the time I tried it fails because PBD cannot be unplugged if there are some dataset (volume). But ok it may work  sometimes but when test fails I think that we will end with SR not forget. But ok it is better than nothing maybe ;)

tests/storage/zfsvol/test_zfsvol_sr.py
```diff
@@ -0,0 +1,72 @@
...
+
+    # *** tests with reboots (longer tests).
+
+    @pytest.mark.reboot
+    @pytest.mark.small_vm
+    def test_reboot(self, vm_on_zfsvol_sr, host, zfsvol_sr):
+        sr = zfsvol_sr
+        vm = vm_on_zfsvol_sr
+        host.reboot(verify=True)
+        wait_for(sr.all_pbds_attached, "Wait for PBD attached")
```
gthvn1, line 66: when testing that zpool is mounted after reboot I observed that it needs several minutes to be mounted. I mean once rebooted I logged into the host and running `zpool list` show me nothing and xapi was already started. So it seems that there is a few delay to be mounted. But wait_for is waiting 2 minutes so it should be ok but if this test failed we need to check if it is not an issue with timeout.

gthvn1 (approved):

tests/storage/zfsvol/conftest.py
ydirson: sr-forget has no problem with remaining VDI (as long as there is eg. no running VM using them), as opposed to sr-destroy, this is why it works much better in case of test failure than trying to be symmetric and calling `sr.destroy()`.

tests/storage/zfs_ng/conftest.py
```diff
@@ -0,0 +1,34 @@
+import logging
+import pytest
+
+# Explicitly import package-scoped fixtures (see explanation in pkgfixtures.py)
+from pkgfixtures import host_with_saved_yum_state, sr_disk_wiped
+
+@pytest.fixture(scope='package')
+def host_with_zfs(host_with_saved_yum_state):
```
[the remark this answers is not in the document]
ydirson: Now done, as `zfsvol`

tests/storage/zfs_ng/test_zfs_sr.py
```diff
@@ -0,0 +1,69 @@
...
+# - access to XCP-ng RPM repository from the host
+
+@pytest.mark.usefixtures("sr_disk_wiped")
+class TestZFSSRCreateDestroy:
```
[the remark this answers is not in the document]
ydirson: same, done

tests/storage/zfs_ng/test_zfs_sr.py
ydirson: same, done

tests/storage/zfsvol/test_zfsvol_sr.py
```diff
@@ -0,0 +1,72 @@
+import logging
+import time
+import pytest
+
+from lib.commands import SSHCommandFailed
+from lib.common import wait_for, vm_image
+from tests.storage import vdi_is_open
+
+# Requirements:
+# - one XCP-ng host >= 8.2 with an additional unused disk for the SR
```
stormi, line 10: Same comment as on the dependent "more tests" PR: minimum XCP-ng host version should be 8.3. And thus the whole tests should be requiring the `host_at_least_8_3` fixture:

```
```

gthvn1: If SR.forget is a success then PBD objects are removed and doesn't exist anymore.

tests/storage/zfsvol/conftest.py
gthvn1: I just tested a forget with a remaining VDI (no running VM using them because no running VM created) and I got:
```
[10:10 xcp-ng-opkwktak ~]# xe sr-forget uuid=fadda3f0-834a-50db-adce-b64520d2e82b
The SR is still connected to a host via a PBD. It cannot be destroyed or forgotten.
sr: fadda3f0-834a-50db-adce-b64520d2e82b (ZFS-VOL SR)
```

gthvn1: Note for @stormi if the sr.forget is a success then PBDs are deleted and don't exist anymore.

gthvn1: But ok I just look the code of "forget" in sr.py and it is doing the unplug of the PBDs so it should work :) ... and PBD are deleted after being unplugged and sr forgotten...

ydirson: > The SR is still connected to a host via a PBD

`xe pbd-unplug` must be performed before any sr-forget or sr-destroy

gthvn1: and it is the same for VDIs at least if they are not attached to a VM as @ydirson said earlier.

gthvn1: > > The SR is still connected to a host via a PBD
> 
> `xe pbd-unplug` must be performed before any sr-forget or sr-destroy

Yes and it is why I added that in fact `sr.forget()` is doing the unplugged (it is not clear just regarding the name of the function that suggest only a call to sr.forget()). So yes with the unplugged it should be ok.

gthvn1 (approved):

tests/storage/zfsvol/test_zfsvol_sr.py
gthvn1: > It's best to investigate why such a delay can happen, but if really needed `wait_for` can accept a different timeout value.

It is because after a reboot the zpool will be mounted when SR.attach is called the first time so it is not it is not necessarily at the very beginning of starting the xapi.

tests/storage/zfsvol/conftest.py
ydirson: Indeed this code was duly exercised by the various tests written before the code implements the features.

tests/storage/zfsvol/test_zfsvol_sr.py
ydirson: already noted above :)

ydirson:
Added the migration tests, marked as xfail, so they don't get forgotten.

tests/storage/zfsvol/test_zfsvol_sr.py
stormi: And what calls SR.attach? For all existing smapiv1 drivers, PBDs are plugged directly after boot so that the SR is immediately available to VMs.

gthvn1: It can be called at different places. If the PBD is plugged then SR.attach() has been called. PBDs are not plugged "directly" after the boot. There is the initialization of the XAPI itself that happens after systemd has started some services required by the XAPI. So after booting you have some delay that can be one or 2 minutes. For example I can ssh to the host while XAPI has not yet started... So ssh is started before the XAPI.

ydirson: Yes, `SR.attach` is the implementation of `xe pbd-plug`, the kind of stuff that XAPI will run when it starts up to bring the system up in sync with its db.

tests/storage/zfsvol/test_zfsvol_sr.py
```diff
@@ -0,0 +1,72 @@
...
+# - access to XCP-ng RPM repository from the host
+
+@pytest.mark.usefixtures("sr_disk_wiped")
```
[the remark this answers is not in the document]
ydirson: fixed

ydirson:
Rebased onto master after merge of #217

tests/storage/zfsvol/test_zfsvol_sr.py
stormi: In the tests, we wait for XAPI to be available, and PBDs are then plugged automatically, so I don't understand why we reach a timeout specifically for this SR driver and not others.

ydirson: What I see is a rather unreasonable waiting time to notice the host is back up after reboot:
```
Apr 02 16:24:48 INFO Wait for host up
Apr 02 16:29:49 INFO Wait for PBD attached
Apr 02 16:29:50 INFO Start VM
```

It seems like `ssh` stays stuck quite some time when launched while the IP is not reachable yet.  If we check the IP with `ping` first, the waiting time is much reduced:

```
Apr 02 16:47:34 INFO Wait for host down
Apr 02 16:47:38 INFO Wait for host up
Apr 02 16:47:38 INFO Wait for XAPI to be ready
Apr 02 16:49:10 INFO Wait for PBD attached
```

Will submit a separate PR for this.

ydirson: In fact there are cases where the nested host under test takes much more time to boot (can be see in the console in XO):
```
Apr 02 17:24:48 INFO Wait for host down
Apr 02 17:24:49 INFO Wait for ssh up on host
Apr 02 17:25:45 INFO Wait for XAPI to be ready
Apr 02 17:26:22 INFO Wait for PBD attached
```

So likely "not the fault of anything SMAPIv3"

ydirson: And maybe the root cause of the delay is not so obvious, or not single:
```
Apr 02 17:32:18 INFO Wait for host down
Apr 02 17:32:22 INFO Wait for host up
Apr 02 17:32:22 INFO Wait for ssh up on host
Apr 02 17:33:10 INFO Wait for XAPI to be ready
Apr 02 17:33:50 INFO Wait for PBD attached
```

ydirson: submitted #219 for that specific issue

ydirson:
Rebased, and updated `zfsvol_sr` to reflect recent changes in `SR.detach`.

ydirson:
Rebased on latest master, all remarks addressed

gthvn1 (approved):

tests/storage/zfsvol/test_zfsvol_sr.py
```diff
@@ -0,0 +1,73 @@
...
+
+from lib.commands import SSHCommandFailed
+from lib.common import wait_for, vm_image
+from tests.storage import vdi_is_open
+
+# Requirements:
+# - one XCP-ng host >= 8.3 with an additional unused disk for the SR
+# - access to XCP-ng RPM repository from the host
+
+pytestmark = pytest.mark.usefixtures("host_at_least_8_3")
```
stormi, line 13: Is assigning to the pytestmark variable useful? We never had to do so in all our other uses of `pytest.mark.usefixtures`.

stormi (approved):
LGTM, with a minor comment, and remarks/questions.

* Before merging, we'll wait for the updated xcp-ng-xapi-storage to be available in the 8.3/ci repo, to avoid breaking CI runs
* Are the new tests stable enough that they won't cause unexpected failures of the storage related jobs in CI from time to time, as far as we can tell?  Did you run the tests several times in a row without any "random" issues?

ydirson:
> LGTM, with a minor comment, and remarks/questions.
> 
>     * Before merging, we'll wait for the updated xcp-ng-xapi-storage to be available in the 8.3/ci repo, to avoid breaking CI runs
> 
>     * Are the new tests stable enough that they won't cause unexpected failures of the storage related jobs in CI from time to time, as far as we can tell?  Did you run the tests several times in a row without any "random" issues?

Yes last week I had tests run a full night without a single unexpected failure.

tests/storage/zfsvol/test_zfsvol_sr.py
ydirson: That's the official way to apply the feature to the whole module (it's not possible to use annotations for that).

Just tested against a 8.2 host, all tests properly skipped with:
```
SKIPPED [1] tests/storage/zfsvol/test_zfsvol_sr.py:22: This test requires an XCP-ng >= 8.3 host
```

stormi: Right. I thought I had used fixtures applied to whole modules before in xcp-ng-tests, but apparently I didn't.

PR #214, opened 2024-03-11
More tests for zfs-ng SMAPIv3 driver

ydirson:
Assorted tests:
- some show features working
- some show expected failures
- some are still exposing flaky result (very unsatisfactorily)

*Note*: based on top of #202

### DIFF
--- a/tests/storage/zfsvol/test_zfsvol_sr.py
+++ b/tests/storage/zfsvol/test_zfsvol_sr.py
@@ -20,13 +20,18 @@ class TestZfsvolSRCreateDestroy:
     """
 
     def test_create_and_destroy_sr(self, host, sr_disk_wiped):
-        # Create and destroy tested in the same test to leave the host as unchanged as possible
-        sr = host.sr_create('zfs-vol', "ZFS-local-SR-test", {'device': '/dev/' + sr_disk_wiped}, verify=True)
-        # import a VM in order to detect vm import issues here rather than in the vm_on_xfs_fixture used in
-        # the next tests, because errors in fixtures break teardown
-        vm = host.import_vm(vm_image('mini-linux-x86_64-bios'), sr_uuid=sr.uuid)
-        vm.destroy(verify=True)
-        sr.destroy(verify=True)
+        # Create and destroy tested in the same test to leave the host
+        # as unchanged as possible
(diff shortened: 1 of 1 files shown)


tests/storage/zfs_ng/test_zfs_sr.py
```diff
@@ -134,3 +134,29 @@ def test_vdi_and_snaps_destroy_intermediate_vdi(self, zfs_sr):
         # tell fixture we destroyed that one
         vdis[2:3] = []
         teardown_vdi_and_snaps_chain(zfs_sr, vdis)
+
+@pytest.fixture(scope='package')
```
stormi, line 138: A package-scoped fixture should be in the Package's `conftest.py`, if you expect to use it in several different test modules. Otherwise, `module` scope is enough.

tests/storage/zfs_ng/test_zfs_sr_crosspool_migration.py
```diff
@@ -0,0 +1,21 @@
+import pytest
+from tests.storage import cold_migration_then_come_back, live_storage_migration_then_come_back
+
+# Requirements:
+# From --hosts parameter:
+# - host(A1): first XCP-ng host >= 8.2 with an additional unused disk for the SR.
```
stormi, line 6: Depend on `host_at_least_8_3`?
```suggestion
# - host(A1): first XCP-ng host >= 8.3 with an additional unused disk for the SR.
```

(Same for `test_zfs(_ng)_sr_intrapool_migration

tests/storage/zfs_ng/test_zfs_sr_crosspool_migration.py
stormi, line 1: I'd name it `test_zfs_ng_sr_crosspool_migration.py` (same for the second file)

gthvn1 (approved):

tests/storage/zfs_ng/test_zfs_sr.py
ydirson: that was an outdated commit marked for rework, now reworked

tests/storage/zfs_ng/test_zfs_sr_crosspool_migration.py
ydirson: applied at module level in #202

PR #215, opened 2024-03-19
largeblock: add tests for largeblock driver

Nambrok:


benjamreis (approved):

tests/storage/largeblock/conftest.py
```diff
@@ -0,0 +1,24 @@
+import logging
+import pytest
+
+@pytest.fixture(scope='package')
+def largeblock_sr(host, sr_disk):
+    """ An LARGEBLOCK SR on first host. """
```
stormi, line 6:
```suggestion
    """ A LARGEBLOCK SR on first host. """
```

tests/storage/largeblock/test_largeblock_sr.py
```diff
@@ -0,0 +1,67 @@
+import pytest
+from lib.common import wait_for, vm_image
+from tests.storage import try_to_create_sr_with_missing_device, vdi_is_open
+
+# Requirements:
+# - one XCP-ng host with an additional unused disk for the SR
```
stormi, line 6: And this disk must support 4K, right?

stormi (changes requested):
If we were to merge this as is this would break several test jobs that currently run on disks that don't support 4K blocs. On the other hand, if we were to run test jobs on a host whose extra disk requires 4K blocs, it's many other test jobs which would likely fail.

To handle this, here's one possibility of solution:
* create new, dedicated jobs for 4K storage in `jobs.py`, jobs that will have to run on specific hardware in jenkins
* create a `sr_4k_disk` fixture along with a new `--sr-4k-disk` parameter for pytest, and make the test depend on it. This way, if we someday have both 512k and 4K disks available on test hosts, we can run both types of jobs.

I'm not fond of the solution, so better ideas welcome.

benjamreis (approved):
There's no migration tests for this driver, is it wanted?

lib/host.py
```diff
@@ -397,14 +397,22 @@ def disks(self):
     def disk_is_available(self, disk):
         return len(self.ssh(['lsblk', '-n', '-o', 'MOUNTPOINT', '/dev/' + disk]).strip()) == 0
 
-    def available_disks(self):
+    def available_disks(self, blocksize="512"):
```
benjamreis, line 400: a size being a string looks weird to me.
I'd make an int and use `str(blocksize)`, or even better `int(sec_size)`, when comparing to the return of `lsblk`

benjamreis (approved):

benjamreis (approved):

conftest.py
```diff
@@ -227,6 +235,21 @@ def sr_disk(request, host):
             f"disk or block device {disk} is either not present or already used on master host"
     yield disk
 
+@pytest.fixture(scope='session')
+def sr_disk_4k(request, host):
+    disk = request.param
+    if disk == "auto":
+        logging.info(">> Check for the presence of a free 4KiB block device on the master host")
+        disks = host.available_disks(4096)
+        assert len(disks) > 0, "a free disk 4KiB device is required on the master host"
```
stormi, line 244:
```suggestion
        assert len(disks) > 0, "a free 4KiB block device is required on the master host"
```

conftest.py
```diff
@@ -227,6 +235,21 @@ def sr_disk(request, host):
...
+        disk = disks[0]
+        logging.info(f">> Found free disk device(s) on hostA1: {' '.join(disks)}. Using {disk}.")
+    else:
+        logging.info(f">> Check that disk or block device {disk} is available on the master host")
+        assert disk in host.available_disks(4096), \
+            f"block device {disk} is either not present or already used on master host"
```
stormi, line 250: The assert text should state what is expected, not what the problem is. This problem was already present in `sr_disk`.

tests/storage/largeblock/test_largeblock_sr.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def test_create_sr_with_missing_device(self, host):
+        try_to_create_sr_with_missing_device('largeblock', 'LARGEBLOCK-local-SR-test', host)
+
+    def test_create_and_destroy_sr(self, host, sr_disk_4k):
+        # Create and destroy tested in the same test to leave the host as unchanged as possible
+        sr = host.sr_create('largeblock', "LARGEBLOCK-local-SR-test", {'device': '/dev/' + sr_disk_4k}, verify=True)
+        # import a VM in order to detect vm import issues here rather than in the vm_on_xfs_fixture used in
+        # the next tests, because errors in fixtures break teardown
+        vm = host.import_vm(vm_image('mini-linux-x86_64-bios'), sr_uuid=sr.uuid)
```
stormi, line 23: If VM import fails, then the SR is not destroyed. Someday we should fix this, but it's not specific to this test. Just mentioning it because I see it.

tests/storage/largeblock/test_largeblock_sr_crosspool_migration.py
```diff
@@ -0,0 +1,19 @@
+import pytest
+from tests.storage import cold_migration_then_come_back, live_storage_migration_then_come_back
+
+# Requirements:
+# From --hosts parameter:
+# - host(A1): first XCP-ng host >= 8.2 with an additional unused 4KiB disk for the SR.
```
stormi, line 6: It will be >= 8.3 at first. If we let this test execute on 8.2 then we can't merge the PR until the driver is available in XCP-ng 8.2.

tests/storage/largeblock/test_largeblock_sr.py
```diff
@@ -0,0 +1,67 @@
+import pytest
+from lib.common import wait_for, vm_image
+from tests.storage import try_to_create_sr_with_missing_device, vdi_is_open
+
+# Requirements:
+# - one XCP-ng host >= 8.3 with an additional unused 4KiB disk for the SR
```
stormi, line 6: To enforce the version requirement, you can include the `host_at_least_8_3` fixture.

See zfsvol tests for an example.

Wescoeur (approved):

conftest.py
```diff
@@ -227,6 +235,21 @@ def sr_disk(request, host):
...
+    if disk == "auto":
+        logging.info(">> Check for the presence of a free 4KiB block device on the master host")
+        disks = host.available_disks(4096)
+        assert len(disks) > 0, "a free 4KiB block device is required on the master host"
+        disk = disks[0]
+        logging.info(f">> Found free block device(s) on hostA1: {' '.join(disks)}. Using {disk}.")
+    else:
+        logging.info(f">> Check that block device {disk} is available on the master host")
+        assert disk in host.available_disks(4096), \
+            f"block device {disk} must be available for use on master host"
```
ydirson, line 250: The message should likely mention we're looking for a 4k disk here, and the pre-existing message shoud likewise mention 512B

ydirson (changes requested):

conftest.py
Nambrok: I changed the sr_disk_4k messages but this PR hasn't touched normal sr_disk.

ydirson: > this PR hasn't touched normal sr_disk

Sure, but my point was that as the context changes and we support both 512 and 4096, we may want to make it visible when a test is for 512.  OTOH since the standard code path will also apply to 4K-able SMAPIv3, changing the other message would indeed be wrong.

ydirson (approved):

conftest.py
Nambrok: Well, the new available_disks by default gives only 512 disks as to not break existing SMAPIv1 tests on machines with multiple devices with different blocksizes. Maybe something specific to SMAPIv3 needs to be done for SR disks. Or tests specifically aimed at 4KiB can be written using the sr_disk_4k fixture for SMAPIv3 drivers.

ydirson: So my comment indeed holds :)

Maybe what we indeed need is to have `None` be the default block size to mean "any", so size-agnostic tests are possible, and make sure to pass an explicit `512` for old SMAPIv1 drivers?

Nambrok: That would work. We could leave the default being 512 for existing users of sr_disk but add a any parameter to available_disks.

stormi (commented):
We now need to adapt the jobs so that merging this doesn't break the existing storage jobs on hosts where no 4K disk is available.

A possible solution is excluding `largeblock` from the existing test jobs and creating separate jobs for these tests, as we did for the linstor jobs. I'm not fond of the idea, because it means creating 4 new jobs for just one driver.

Another option is making the test skip, without errors, when either `--sr-disk-4k` is not provided, or `--sr-disk-4k=auto` but no matching device was found. I don't like it very much either, because it makes it easy to forget executing the job when it should have been executed, by just omitting `--sr-disk-4k`. Probably a worse idea than creating separate jobs.

So my vote goes to new jobs, for now.

ydirson:
> it makes it easy to forget executing the job when it should have been executed

This would be addressed by some sort of "test coverage" report (as opposed to "code coverage of the product during execution of tests")

Nambrok:
For the moment, omitting `--sr-disk-4k` already is a skip, I think it's the default behavior.
Giving `--sr-disk-4k=auto` without having a 4KiB disk available ends up in a failure.

stormi:
So indeed, omitting the parameter already skips. So merging this PR would not break existing jobs but just cause the test to never be executed because we never specify `--sr-disk-4k` in `jobs.py`.

So we do need new job definitions for these new tests, and `jobs.py` must be enhanced so that we can tell it in a job definition that the `sr-disk-4k` parameter must be specified (and default to `auto`).

stormi (approved):

PR #216, opened 2024-03-25
Add a fixture to get a unique ID identified sr_disk

Nambrok:
Add a `sr_disk_devid` fixture to create a SR on top of a unique disk ID.
`disk_devid` obtain the `/dev/disk/by-id/` identifier of a disk on the host. 
It is written to be an inplace replacement of the `sr_disk` fixture.

It is mainly interesting for the new LargeBlock SR tests since those fails if the disk is not the right one in the PBD.
EXTSR doesn't use the `device-config:device` for attach and detach, it obtains the VG name and LV name to mount by using the SR UUID.
LargeBlockSR creates the loop device on top of the device each time. If the device in the PBD is not the right one, the loop device will be created on top of the wrong device.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -229,7 +229,7 @@ def sr_disk(request, host):
 
 @pytest.fixture(scope="session")
 def sr_disk_devid(host, sr_disk):
-    # [5:] remove /dev/ at is not expected by normal sr_disk fixture
+    # Remove /dev/ as it is not expected by normal `sr_disk` fixture
     yield host.disk_devid(sr_disk)[0][len("/dev/"):]
 
 @pytest.fixture(scope='session')
--- a/conftest.py
+++ b/conftest.py
@@ -229,7 +229,7 @@ def sr_disk(request, host):
 
 @pytest.fixture(scope="session")
 def sr_disk_devid(host, sr_disk):
-    #[5:] remove /dev/ at is not expected by normal sr_disk fixture
+    # [5:] remove /dev/ at is not expected by normal sr_disk fixture
     yield host.disk_devid(sr_disk)[0][len("/dev/"):]
 
 @pytest.fixture(scope='session')
--- a/conftest.py
+++ b/conftest.py
@@ -227,6 +227,11 @@ def sr_disk(request, host):
             f"disk or block device {disk} is either not present or already used on master host"
     yield disk
 
+@pytest.fixture(scope="session")
+def sr_disk_devid(host, sr_disk):
+    #[5:] remove /dev/ at is not expected by normal sr_disk fixture
+    yield host.disk_devid(sr_disk)[0][len("/dev/"):]
+
 @pytest.fixture(scope='session')
 def sr_disk_for_all_hosts(request, host):
     disk = request.param
--- a/lib/host.py
+++ b/lib/host.py
@@ -406,6 +406,9 @@ class Host:
         """
         return [disk for disk in self.disks() if self.disk_is_available(disk)]
 
+    def disk_devid(self, disk):
+        return self.ssh(["find", "-L", "/dev/disk/by-id/", "-samefile", "/dev/"+disk]).splitlines()
+
     def file_exists(self, filepath, regular_file=True):
         option = '-f' if regular_file else '-e'
         return self.ssh_with_result(['test', option, filepath]).returncode == 0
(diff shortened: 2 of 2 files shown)


conftest.py
```diff
@@ -227,6 +227,11 @@ def sr_disk(request, host):
             f"disk or block device {disk} is either not present or already used on master host"
     yield disk
 
+@pytest.fixture(scope="session")
+def sr_disk_devid(host, sr_disk):
+    # [5:] remove /dev/ at is not expected by normal sr_disk fixture
```
benjamreis, line 232: You don't use `[5:]` but `[len('/dev/')]`. ;)
```suggestion
    # Remove /dev/ at is not expected by normal sr_disk fixture
```

benjamreis: Also, your fixture herits from `sr_disk` and not the other way arount so I don't understand why what's expected by the parent fixture matters here.

conftest.py
```diff
@@ -227,6 +227,11 @@ def sr_disk(request, host):
...
+    yield host.disk_devid(sr_disk)[0][len("/dev/"):]
```
benjamreis, line 233: Are we sure `host.disk_devid` won't return an empty list?
Failures in fixture are really tough to handle.

lib/host.py
```diff
@@ -406,6 +406,9 @@ def available_disks(self):
         """
         return [disk for disk in self.disks() if self.disk_is_available(disk)]
 
+    def disk_devid(self, disk):
+        return self.ssh(["find", "-L", "/dev/disk/by-id/", "-samefile", "/dev/" + disk]).splitlines()
```
benjamreis, line 410: Will the dev-id be the same on all hosts?

benjamreis (changes requested):

conftest.py
Nambrok: Yes, since `sr_disk` check that the disk exist and any disk that exist will have a symlink in `/dev/disk/by-id/` where we check. `host.disk_devid` returns an array of the `disk/by-id`. From what I currently understand, it should always have at least one. I will be investigating more on that.

lib/host.py
Nambrok: I'm not sure I understand. A disk devid is linked to the host we get it from like the disk is originally.

benjamreis: sr_disk will respond the same answer on all hosts (for instance `/sdb`)
if the id returned changes on each hosts this fixture wont work on pools with several hosts.

benjamreis (changes requested):

lib/host.py
stormi: Actually `sr_disk` is only a fixture for the master host, contrarily to `sr_disk_for_all_hosts`. So I see no problem with a fixture which just extracts this ID from the disk, as long as we don't try to reuse the dev-id on other hosts of the pool.

conftest.py
```diff
@@ -227,6 +227,11 @@ def sr_disk(request, host):
             f"disk or block device {disk} is either not present or already used on master host"
     yield disk
 
+@pytest.fixture(scope="session")
+def sr_disk_devid(host, sr_disk):
```
stormi, line 231: Making this fixture depend on `sr_disk` will make it impossible to run all storage tests on the same host in the same test job.

I believe a 4K disk should depend on a separate fixture for 4K disks, so that a host can have two disks, one with 512 blocks (`sr_disk`) and one with 4K blocks (`sr_4k_disk`). As suggested in https://github.com/xcp-ng/xcp-ng-tests/pull/215 where I got no answer.

Nambrok: This PR is independent from the LargeBlock one because this fixture could be used by other tests that need a unique identifier for its disks. I'm gonna address the need for a special fixture for 4KiB disks in the #215 . If you do not want this fixture to be added, this PR can be closed.

benjamreis (approved):
LGTM I'll let @stormi decide for the relevance of the PR.
I'm not sure we'll have other storage drivers needing to use the dev id.

stormi:
So, I don't think this PR can be merged as is:
* No existing test will use this fixture
* New tests from PR #215 need to use a new `--sr-4k-disk` CLI parameter, and associated fixtures

Since it will be really specific to 4K disks, I think this should be done as part of PR #215

PR #222, opened 2024-04-08
Only host with zfsvol can create SR of type zfsvol

gthvn1:


tests/storage/zfsvol/test_zfsvol_sr.py
```diff
@@ -12,14 +12,17 @@
 
 pytestmark = pytest.mark.usefixtures("host_at_least_8_3")
 
+
+@pytest.mark.usefixtures("host_with_zfsvol")
```
stormi, line 16: This is not needed at the moment since you also added the fixture to the only test in the class

gthvn1: Ok. I thought that to use host_with_zfsvol as parameter I needed to also specify that it is a feature. I will remove it.

tests/storage/zfsvol/conftest.py
```diff
@@ -8,6 +8,7 @@
 def host_with_zfsvol(host_with_saved_yum_state):
     host = host_with_saved_yum_state
     host.yum_install(['xcp-ng-xapi-storage-volume-zfsvol'])
+    host.restart_toolstack(verify=True)
```
gthvn1, line 11: we should also maybe add `systemctl daemon-reload` ?

Wescoeur: @stormi Can we directly add a `systemctl daemon-reload`  in the spec file of the smapi v3? Is there any contraindication to this?

gthvn1:
As zfs-vol has bugs, running quicktest on zfs vol leaves TAP disk processes that we must clean up manually to be able to remove VDIs on SR storage and finally removes SR. So don't run quicktest until we fix the issue.

tests/storage/zfsvol/conftest.py
stormi: Any RPM which updates a service file should usually call it, there are even macros that do it for you.

jobs.py
```diff
@@ -341,6 +341,9 @@
 BROKEN_TESTS = [
     # not really broken but has complex prerequisites (3 NICs on 3 different networks)
     "tests/migration/test_host_evacuate.py::TestHostEvacuateWithNetwork",
+    # running quicktest on zfsvol generates dandling TAP devices that are hard to
```
stormi, line 344: I think you mean "dangling".
```suggestion
    # running quicktest on zfsvol generates dangling TAP devices that are hard to
```

jobs.py
```diff
@@ -341,6 +341,9 @@
...
+    # cleanup. Bug need to be fixed before enabling quicktest on zfsvol.
```
stormi, line 345:
```suggestion
    # cleanup. Bug needs to be fixed before enabling quicktest on zfsvol.
```

tests/storage/zfsvol/test_zfsvol_sr.py
```diff
@@ -19,7 +19,8 @@ class TestZfsvolSRCreateDestroy:
     and VM import.
     """
 
-    def test_create_and_destroy_sr(self, host, sr_disk_wiped):
+    def test_create_and_destroy_sr(self, host_with_zfsvol, sr_disk_wiped):
```
stormi, line 22: AFAIK the order of fixtures determines which one is executed first. Here, we can ask ourselves: what do we want to fail first? The answer is: `sr_disk_wiped`. If the SR is missing, there's no point in starting to install RPM packages through the `host_with_zfsvol` fixture.

Thus I'd swap both parameters.

jobs.py
gthvn1: dandling is more poetic but yes it is dangling like dangling pointers :)

tests/storage/zfsvol/test_zfsvol_sr.py
gthvn1: yes but where is this function called? Because all others tests are doing the same so I guess that there is some magic somewhere that calls it and we cannot do what we want?

stormi: if other tests do the same, let's keep as is. We can improve them all together later if needed.

stormi: > yes but where is this function called? 

pytest collects the fixtures, orders them, and then initializes them depending on their scope and what tests require them.

stormi: What I observed additionnally is that when all other conditions have been treated by pytest and fixtures are at the same order and remain to be sorted, it looks like it then it oses the order in which it saw them. Could be an internal implementation detail and subject to change.

jobs.py
stormi: It's also in the commit message

tests/storage/zfsvol/conftest.py
stormi, line 11: This should be in a separate commit.

ydirson (approved):
lgtm

Wescoeur (approved):

PR #223, opened 2024-04-15
Pci passthrough tests

benjamreis:
- new fixtures:
  - `enabled_pci_uuid`
  - `enabled_pgpu_uuid`

- new tests
  - `TestPCIPassthrough`: tests XAPI API to enable/disable
dom0 access of PCI devices and the determination of said
access
  - `TestPGPUPCIInheritance`: tests the dom0 access of PGPU
is always sync to the dom0 access of its associated PCI device
when we use the APIs.

### DIFF
--- a/jobs.py
+++ b/jobs.py
@@ -332,6 +332,16 @@ JOBS = {
         "nb_pools": 1,
         "params": {},
         "paths": ["tests/xo"],
+    },
+    "pci-passthrough": {
+        "description": "Testing PCI passthrough functionalities",
+        "requirements": [
+            "A XCP-ng host >= 8.3 with a PGPU and a PCI to passthrough.",
+            "The host will be rebooted by the tests."
+        ],
+        "nb_pools": 1,
+        "params": {},
+        "paths": ["tests/pci_passthrough"],
     }
 }
 
--- a/tests/pci_passthrough/conftest.py
+++ b/tests/pci_passthrough/conftest.py
@@ -1,0 +1,50 @@
+import logging
+import pytest
+
+from lib.common import safe_split
+
+@pytest.fixture(scope="session")
+def enabled_pci_uuid(host):
+    pci_uuids = safe_split(host.xe("pci-list", {"host-uuid": host.uuid}, minimal=True), ',')
+
+    pci_uuid = None
+    for uuid in pci_uuids:
+        dom0_access = host.xe("pci-get-dom0-access-status", {"uuid": uuid})
+        if dom0_access == "enabled":
+            pci_uuid = uuid
+            break
+
+    if pci_uuid is None:
+        pytest.skip("This test requires a PCI to hide from dom0")
+
+    yield pci_uuid
+
+    # Put PCI back in initial state
+    if host.xe("pci-get-dom0-access-status", {"uuid": pci_uuid}) != "enabled":
+        host.xe("pci-enable-dom0-access", {"uuid": pci_uuid})
+        if host.xe("pci-get-dom0-access-status", {"uuid": pci_uuid}) != "enabled":
+            host.reboot(verify=True)
+
+@pytest.fixture(scope="session")
+def enabled_pgpu_uuid(host):
+    pgpu_uuids = safe_split(host.xe("pgpu-list", {"host-uuid": host.uuid}, minimal=True), ',')
+
+    pgpu_uuid = None
+    for uuid in pgpu_uuids:
+        pci_uuid = host.xe("pgpu-param-get", {"uuid": uuid, "param-name": "pci-uuid"})
+        dom0_access = host.xe("pci-get-dom0-access-status", {"uuid": pci_uuid})
+        if dom0_access == "enabled":
+            pgpu_uuid = uuid
+            break
+
+    if pgpu_uuid is None:
+        pytest.skip("This test requires a PGPU to hide from dom0")
+
+    yield pgpu_uuid
+
+    # Put PGPU back in initial state
+    pci_uuid = host.xe("pgpu-param-get", {"uuid": pgpu_uuid, "param-name": "pci-uuid"})
+    if host.xe("pci-get-dom0-access-status", {"uuid": pci_uuid}) != "enabled":
+        host.xe("pci-enable-dom0-access", {"uuid": pci_uuid})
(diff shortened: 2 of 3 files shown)


benjamreis:
Wait for https://github.com/xapi-project/xen-api/pull/5542 to be merged + backported in our repos.

benjamreis:
I've created a specific job for this tests as the requirements on a PGPU and PCI are specific to these tests.

gduperrey (approved):

conftest.py
```diff
@@ -158,12 +158,14 @@ def host_at_least_8_3(host):
     version_str = "8.3"
     if not host.xcp_version >= version.parse(version_str):
         pytest.skip(f"This test requires an XCP-ng >= {version_str} host")
+    return host
```
stormi, line 161: So, this fixture doesn't return anything on purpose, so that we use it outside function parameters (usually globally at the file level or the class level).

Maybe it's a good idea to have fixtures depend on this one, but I'm not sure. It hides the requirement from the test files.

jobs.py
```diff
@@ -332,6 +332,15 @@
         "nb_pools": 1,
         "params": {},
         "paths": ["tests/xo"],
+    },
+    "pci-passthrough": {
+        "description": "Testing PCI passthrough functionalities",
+        "requirements": [
+            "A XCP-ng host >= 8.3 with a PGPU and a PCI to passthrough.",
```
stormi, line 339: Better also mention that it can be rebooted.

tests/pci_passthrough/test_pci_passthrough.py
```diff
@@ -0,0 +1,82 @@
...
+
+XEN_CMDLINE = '/opt/xensource/libexec/xen-cmdline'
+
+@pytest.mark.reboot # reboots the host
+@pytest.mark.incremental # tests depend on each other. If one test fails, don't execute the others
+class TestPCIPassthrough:
+    def test_disable_dom0_access(self, host, enabled_pci_uuid):
+        host.xe("pci-disable-dom0-access", {"uuid": enabled_pci_uuid})
+        assert host.xe("pci-get-dom0-access-status", {"uuid": enabled_pci_uuid}) == "disable_on_reboot"
+
```
stormi, line 15: On some other tests, you don't add a newline here, so I think it's not necessary.
```suggestion
```

tests/pci_passthrough/test_pci_passthrough.py
```diff
@@ -0,0 +1,82 @@
...
+        host.reboot(verify=True)
+        assert host.xe("pci-get-dom0-access-status", {"uuid": enabled_pci_uuid}) == "disabled"
+
+    def test_enable_dom0_access(self, host, enabled_pci_uuid):
+        host.xe("pci-enable-dom0-access", {"uuid": enabled_pci_uuid})
+        assert host.xe("pci-get-dom0-access-status", {"uuid": enabled_pci_uuid}) == "enable_on_reboot"
+
```
stormi, line 22:
```suggestion
```

tests/pci_passthrough/conftest.py
```diff
@@ -0,0 +1,52 @@
...
+    if pci_uuid is None:
+        pytest.skip("This test requires a PCI to hide from dom0")
+
+    yield pci_uuid
+
+    # Put PCI back in initial state
+    if host.xe("pci-get-dom0-access-status", {"uuid": pci_uuid}) != "enabled":
+        host.xe("pci-enable-dom0-access", {"uuid": pci_uuid})
+        if host.xe("pci-get-dom0-access-status", {"uuid": pci_uuid}) != "enabled":
+            host.reboot(verify=True)
```
stormi, line 27: I'm not very comfortable with a fixture causing a reboot at the end of the tests, but I don't really have a better suggestion at the moment :)

tests/pci_passthrough/test_pci_passthrough.py
```diff
@@ -0,0 +1,82 @@
+import pytest
+import logging
+
+# Requirements:
+# - an XCP-ng host (--hosts) >= 8.3 with a PGPU to hide from dom0
+
+XEN_CMDLINE = '/opt/xensource/libexec/xen-cmdline'
+
+@pytest.mark.reboot # reboots the host
+@pytest.mark.incremental # tests depend on each other. If one test fails, don't execute the others
```
stormi, line 10: Incremental tests are not a very good idea. I regret I ever used it in the first place, because people often get bitten when they try to isolate a test to reproduce an issue. Here, we could avoid it by simply grouping "disable" and "enable" tests together so that the host gets back to its initial state when the test ends. Actually, that's already what you do in `test_access_status_manual_modification`, so I strongly suggest we do it for the rest too.

conftest.py
benjamreis: The requirements are also documented in the test initial comments and in the jobs description etc.
It simplifies some cases imho.

tests/pci_passthrough/conftest.py
benjamreis: I try to reboot the least possible (ie in case of bad tests execution) in order to leave the host in the same state it began.

tests/pci_passthrough/test_pci_passthrough.py
benjamreis: I see what you mean, I have no preference, the tests would be longer and less specific but this doesn"t bother that much. :+1:

conftest.py
stormi: What does it simplify?

benjamreis: https://github.com/xcp-ng/xcp-ng-tests/pull/223/files#diff-006a1a28400f0825b0c1859d4cade24243e13ee0f16aa58fa072e57b08621d78R30

stormi: I did check where it was used in the PR before adding my comments

stormi: I don't think it's the right choice, but I've already given my arguments and if you think it's more important that a fixture can require this fixture that is made to skip tests, so be it.

stormi (approved):

stormi:
LGTM. Holding my approve so that we don't merge the PR accidentally before the feature is ready.

stormi (approved):

PR #224, opened 2024-04-16
Fix wait_for behavior

Wescoeur:
- The old version does not take into account the execution time of `fn`. So instead of waiting 2 minutes (using the default timeout_secs delay), we can end up with much longer delays like 4 minutes.

- And for basic functions, the timeout is not respected. For example if I patch common.py like that:

```
def toto():
  return False

wait_for(toto, timeout_secs=10)
```

Run wait_for call:
```
> time ./common.py
  File "/home/ronan/Projets/sm/./test.py", line 27, in <module>
    wait_for(toto, timeout_secs=10)
  File "/home/ronan/Projets/sm/./test.py", line 18, in wait_for
    raise TimeoutError(
Exception: Timeout reached while waiting for fn call to yield True (10).

real    0m8,040s
user    0m0,021s
sys     0m0,016s
```

### DIFF
--- a/lib/common.py
+++ b/lib/common.py
@@ -32,15 +32,14 @@ def prefix_object_name(label):
 def wait_for(fn, msg=None, timeout_secs=2 * 60, retry_delay_secs=2, invert=False):
     if msg is not None:
         logging.info(msg)
-    time_left = timeout_secs
+    start_time = time.perf_counter()
     while True:
         ret = fn()
         if not invert and ret:
             return
         if invert and not ret:
             return
-        time_left -= retry_delay_secs
-        if time_left <= 0:
+        if time.perf_counter() - start_time >= timeout_secs:
             expected = 'True' if not invert else 'False'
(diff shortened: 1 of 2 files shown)


Wescoeur:
Since this change has an impact on the existing code, it will probably be necessary to adapt the parameter `timeout_secs` on few `wait_for` calls.

ydirson (approved):

benjamreis (commented):
So we need to run CI to detect all calls that needs to be changed?

stormi:
I suggest "Make wait_for more accurate" as a commit title. It's not that it was broken. It was not made for use with synchronous long calls.

stormi:
LGTM. We need to run the whole test suite by this branch due to the change in behaviour. There may be new failed timeouts.

stormi:
CC @gduperrey

gduperrey:
With the latest commit, run CI :white_check_mark:

PR #225, opened 2024-04-22
xen: Check for a valid Xen ChangeSet in 'xl dmesg'

tescande:
This commit adds the xen/test_misc.py file. The first test defined is test_xen_changeset() that checks for a valid git sha1 in the 'Latest ChangeSet' trace from 'xl dmesg'.

### DIFF
--- a/tests/xen/test_misc.py
+++ b/tests/xen/test_misc.py
@@ -1,0 +1,22 @@
+import logging
+import pytest
+import re
+
+# Misc tests for Xen host
+#
+# Requirements:
+# - XCP-ng host
+
+# Check for a valid "Latest ChangeSet" trace in 'xl dmesg'
+def test_xen_changeset(host):
+    changeset = host.ssh(['xl', 'dmesg', '|', 'grep', '"Latest ChangeSet"'])
+    regexp = r'.*Latest ChangeSet:\s*(([^,]+),.*)'
+
+    m = re.match(regexp, changeset)
+    assert m is not None, "'Latest ChangeSet' should be found in 'xl dmesg'"
+
+    full_changeset, git_sha = m.groups()
+    logging.info(f"Xen Latest ChangeSet: {full_changeset}")
+
+    m = re.match(r'[0-9a-fA-F]+', git_sha)
+    assert m is not None, "Xen Latest ChangeSet should be set and formatted as a commit hash"


gduperrey (approved):

tests/xen/test_misc.py
```diff
@@ -0,0 +1,22 @@
...
+    changeset = host.ssh(['xl', 'dmesg', '|', 'grep', '"Latest ChangeSet"'])
+    regexp = r'.*Latest ChangeSet:\s*(([^,]+),.*)'
+
+    m = re.match(regexp, changeset)
+    assert m is not None, "'Latest ChangeSet' should be found in 'xl dmesg'"
+
+    full_changeset, git_sha = m.groups()
+    logging.info(f"Xen Latest ChangeSet: {full_changeset}")
+
+    m = re.match('[0-9a-fA-F]+', git_sha)
```
stormi, line 21: Add `r` as in the first pattern?
```suggestion
    m = re.match(r'[0-9a-fA-F]+', git_sha)
```

ydirson: `r''` is really only useful when there is a `\` in the regexp

tescande: Right but still, it's a pattern so...

stormi (approved):

tests/xen/test_misc.py
ydirson: A pattern does not *have to* be a *raw string* - `r` does not stand for *regexp* ;)

tests/xen/test_misc.py
```diff
@@ -0,0 +1,19 @@
...
+
+# Check for a valid "Latest ChangeSet" trace in 'xl dmesg'
+def test_xen_changeset(host):
+    changeset = host.ssh(['xl', 'dmesg', '|', 'grep', '"Latest ChangeSet"'])
+    regexp = r'.*Latest ChangeSet:\s*([0-9a-fA-F]+),.*'
+
+    m = re.match(regexp, changeset)
+    assert m is not None, "Xen Latest ChangeSet not set"
+
+    logging.info(f"Xen Latest ChangeSet: {m.group(1)}")
```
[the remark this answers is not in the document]
tescande: Ooops, I missed that comment. I can moved the info only to the assert if that works for everybody...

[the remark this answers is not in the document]
tescande: But printing it anyway is ok for me actually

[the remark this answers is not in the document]
stormi: I think it's good as it is. If we were to do it differently, I would not add too much information into the assert, as an assert is for simple tests easy to read, and would use a if... log.error... pytest.fail (IIRC)... construct. But I don't think it's necessary to change.

PR #228, opened 2024-05-15
uefi_sb: Test pool and VM cert states when no custom cert installed

stormi:
We previously didn't check that a pool without custom certs only has PK.auth in `/var/lib/varstored`. Check this.

We also didn't test what happens when you boot a fresh UEFI VM on a pool which only has PK.auth, and this bit us. We thought it would behave some way, and it actually behaved the other way. We changed it so that it behaves as we need, so add a test which ensures this works and doesn't regress in the future.

### DIFF
--- a/tests/uefi_sb/test_varstored_cert_flow.py
+++ b/tests/uefi_sb/test_varstored_cert_flow.py
@@ -51,6 +51,8 @@ class TestPoolToDiskCertPropagationToAllHosts:
         for h in host.pool.hosts:
             logging.info(f"Check host {h} has no custom certificates on disk.")
             assert h.is_symlink(host.varstore_dir())
+            logging.info(f"Check host {h} only has PK, and no other certs.")
+            assert h.ssh(['ls', '/var/lib/varstored/']) == 'PK.auth'
 
 @pytest.mark.small_vm
 @pytest.mark.usefixtures("host_at_least_8_3")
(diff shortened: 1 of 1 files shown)


benjamreis (approved):

tests/uefi_sb/test_varstored_cert_flow.py
```diff
@@ -116,10 +118,25 @@ def test_start_vm_without_uefi_vars(self, uefi_vm):
...
         vm.start()
-        logging.info("Check that the VM certs were updated")
+        logging.info("Check that the VM certs were updated: PK, KEK, db, dbx")
         for key in ['PK', 'KEK', 'db', 'dbx']:
             check_vm_cert_md5sum(vm, key, pool_auths[key].auth)
 
+    def test_start_vm_without_uefi_vars_on_pool_with_only_pk(self, uefi_vm):
+        # When a VM first starts but the pool doesn't have certs configured,
+        # this used, until late in 8.3 development, to *not* propagate the certs to the VM
+        # and we had not test that detected this situation.
```
tescande, line 128:
```suggestion
        # and we had no test that detected this situation.
```

tescande (approved):

tests/uefi_sb/test_varstored_cert_flow.py
stormi: Thanks! Fixed.

PR #229, opened 2024-05-28
Reduce usage of pytest test-parameters

ydirson:
There is no real need to store the parameters coming from the CLI in a (single) test param, and it makes it much harder to describe test dependencies.

This PR deals with `--hosts` and `--sr-disk` (easily tested), applies same strategy to  `--second-network` (untested), and removes unused `--additional-repos`.

The same can be applied at least to `--sr-device-config`, but I'm leaving it out for now as it is more complex to test.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -55,12 +55,6 @@ def pytest_addoption(parser):
              "Example: 'server:10.0.0.1,serverpath:/vms,nfsversion:4.1'.",
     )
     parser.addoption(
-        "--additional-repos",
-        action="append",
-        default=[],
-        help="Additional repo URLs added to the yum config"
-    )
-    parser.addoption(
         "--second-network",
         action="store",
         default=None,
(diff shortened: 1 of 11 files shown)


stormi (commented):
This looks good to me. I suggest you handle the remaining parameters in the same PR, on which we will anyway run the full testsuite to make sure nothing regressed.

benjamreis (approved):

conftest.py
```diff
@@ -236,8 +231,11 @@ def sr_disk(request, host):
     yield disk
 
 @pytest.fixture(scope='session')
-def sr_disk_4k(request, host):
-    disk = request.param
+def sr_disk_4k(pytestconfig, host):
+    disks = pytestconfig.getoption("sr_disk")
```
Nambrok, line 235: Should be `sr_disk_4k`. It's a different parameter than normal `sr_disk`.

Nambrok (changes requested):

conftest.py
ydirson: right, fixed

Nambrok (approved):

stormi:
LGTM, provided a full suite run doesn't reveal any drawbacks.

ydirson:
Added a fix for a wrong usage of a package-scoped fixture (and checked no other such wrong usage exists in the codebase)

ydirson:
Rebased onto master after #232 merge

PR #233, opened 2024-06-13
packages: import pkgfixture only once

ydirson:
With current pytest (8.2.2) having the fixture imported in 2 subpackages of "tests.packages" causes confusion on the execution order of setup/teardown and causes host_with_saved_yum_state to be run nested, triggering protective assertion.

This is a (tentative) workaround for pytest handling of package-scoped fixtures, and can possibly break something else.


ydirson:
obsoleted by #234

PR #235, opened 2024-06-17
Change the timeout for tests largeblocks reboot

gduperrey:


tests/storage/largeblock/test_largeblock_sr.py
```diff
@@ -58,7 +58,7 @@ def test_reboot(self, host, largeblock_sr, vm_on_largeblock_sr):
         sr = largeblock_sr
         vm = vm_on_largeblock_sr
         host.reboot(verify=True)
-        wait_for(sr.all_pbds_attached, "Wait for PBD attached")
+        wait_for(sr.all_pbds_attached, "Wait for PBD attached", timeout_secs=10 * 60)
```
stormi, line 61: Waiting for the reboot is done on line 60. If we really need to wait more than two minutes for the PBDs to be attached, there's something wrong somewhere.

PR #236, opened 2024-06-26
Add a test to check compatibility between mellanox drivers

gthvn1:
The mlx4 driver requires mlx_compat.ko from the mlx5 driver and the version differ. Someday compatibility may break. This test will detect it.

### DIFF
--- a/tests/packages/mlx/conftest.py
+++ b/tests/packages/mlx/conftest.py
@@ -1,0 +1,25 @@
+import pytest
+import logging
+
+# Explicitly import package-scoped fixtures (see explanation in pkgfixtures.py)
+from pkgfixtures import host_with_saved_yum_state
+
+@pytest.fixture(scope="package")
+def host_without_mlx_compat_loaded(host_with_saved_yum_state):
+    host = host_with_saved_yum_state
+
+    # We need to check if mlx_compat module is loaded. If it is already loaded,
+    # we need to unload it before starting the test and load it again when the test ends.
+    mlx_compat_loaded = host.ssh_with_result(['lsmod', '|', 'grep', 'mlx_compat']).returncode == 0
+    if mlx_compat_loaded:
+        logging.info("mlx_compat is loaded so unload it before testing")
+        host.ssh(['modprobe', '-r', '-v', 'mlx_compat'])
+    else:
+        logging.info("mlx_compat is not loaded")
+
+    yield host
+
+    # Reload the mlx_compat module if it was loaded when starting the test
+    if mlx_compat_loaded:
+        logging.info("test is done so reload mlx_compat")
+        host.ssh(['modprobe', '-v', 'mlx_compat'])
--- a/tests/packages/mlx/test_mellanox-modules-compat.py
+++ b/tests/packages/mlx/test_mellanox-modules-compat.py
@@ -1,0 +1,26 @@
+# Requirements:
+# From --hosts parameter:
+# - host(A1): any master host of a pool, with access to XCP-ng RPM repositories.
+
+MLX4_MODULE = 'mlx4_en'
+
+def load_unload_mlx_module(host):
+    host.ssh(['modprobe', '-v', MLX4_MODULE])
+    host.ssh(['modprobe', '-r', '-v', MLX4_MODULE])
+
+def test_install_mlx_modules_alt(host_without_mlx_compat_loaded):
+    host = host_without_mlx_compat_loaded
+
+    # Ensure the modules are unloaded
+    host.yum_remove(['mlx4-modules-alt', 'mellanox-mlnxen-alt'])
+
+    # Start by loading mlx4
+    host.yum_install(['mlx4-modules-alt'])
+    load_unload_mlx_module(host)
+
+    # Ensure that mlx_compat is still unloaded
+    assert host.ssh_with_result(['lsmod', '|', 'grep', 'mlx_compat']).returncode == 1
+
+    # Now load mellanox-mlnxen-alt
+    host.yum_install(['mellanox-mlnxen-alt'])
+    load_unload_mlx_module(host)


tests/packages/mlx/test_mellanox-modules-compat.py
```diff
@@ -0,0 +1,19 @@
+# Requirements:
+# From --hosts parameter:
+# - host(A1): any master host of a pool, with access to XCP-ng RPM repositories and reports.xcp-ng.org.
+import logging
+from pkgfixtures import host_with_saved_yum_state
```
stormi, line 5: Must be imported in `conftest.py` for the package (and it's better if the package contains an empty `__init__.py` file to indicate it is a package) for the workaround to work correctly, IIUC. (cc @ydirson)

tests/packages/mlx/test_mellanox-modules-compat.py
```diff
@@ -0,0 +1,19 @@
...
+
+def test_install_mlx_modules_alt(host_with_saved_yum_state):
+    host = host_with_saved_yum_state
+    logging.info(f"Installing mlx4-modules-alt alone")
+    host.yum_install(['mlx4-modules-alt'])
+    host.ssh(['modprobe', 'mlx4_ib'])
+    host.ssh(['rmmod', 'mlx4_ib'])
```
stormi, line 12: Why not use `modprobe -r` here?

Also, we should remove from memory modules that were pulled by mlx4_en, especially `mlx_compat.ko` which is the one that might become incompatible someday. Otherwise they will remain and cause the next execution of these tests to be inaccurate.

tests/packages/mlx/test_mellanox-modules-compat.py
```diff
@@ -0,0 +1,19 @@
...
+# From --hosts parameter:
+# - host(A1): any master host of a pool, with access to XCP-ng RPM repositories and reports.xcp-ng.org.
+import logging
+from pkgfixtures import host_with_saved_yum_state
+
+def test_install_mlx_modules_alt(host_with_saved_yum_state):
+    host = host_with_saved_yum_state
+    logging.info(f"Installing mlx4-modules-alt alone")
+    host.yum_install(['mlx4-modules-alt'])
+    host.ssh(['modprobe', 'mlx4_ib'])
```
stormi, line 11: `modprobe -v` may give us slightly better debug logs.

tests/packages/mlx/test_mellanox-modules-compat.py
```diff
@@ -0,0 +1,19 @@
+# Requirements:
+# From --hosts parameter:
+# - host(A1): any master host of a pool, with access to XCP-ng RPM repositories and reports.xcp-ng.org.
+import logging
+from pkgfixtures import host_with_saved_yum_state
+
+def test_install_mlx_modules_alt(host_with_saved_yum_state):
```
stormi, line 7: `host_with_saved_yum_state` will only restore the saved state after all tests of the package that need it have run. This means that it will not restore the state between your two tests. We don't have a fixture which does this at the test level, currently, but maybe we should. The problem is, our current implementation doesn't allow stacking saved yum states (and this sometimes allows us to detect weird overlaps of fixtures).

tests/packages/mlx/test_mellanox-modules-compat.py
stormi, line 11: It's `mlx4_en`. When in doubt due to incomplete specs, just ask :)

We're not sure `mlx4_ib` is of any use but we left it in the package as a build product.

```suggestion
    host.ssh(['modprobe', 'mlx4_ib'])
```

stormi (changes requested):

tests/packages/mlx/test_mellanox-modules-compat.py
stormi: Now that I think of it, nothing prevents us from restoring the yum state at the end of each test, but we must do so without discarding the saved state established by the `host_with_saved_yum_state` fixture before the first test.

stormi: Following this idea, in `lib/hosts.py`, 

```
def yum_restore_saved_state(self):
```

could become

```
def yum_restore_saved_state(self, discard_saved_state=True):
```

tests/packages/mlx/conftest.py
```diff
@@ -0,0 +1,14 @@
...
+def host_with_mlx4(host_with_saved_yum_state):
+    host = host_with_saved_yum_state
+    host.yum_install(['mlx4-modules-alt'])
+    yield
+
+@pytest.fixture(scope="package")
+def host_with_mlnxen(host_with_saved_yum_state):
+    host = host_with_saved_yum_state
+    host.yum_install(['mlx4-modules-alt', 'mellanox-mlnxen-alt'])
+    yield
```
gthvn1, line 14: host missing

tests/packages/mlx/test_mellanox-modules-compat.py
gthvn1: or I can also just uninstall the `mlx4-modules-alt` at the end of the first test...

gthvn1: I mean yum insall and yum remove within the test itself. So I save the state and within the test I install/remove what I need and at the end the state is restored...

tests/packages/mlx/test_mellanox-modules-compat.py
```diff
@@ -0,0 +1,27 @@
...
+
+from pkgfixtures import host_with_saved_yum_state
+
+MLX4_MODULE = 'mlx4_en'
+
+def load_unload_mlx_module(host):
+    logging.info("loading mlx module")
+    host.ssh(['modprobe -v', MLX4_MODULE])
+    logging.info("unloading mlx module")
+    host.ssh(['modprobe -r', MLX4_MODULE])
```
stormi, line 14: Let's be also verbose when we unload the module?
```suggestion
    host.ssh(['modprobe -rv', MLX4_MODULE])
```

tests/packages/mlx/test_mellanox-modules-compat.py
```diff
@@ -0,0 +1,27 @@
...
+
+def test_install_mlx_modules_alt(host_with_saved_yum_state):
+    host = host_with_saved_yum_state
+    # Ensure the modules are unloaded
+    host.yum_remove(['mlx4-modules-alt', 'mellanox-mlnxen-alt'])
+
+    # Start by loading mlx4
+    host.yum_install(['mlx4-modules-alt'])
+    load_unload_mlx_module(host)
```
stormi, line 23: This doesn't unload `mlx_compat.ko` to make space for the one provided by `mellanox-mlnxen-alt` in the next part of the text.

tests/packages/mlx/test_mellanox-modules-compat.py
ydirson: yes, `__init__.py` is necessary too, or the directory is not even a package and the workaround cannot help, as we saw recently

tests/packages/mlx/conftest.py
```diff
@@ -0,0 +1 @@
+from pkgfixtures import host_with_saved_yum_state
```
stormi, line 1: Because it's a workaround, we repeat this comment when we import such fixtures.

```suggestion
# Explicitly import package-scoped fixtures (see explanation in pkgfixtures.py)
from pkgfixtures import host_with_saved_yum_state
```

tests/packages/mlx/test_mellanox-modules-compat.py
gthvn1: modprobe -r remove also dependencies no? So it should remove `mlx4_en` and its dependencies...
Or are you saying that `mlx_compat.ko` is already installed and so not installed as a dependency? So it should be removed at the beginning of the test. I looked on my host and it is not loaded after running test. But I have not a clean host so I will test again using fresh install.

stormi: I think you're right. But, yes, `mlx_compat.ko` is already installed in the case of XCP-ng 8.3. Not loaded if there's no hardware that requires it, but installed it is.

gthvn1: Just tested on my system and I have:
```
[18:05 xcp-gtn-cirepo ~]# modprobe -v mlx4_en
insmod /lib/modules/4.19.0+1/kernel/net/core/devlink.ko
insmod /lib/modules/4.19.0+1/updates/mlx_compat.ko
insmod /lib/modules/4.19.0+1/override/mlx4_core.ko
insmod /lib/modules/4.19.0+1/override/mlx4_en.ko
```
```
[18:06 xcp-gtn-cirepo ~]# modprobe -rv mlx4_en
rmmod mlx4_en
rmmod mlx4_core
rmmod devlink
rmmod mlx_compat
```

gthvn1: Note that I have chosen at fist to load `mlx4_ib` because it loads all dependencies including `mlx4_en`.

gthvn1: And on my fresh install of xcp-ng 8.3 `mlx_compat`  was not loaded. But ok I can check and ensure that it is not loaded before testing. But does that mean that if it is already loaded before testing I need to load it when test ends. And the other way, if it was not loaded it should remains unloaded? or is it too much?

stormi: On hardware which uses the driver, we should not unload it, so I suggest you just add a fixture which checks that the module is not loaded and fails the test otherwise.

stormi: Or, since there's just one test, assert at the beginning of the test, without taking any measures to solve it directly. If this causes issues in the future, we can think of something more elaborate.

stormi: Let's also assert between the two parts of the test.

gthvn1: > On hardware which uses the driver, we should not unload it, so I suggest you just add a fixture which checks that the module is not loaded and fails the test otherwise.

but it is only use for testing so we know on which hardware we are running right? I will check if the module is not loaded and if it is we will fail. We will update the test in the future if needed as you suggest.

stormi: We know our current hardware, not future hardware, nor on what hardware someone could want to run the tests.

tests/packages/mlx/test_mellanox-modules-compat.py
```diff
@@ -0,0 +1,41 @@
...
+MLX4_MODULE = 'mlx4_en'
+
+def load_unload_mlx_module(host):
+    host.ssh(['modprobe -v', MLX4_MODULE])
+    host.ssh(['modprobe -rv', MLX4_MODULE])
+
+def test_install_mlx_modules_alt(host_with_saved_yum_state):
+    host = host_with_saved_yum_state
+
+    # We need to check if mlx_compat module is loaded. If it is already loaded,
```
gthvn1, line 15: I will move this part of testing mlx_compat in conftest.py

tests/packages/mlx/test_mellanox-modules-compat.py
gthvn1: I finally unload the mlx_compat module and reload it at the end if needed. But if you prefer to just failed if the module is loaded I can just assert that the module is not loaded. I tested both condition (module already loaded before the test and not loaded before the test). It looks ok.

stormi: Will this work if `mlx5_core` is loaded and depends on it to operate the network?

gthvn1: If `mlx5_core` is loaded the test will failed while trying to unload `mlx_compat`. So the test will fail and the network will continue to work. The command `modprobe -r mlx_compat` cannot succeed if the module is in use by another module (like mlx5_core).

stormi (approved):

PR #238, opened 2024-07-03
xcpng-fs-diff.py: Update ignored files list

tescande:


gduperrey (approved):

scripts/xcpng-fs-diff.py
```diff
@@ -270,18 +273,23 @@ def main():
...
         '/etc/sysconfig/xencommons',
+        '/etc/sysctl.d/91-net-ipv6.conf',
         '/etc/vconsole.conf',
         '/etc/xensource-inventory',
         '/etc/xensource/boot_time_cpus',
         '/etc/xensource/ptoken',
+        '/etc/xensource/xapi-pool-tls.pem',
         '/etc/xensource/xapi-ssl.pem',
         '/opt/xensource/gpg/trustdb.gpg',
+        '*.pyc',
```
stormi, line 292: Can we be more precise? Changes in pyc files inside /usr, for example, would indicate that there's a packaging issue.

scripts/xcpng-fs-diff.py
```diff
@@ -250,15 +250,18 @@ def main():
     folders = ["/boot", "/etc", "/opt", "/usr"]
     ignored_file_patterns = [
         '/boot/initrd-*',
+        '/boot/efi/*',
```
stormi, line 253: I assume this is so that comparing a UEFI host with a BIOS can be done?

stormi:
Loosely related: are changes in the ignored files displayed somewhere, without causing an error, or are they completely masked?

scripts/xcpng-fs-diff.py
tescande: These files are generated at first run and are not distributed with packages, right ? So .pyc files might differ if they include some kind of generation timestamp.

stormi: They are supposed to distributed with packages, on the contrary, so that /usr may be read-only. But there remains some which aren't. For example, the XAPI plugins in /etc, IIRC. That's why I would not ignore all of them, only the ones we know we want to ignore.

scripts/xcpng-fs-diff.py
tescande: Right. I need to check if there is no other difference between UEFI and BIOS hosts but this is the only one that I remember.

tescande:
> Loosely related: are changes in the ignored files displayed somewhere, without causing an error, or are they completely masked?

Just added `--show-ignored` command line option

gthvn1 (approved):

gthvn1:
Just tested on fresh install and it is cool to not see the *.pyc file without using `grep -v` ;)
```
Results:
file differs: /etc/firstboot.d/data/management.conf
file doesn't exist on reference host: /etc/xapi.d/plugins/vmssc
file doesn't exist on tested host: /etc/openldap/certs/secmod.db
file doesn't exist on tested host: /etc/openldap/certs/password
file doesn't exist on tested host: /etc/openldap/certs/cert8.db
file doesn't exist on tested host: /etc/openldap/certs/key3.db
```

stormi (approved):

PR #240, opened 2024-07-18
Add tests for systemd units

tescande:


tests/fs_diff/test_fs_diff.py
```diff
@@ -0,0 +1,25 @@
...
+import subprocess
+
+# Requirements:
+# - 2 XCP-ng host of same version
+
+pytest.fixture(scope='module')
+def test_fs_diff(hosts):
+    assert len(hosts) == 2, "This test requires exactly 2 hosts"
+
+    assert (hosts[0].xcp_version == hosts[1].xcp_version), f"Hosts versions differ"
```
stormi, line 12: The message associated to an assert should state what is expected, not what's wrong when it fails.

jobs.py
```diff
@@ -407,6 +407,15 @@
         "params": {},
         "paths": ["tests/pci_passthrough"],
     },
+    "fs-diff": {
+        "description": "Check for difference between 2 hosts file system",
+        "requirements": [
+            "2 XCP-ng host >= 8.2"
+        ],
+        "nb_pools": 1,
```
stormi, line 415: Two here.

tests/systemd/test_systemd.py
stormi, line 1: Is this test related to all the other commits in this PR?

tests/systemd/test_systemd.py
```diff
@@ -0,0 +1,27 @@
...
+        logging.error(f"Unit {unit.split()[0]} failed")
+
+    assert not failed_services
+
+pytest.fixture(scope='module')
+def test_unit_dep_cycles(host):
+    analyse = host.ssh(['systemd-analyze', 'verify', 'default.target'])
+    err = False
+    for line in analyse.splitlines():
+        m = re.match("^Found ordering cycle on .*$", line)
```
stormi, line 22: Is this the only kind of error `systemd-analyze verify` could reveal?

jobs.py
```diff
@@ -416,6 +416,15 @@
...
     },
+    "systemd": {
+        "description": "Testing systemd failed units and dependency cycles",
+        "requirements": [
+            "A XCP-ng host >= 8.2"
+        ],
+        "nb_pools": 1,
+        "params": {},
+        "paths": ["tests/systemd"],
+    },
```
stormi, line 427: I'd rather have it in the main job, not separated into its own job. Also, I'm not sure we need a new `systemd` directory in `tests`. `tests/system` could be a fit.

tests/systemd/test_systemd.py
tescande: Sure not, but in this test `test_unit_dep_cycles()` we're only look for dep cycle issues.

jobs.py
tescande: Works for me. Will move it into system (it was there in the first place actually :))

tests/fs_diff/test_fs_diff.py
tescande: Fixed

jobs.py
```diff
@@ -407,6 +407,15 @@
         "params": {},
         "paths": ["tests/pci_passthrough"],
     },
+    "fs-diff": {
+        "description": "Check for difference between 2 hosts file system",
```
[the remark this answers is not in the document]
tescande: Fixed

tests/systemd/test_systemd.py
tescande: Moved to a new PR: https://github.com/xcp-ng/xcp-ng-tests/pull/246

jobs.py
tescande: Fixed

tests/systemd/test_systemd.py
tescande: Will investigate other useful systemd tests

ydirson (dismissed):
Rather than starting a blacklist with one failure case we happen to know about (cycles), what about instead using a whitelist with the specific warnings we did evaluate as being acceptable?

PR #243, opened 2024-07-25
Test diskless resource becomes tiebreaker upon VM shutdown

benjamreis:


gduperrey:
As seen with the failed run on the CI:
* 8.2 xen :x:
Currently the CI is running through Jenkins on AlmaLinux release 8.10 (Cerulean Leopard) with Python 3.6.8.
It seems the version is too old for these changes.
So there are discussions to be about modification of the tests or trying to upgrade the almalinux of the Jenkins servers (3) to version 9 (Python 3.9.18).

benjamreis:
Unecessary.

PR #250, opened 2024-08-29
ssh: don't use a separate logger

ydirson:
The logger setup was confusing when using --log-file-level=DEBUG, showing this debug-level output inside default --log-cli-level=INFO output, but without the debug context that only went into the logfile, and without any hint of what kind of output it was.

This makes it use the same logger as all other debug output, preventing it from appearing in the wrong place, and adds a marker to hint it is output data - at the same time fixing a potential formatting issue, where arbitrary ssh output lines were interpreted as *format strings*.

### DIFF
--- a/lib/commands.py
+++ b/lib/commands.py
@@ -59,12 +59,6 @@ def _ellide_log_lines(log):
         reduced_message.append("(...)")
     return "\n{}".format("\n".join(reduced_message))
 
-OUPUT_LOGGER = logging.getLogger('output')
-OUPUT_LOGGER.propagate = False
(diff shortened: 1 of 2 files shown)


benjamreis:
If the output look ok to everyone - fine by me :)

benjamreis (approved):

stormi:
I'm OK on the principle, but I want to run some tests.

stormi (approved):
LGTM. I had a weird output where I'd get the failed quicktest output after the captured test and teardown output, but couldn't reproduce, with or without `-s`.

PR #252, opened 2024-09-02
Install-driven enhancements, round 2

ydirson:


lib/host.py
```diff
@@ -566,3 +566,6 @@ def enable_hsts_header(self):
     def disable_hsts_header(self):
         self.ssh(['rm', '-f', f'{XAPI_CONF_DIR}/00-XCP-ng-tests-enable-hsts-header.conf'])
         self.restart_toolstack(verify=True)
+
+    def get_vdi_sr_uuid(self, vdi_uuid):
```
stormi, line 570: What about moving it to the VDI class?

ydirson: My thought was rather to move it to `Pool` instead.  In `VDI` it would need to be a class- or static-method, and would need to get a ref to the pool anyway.

lib/pool.py
```diff
@@ -17,6 +17,14 @@ def __init__(self, master_hostname_or_ip):
         assert master.is_master(), f"Host {master_hostname_or_ip} is not a master host. Aborting."
         self.master = master
         self.hosts = [master]
+
+        # wait for XAPI startup to be done, or we can get "Connection
+        # refused (calling connect )" when calling self.hosts_uuids()
+        wait_for(lambda: commands.ssh(master_hostname_or_ip, ['xapi-wait-init-complete', '60'],
+                                      check=False, simple_output=False).returncode == 0,
+                 f"Wait for XAPI init to be complete on {master_hostname_or_ip}",
+                 timeout_secs=30 * 60)
```
stormi, line 26: This could be useful in other places, if this implementation is better than the existing one which does a similar check by waiting for `xe` to return that the current host is enabled, as we do each time we reboot a host or restart XAPI.

I don't know which is safer:
* I'm not 100% confident in this `xapi-wait-init-complete` check
* I've seen the enabled status displayed by XO flip to enable, then shortly to disabled again, then enabled again for good.

I'd start with making it a helper method of the Host class, and decide later in what direction to harmonize (maybe both checks added together?)

lib/vm.py
```diff
@@ -474,7 +474,9 @@ def destroy_vtpm(self):
         return self.host.xe('vtpm-destroy', {'uuid': vtpm_uuid}, force=True)
 
     def clone(self):
-        name = self.name() + '_clone_for_tests'
+        name = self.name()
+        if not name.endswith('_clone_for_tests'):
+            name += '_clone_for_tests'
```
stormi, line 479: I'd rather not merge this change until a better handling is done, otherwise cloned VMs created from cached VMs will be hard to differentiate from cache by their name.

lib/host.py
stormi: To me, the SR looks like an attribute of the VDI object. Why would it be a static method? You even added the `VDI.param_get` method that you could use.

lib/vm.py
ydirson: yes there is a FIXME for this already

stormi: I didn't know if the FIXME was meant for before or after the merge :)

ydirson: inside commit message is explicitly *before* ;)

lib/host.py
ydirson: The caller has no VDI object to start with, we want to get one from just a UUID.

ydirson: (moved to `Pool` in last push)

lib/vm.py
ydirson: done in last push

ydirson (commented):
Note this includes the use of `rt_answerfile` as workaround to the Restore

gthvn1 (approved):

lib/host.py
stormi: I thought we could instanciate objets, but there's a chicken and egg problem, since we ask for a SR object in the constructor. For now, a `Pool` method will do.

ydirson:
Reworked/cleaned the `install_xcpng` split into `pxe.py`, and reinstated the `rt_answerfile` usage.
Brings this PR out of Draft.

lib/pool.py
ydirson: > This could be useful in other places

In theory yes, but in practice, how often do we interact with XAPI without building a `Pool` object?

> I'm not 100% confident in this xapi-wait-init-complete check

Neither do I, but I have not seen any practical issue with it, and it does fix those flakiness issues AFAICT.

> I've seen the enabled status displayed by XO flip to enable, then shortly to disabled again, then enabled again for good.

That would look like a bug in XAPI

stormi (approved):

gduperrey (approved):
Run CI on 8.3 :white_check_mark:

gthvn1 (approved):

PR #257, opened 2024-09-24
fs-diff: Ignored snmp and nagios configuration files

tescande:


gduperrey (approved):

stormi:
Ideally we only want to ignore them when transitioning from XenServer 8, so let's keep this as a draft for now.

PR #259, opened 2024-10-02
Accept 2 errors when certs incompatible

benjamreis:
These 2 errors can be raised depending on timing and wether the host is perceived as live or no after the toolstask restart


gduperrey (approved):

stormi (approved):
Small typo in commit message: wether => whether.

PR #260, opened 2024-10-04
fsdiff enhancements to diff against a reference dump

ydirson:


tests/fs-diff/test_fs_diff.py
```diff
@@ -5,7 +5,6 @@
 # Requirements:
 # - 2 XCP-ng host of same version
 
-pytest.fixture(scope='module')
```
stormi, line 8: This one is fun :) Nice catch.

lib/host.py
```diff
@@ -566,3 +566,8 @@ def enable_hsts_header(self):
     def disable_hsts_header(self):
         self.ssh(['rm', '-f', f'{XAPI_CONF_DIR}/00-XCP-ng-tests-enable-hsts-header.conf'])
         self.restart_toolstack(verify=True)
+
+    def firmware_type(self):
+        retcode = self.ssh(['test', '-d', '/sys/firmware/efi/'],
+                           check=False, simple_output=False).returncode
+        return "uefi" if retcode == 0 else "bios"
```
stormi, line 573: In the VM object, there's `is_uefi()`. Would it make sense to be consistent with it?

tests/fs-diff/data/8.3.0-uefi
stormi, line 1: Reference files will keep evolving when we test nightly builds, so does it make sense to store them here? How about we give the test a location where it can download them from if they exist, that we would populate after each nightly build?

PR #262, opened 2024-11-05
Add Windows PV tools installer tests

dinhngtu:
This PR adds tests for the new XCP-ng 9.x Windows PV Tools installer.
It tests various actions of the installer (install, uninstall, upgrade) under various conditions (with XCP-ng v8 tools, with Citrix tools, with Windows Update drivers, etc.)
It adds a dependency on 2 new ISO images, the structures of which are documented below:

```
guest-tools-win.iso
├───package
│   └───XenClean
│       └───x64
│           └───bin
└───testsign
```

```
other-guest-tools-win.iso
├───citrix-9.4.0
│   ├───XenBus
│   │   ├───x64
│   │   └───x86
│   ├───XenIface
│   │   ├───x64
│   │   └───x86
│   ├───XenNet
│   │   ├───x64
│   │   └───x86
│   ├───XenVbd
│   │   ├───x64
│   │   └───x86
│   └───XenVif
│       ├───x64
│       └───x86
├───xcp-ng-8.2.2.200
│   ├───xenbus
│   │   ├───x64
│   │   └───x86
│   ├───xeniface
│   │   ├───x64
│   │   └───x86
│   ├───xennet
│   │   ├───x64
│   │   └───x86
│   ├───xenvbd
│   │   ├───x64
│   │   └───x86
│   └───xenvif
│       ├───x64
│       └───x86
└───xcp-ng-9.0.9000
    ├───package
    │   └───XenClean
    │       └───x64
    │           └───bin
    └───testsign
```

I have run these tests on Windows Server 2016, 2019, 2022 and 2025. I used customized images with testsigning enabled.

### DIFF
--- a/jobs.py
+++ b/jobs.py
@@ -350,6 +350,19 @@ JOBS = {
         "paths": ["tests/guest-tools/unix"],
         "markers": "multi_vms",
     },
+    "tools-windows": {
+        "description": "tests our windows guest tools on a variety of VMs",
+        "requirements": [
+            "A pool >= 8.2. One host is enough.",
+            "A variety of windows VMs supported by our tools installer.",
+        ],
+        "nb_pools": 1,
+        "params": {
+            "--vm[]": "multi/tools_windows",
+        },
+        "paths": ["tests/guest-tools/win"],
+        "markers": "multi_vms",
+    },
     "xen": {
         "description": "Testing of the Xen hypervisor itself",
         "requirements": [
--- a/tests/guest-tools/win/__init__.py
+++ b/tests/guest-tools/win/__init__.py
@@ -1,9 +1,11 @@
 import enum
 import logging
 import re
+import time
 from typing import Any
 
 from data import ISO_DOWNLOAD_URL
+from lib.commands import SSHCommandFailed
 from lib.common import wait_for
 from lib.host import Host
 from lib.sr import SR
@@ -77,10 +79,21 @@ def insert_cd_safe(vm: VM, vdi_name: str, cd_path="D:/", retries=2):
     If this happens, reboot the VM in hopes that it will be detected.
     """
     for _attempt in range(retries):
+        # Eject the existing CD just in case.
+        try:
+            vm.eject_cd()
+            # Leave some time for the guest to realize its CD got ejected.
+            time.sleep(5)
+        except SSHCommandFailed:
+            pass
+
         vm.insert_cd(vdi_name)
+        if not vm.is_running():
+            vm.start()
         # wait_for(vm.file_exists) doesn't handle SSHCommandFailed;
         # we need to check for it via wait_for(vm.is_ssh_up).
         wait_for_vm_running_and_ssh_up_without_tools(vm)
+
         try:
             wait_for(lambda: vm.file_exists(cd_path, regular_file=False), timeout_secs=60)
             return
@@ -89,7 +102,5 @@ def insert_cd_safe(vm: VM, vdi_name: str, cd_path="D:/", retries=2):
             # There might be no VM tools so use SSH instead.
             vm.ssh(WINDOWS_SHUTDOWN_COMMAND)
             wait_for(vm.is_halted, "Wait for VM halted")
-            vm.eject_cd()
-            vm.start()
-            wait_for_vm_running_and_ssh_up_without_tools(vm)
+
     raise TimeoutError(f"Waiting for CD at {cd_path} failed")
--- a/tests/guest-tools/win/__init__.py
+++ b/tests/guest-tools/win/__init__.py
@@ -1,0 +1,95 @@
+import enum
+import logging
+import re
+from typing import Any
+
+from data import ISO_DOWNLOAD_URL
+from lib.common import wait_for
+from lib.host import Host
+from lib.sr import SR
+from lib.vm import VM
+
+
+# HACK: I originally thought that using Stop-Computer -Force would cause the SSH session to sometimes fail.
+# I could never confirm this in the end, but use a slightly delayed shutdown just to be safe anyway.
+WINDOWS_SHUTDOWN_COMMAND = "shutdown.exe -s -f -t 5"
+
+
+class PowerAction(enum.Enum):
+    Nothing = "nothing"
+    Shutdown = "shutdown"
+    Reboot = "reboot"
+
+
+def iso_create(host: Host, sr: SR, param: dict[str, Any]):
+    if param["download"]:
+        vdi = host.import_iso(ISO_DOWNLOAD_URL + param["name"], sr)
+        new_param = param.copy()
+        new_param["name"] = vdi.name()
+        yield new_param
+        vdi.destroy()
+    else:
+        yield param
+
+
+def try_get_and_store_vm_ip_serial(vm: VM, timeout: int):
+    domid = vm.param_get("dom-id")
+    logging.debug(f"Domain ID {domid}")
+    command = f"xl console -t serial {domid} | grep '~xcp-ng-tests~.*~end~' | head -n 1"
+    if timeout > 0:
+        command = f"timeout {timeout} " + command
+    res_host = vm.get_residence_host()
+    report = res_host.ssh(command)
+    logging.debug(f"Got report: {report}")
+    match = re.match("~xcp-ng-tests~(.*)=(.*)~end~", report)
+    if not match:
+        return False
+    vm.ip = match[2]
+    return True
+
+
+def wait_for_vm_running_and_ssh_up_without_tools(vm: VM):
+    wait_for(vm.is_running, "Wait for VM running")
+    wait_for(vm.is_ssh_up, "Wait for SSH up")
+
+
+def enable_testsign(vm: VM, rootcert: str | None):
+    if rootcert is not None:
+        vm.execute_powershell_script(
+            f"""certutil -addstore -f Root '{rootcert}';
+if ($LASTEXITCODE -ne 0) {{throw}}
+certutil -addstore -f TrustedPublisher '{rootcert}';
+if ($LASTEXITCODE -ne 0) {{throw}}"""
+        )
+    vm.execute_powershell_script(
+        "bcdedit /set testsigning on; if ($LASTEXITCODE -ne 0) {throw};" + WINDOWS_SHUTDOWN_COMMAND
+    )
+    wait_for(vm.is_halted, "Wait for VM halted")
+    vm.start()
+    wait_for_vm_running_and_ssh_up_without_tools(vm)
+
+
+def insert_cd_safe(vm: VM, vdi_name: str, cd_path="D:/", retries=2):
+    """
+    Insert a CD with retry.
+
+    HACK: Windows VM may not detect the CD drive in some cases.
+    If this happens, reboot the VM in hopes that it will be detected.
+    """
+    for _attempt in range(retries):
+        vm.insert_cd(vdi_name)
+        # wait_for(vm.file_exists) doesn't handle SSHCommandFailed;
+        # we need to check for it via wait_for(vm.is_ssh_up).
+        wait_for_vm_running_and_ssh_up_without_tools(vm)
+        try:
+            wait_for(lambda: vm.file_exists(cd_path, regular_file=False), timeout_secs=60)
+            return
+        except TimeoutError:
+            logging.warning(f"Waiting for CD at {cd_path} failed, retrying by rebooting VM")
+            # There might be no VM tools so use SSH instead.
+            vm.ssh(WINDOWS_SHUTDOWN_COMMAND)
+            wait_for(vm.is_halted, "Wait for VM halted")
+            vm.eject_cd()
+            vm.start()
+            wait_for_vm_running_and_ssh_up_without_tools(vm)
+    raise TimeoutError(f"Waiting for CD at {cd_path} failed")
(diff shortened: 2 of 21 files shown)


tests/guest-tools/win/test_guest_tools_win.py
```diff
@@ -0,0 +1,346 @@
...
+    )
+    wait_for(vm.is_halted, "Shutdown VM")
+    snapshot = vm.snapshot()
+    yield (vm, snapshot)
+    snapshot.destroy(verify=True)
+
+
+@pytest.fixture
+def vm_instance(vm_prepared: tuple[VM, Snapshot]):
+    (vm, snapshot) = vm_prepared
```
stormi, line 198: AFAIK the usual convention is to omit the optional parentheses of the tuple when unpacking to variables (several instances of this in the file).

```suggestion
    vm, snapshot = vm_prepared
```

tests/guest-tools/win/test_guest_tools_win.py
```diff
@@ -0,0 +1,346 @@
...
+        ]
+    )
+    wait_for(vm.is_halted, "Shutdown VM")
+    snapshot = vm.snapshot()
+    yield (vm, snapshot)
+    snapshot.destroy(verify=True)
+
+
+@pytest.fixture
+def vm_instance(vm_prepared: tuple[VM, Snapshot]):
```
stormi, line 197:
```suggestion
def running_unsealed_windows_vm(vm_prepared: tuple[VM, Snapshot]):
```

tests/guest-tools/win/test_guest_tools_win.py
```diff
@@ -0,0 +1,346 @@
...
+    wait_for_guest_start(vm)
+    yield vm
+    snapshot.revert()
+
+
+@pytest.mark.multi_vms
+@pytest.mark.usefixtures("windows_vm")
+class TestGuestToolsWindows:
+    @pytest.fixture(scope="class")
+    def vm_install(self, vm_prepared):
```
stormi, line 209: I had to read the fixture definition to understand what `install` means in this context, so here also I'd go for a longer, more explicit name.

For example `vm_with_our_tools`?

tests/guest-tools/win/test_guest_tools_win.py
```diff
@@ -0,0 +1,346 @@
...
+        yield vm
+        snapshot.revert()
+
+    def test_tools_after_reboot(self, vm_install: VM):
+        vm = vm_install
+        assert is_drivers_installed(vm)
+
+    def test_drivers_detected(self, vm_install: VM):
+        vm = vm_install
+        assert vm.param_get("PV-drivers-detected")
```
stormi, line 223: We don't test the management agent. I suppose we'll expand the test later when we provide it with the drivers?

tests/guest-tools/win/test_guest_tools_win.py
```diff
@@ -0,0 +1,346 @@
+import enum
+import logging
+import pytest
+import time
+from lib.common import wait_for
+from lib.snapshot import Snapshot
+from lib.vm import VM
+from . import *
+
+
```
stormi, line 10: We describe the requirements on top of test files : pools, hosts, VMs, additional resources, specific configurations...

tests/guest-tools/win/test_guest_tools_win.py
```diff
@@ -0,0 +1,346 @@
...
+            vm,
+            "xcp-ng-9.0.9000\\package\\XenDrivers-x64.msi",
+            is_msi=True,
+        )
+        return vm
+
+    @pytest.fixture
+    def vm_install_wu_drivers(self, vm_prepared: VM):
+        (vm, snapshot) = vm_prepared
+        vm.param_set("has-vendor-device", True)
```
stormi, line 291: This makes me think you should probably verify that it's false, in the initial fixture. We attempt to detect failure causes as early as possible, as this is usually easier to understand when a test fails. Here, if someone attemps to run the test on a badly made XVA, or on an existing VM without realizing that the setting had to be false for the tests where we install our tools, the resulting error might not be obvious.

tests/guest-tools/win/test_guest_tools_win.py
```diff
@@ -0,0 +1,346 @@
...
+        vm.start()
+        wait_for_guest_start(vm)
+        install_other_drivers(
+            vm,
+            "citrix-9.4.0",
+            is_msi=False,
+        )
```
stormi, line 298: Does this install the drivers via Windows Update?

tests/guest-tools/win/test_guest_tools_win.py
```diff
@@ -0,0 +1,346 @@
...
+
+        install_other_drivers(
+            vm,
+            "xcp-ng-9.0.9000\\package\\XenDrivers-x64.msi",
+            is_msi=True,
+        )
+        return vm
+
+    @pytest.fixture
+    def vm_install_wu_drivers(self, vm_prepared: VM):
```
stormi, line 289: It's not obvious that "wu" means "Windows Update". Expand the acronym or add a comment?

tests/guest-tools/win/test_guest_tools_win.py
```diff
@@ -0,0 +1,346 @@
...
+            "powershell.exe",
+            "-noprofile",
+            "-noninteractive",
+            "Stop-Computer",
+            "-Force",
+        ]
+    )
+    wait_for(vm.is_halted, "Shutdown VM")
+    snapshot = vm.snapshot()
+    yield (vm, snapshot)
```
stormi, line 192: We've done such fixtures in the past, and this works well. However, restoring a snapshot is a rather costly operation, and it's possible that we would get better performance by using cloned VMs instead (usually very fast thanks to copy-on-write). And this simplifies the `yield` part as you just need to yield the cloned VM.

tests/guest-tools/win/test_guest_tools_win.py
```diff
@@ -0,0 +1,346 @@
...
+        exitcode = install_cert_and_tools(
+            vm_install_citrix_tools, PowerAction.Nothing, check=False
+        )
+        assert exitcode == ERROR_INSTALL_FAILURE
+
+    def test_install_with_xcpng_v8_tools(self, vm_install_xcpng_v8_tools: VM):
+        exitcode = install_cert_and_tools(
+            vm_install_xcpng_v8_tools, PowerAction.Nothing, check=False
+        )
+        assert exitcode == ERROR_INSTALL_FAILURE
```
stormi, line 335: Do you also envision tests where we start with either of the old v8 tools, xcpng_v9_tools, old citrix tools, recent citrix tools, remove them and replace them with the latest we built?

stormi (commented):
This will be a nice addition to the tests. Here are my review comments.

tests/guest-tools/win/__init__.py
```diff
@@ -0,0 +1,123 @@
...
+        "serial",
+        domid,
+        "|",
+        "grep",
+        r"~xcp-ng-tests~.*~end~",
+        "|",
+        "head",
+        "-n",
+        "1",
+    ]
```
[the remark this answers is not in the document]
dinhngtu: `commands.py _ssh()` does a `command = " ".join(cmd)` unconditionally, which breaks if `cmd` is a string. Should I add a check of `if cmd is not str` there?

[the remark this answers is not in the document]
stormi: Yes, I suppose we could check the type.

tests/guest-tools/win/__init__.py
```diff
@@ -0,0 +1,123 @@
...
+    logging.debug(f"Installed Xen device status: {output}")
+    if output == "CM_PROB_NONE":
+        return True
+    elif output == "CM_PROB_FAILED_INSTALL":
+        return False
+    else:
+        raise Exception(f"Unknown problem status {output}")
+
+
+def run_command_powershell(vm: VM, program: str, args: List[str]):
```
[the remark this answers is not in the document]
dinhngtu: It uses `Start-Process` to run a single executable, and cannot run a script unlike `powershell.exe -encodedcommand`. I'll add a separate `vm.execute_powershell_script` to this effect.

tests/guest-tools/win/__init__.py
```diff
@@ -0,0 +1,123 @@
...
+                "-noprofile",
+                "-noninteractive",
+                "-encodedcommand",
+                cmd,
+            ]
+        )
+    )
+
+
+def start_background_powershell(vm: VM, cmd: str):
```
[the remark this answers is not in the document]
dinhngtu: The main motivation was to retrieve exit codes higher than 255 (which ssh didn't like) and to avoid Git Bash's path escaping behavior, which undesirably transforms arguments (e.g. `/x` -> `x:/`)

tests/guest-tools/win/test_guest_tools_win.py
dinhngtu: It installs Citrix drivers with the vendor device enabled to mimic situations where PV drivers are installed via Windows Update. I'll rename the fixture.

tests/guest-tools/win/test_guest_tools_win.py
```diff
@@ -0,0 +1,346 @@
...
+    start_background_powershell(vm, ";".join(uninstall_cmd))
+    if action != PowerAction.Nothing:
+        wait_for(vm.is_halted, "Wait for VM halted")
+    if action == PowerAction.Reboot:
+        vm.start()
+        wait_for_guest_start(vm)
+
+
+def install_cert_and_tools(vm: VM, action: PowerAction, check: bool = True):
+    vm.insert_cd("guest-tools-win.iso")
```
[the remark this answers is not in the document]
dinhngtu: Would the Windows tools be tested as part of the XCP-ng package or separately from it? If it's the latter then IMO it's better to download the guest ISO like the other-tools one.

tests/guest-tools/win/test_guest_tools_win.py
dinhngtu: I've tried that, but came back to the snapshot approach since there's no need to rediscover VM IP after every run. I'm also not sure if deleting a cloned VM is faster than restoring from a snapshot since both are COW-based anyway.

tests/guest-tools/win/test_guest_tools_win.py
dinhngtu: Some tools (e.g. xcp-ng v8) don't uninstall cleanly and crash the VM. I'll try to add scenarios where some xenbus/etc. drivers are already installed (without the management packages).

stormi: Could we use the clean-up tool? This would allow to test it in the process, too.

benjamreis (approved):
LGTM! :+1:

dinhngtu:
I'm doing some refactoring to make testing multiple guest tools/other tools versions a bit easier. Currently running final tests.

dinhngtu:
I've refactored the tests as following:

- The ISO files of our test tools and other tools could be configured in `data.py` (see `WIN_GUEST_TOOLS_ISOS` and `OTHER_GUEST_TOOLS_ISO`).
- Each ISO configuration now has a `download` parameter that lets the test use an included ISO file or download it from a HTTP location into a local ISO SR.
- Each configuration in `WIN_GUEST_TOOLS_ISOS` specifies a ISO location, whether it requires testsigning, MSI file name, etc.
- We can configure the contents of our other tools ISO file via `OTHER_GUEST_TOOLS` using a similar method.
- All guest and other tools versions now share the same parameterized tests.
- I've split guest-tools/win into several separate module files.
- XenClean tests have been added.

stormi (commented):
Looks good. Just a few minor comments.

I will delay my approval until we have setup the tests in our CI so that they run once merged on master.

jobs.py
```diff
@@ -350,6 +350,19 @@
         "paths": ["tests/guest-tools/unix"],
         "markers": "multi_vms",
     },
+    "tools-windows": {
+        "description": "tests our windows guest tools on a variety of VMs",
+        "requirements": [
+            "A pool >= 8.3. One host is enough.",
```
gduperrey, line 356: Does it work in 8.2?

dinhngtu: Yes, it should work.

dinhngtu:
I've added some tools and scripts useful for reducing Windows image sizes.

dinhngtu:
I've switched the ISO download code to use the existing NFS ISO SR config.

dinhngtu:
Changes since last time:
- Reworked `OTHER_GUEST_TOOLS is_msi` property into a nullable enum with values `msi/inf/None`. Separate the "vendor/Windows Update" case and the "Citrix inf" case since these are actually 2 different cases.
- Added several hacks to work around edge cases that caused test failures (e.g. CD insertion hang, CD drive not detected).
- Added the commit "lib/host: Allow specifying default SR UUID" to ease running tests on specific hardware (e.g. particular storage types).
- Added `scripts/guests/windows/install-drivers.ps1`.
- Refactor Windows shutdown command into a constant.

Upgrades are marked as xfail since they require multiple reboots, which I haven't yet taken into account.

gduperrey:
CI:
* 8.2: :white_check_mark:
* 8.3: :white_check_mark:

gduperrey (approved):

stormi:
Thanks for a good PR!

PR #263, opened 2024-11-22
Add a test about keeping management address type when disabling

benjamreis:
Before it was forced to change into IPv4 and would break IPv6 setups

### DIFF
--- a/tests/network/test_management_disable_address_type.py
+++ b/tests/network/test_management_disable_address_type.py
@@ -1,0 +1,12 @@
+# Requirements:
+# - one XCP-ng host (--host) (>= 8.3 for IPv6 test)
+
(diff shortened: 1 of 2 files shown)


tests/network/test_management_disable_address_type.py
```diff
@@ -0,0 +1,23 @@
+import pytest
+
+# Requirements:
+# - one XCP-ng host (--host) (>= 8.3 for IPv6 test)
```
stormi, line 4: Should it require the fixture which skips tests when the version is not appropriate?

benjamreis: Fixed by only having on test we'll run on 2 different hosts.

stormi (approved):

PR #267, opened 2025-01-10
Improvements to set_xva_bridge.sh

dinhngtu:
This PR introduces a few improvements to `set_xva_bridge.sh`:

- Fix cases where XVA_NAME contains spaces
- Get TMP_LIST from original archive instead of sorting to avoid situations where `sort` doesn't put ova.xml in the correct order
- Set owner and permissions in tar command instead of using `sudo` to read unreadable files
- Use `file -b` to simplify filetype parsing
- Don't check existing bridge name to avoid unnecessary tar pass

### DIFF
--- a/scripts/set_xva_bridge.sh
+++ b/scripts/set_xva_bridge.sh
@@ -105,10 +105,10 @@ else
     exit 1
 fi
 
+chmod -R u+rX "${TMPFOLDER}"
+
 if [ -e "${TMPFOLDER}/ova.xml" ]; then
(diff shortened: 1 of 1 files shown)


stormi (approved):

stormi:
Ping @gduperrey

gduperrey:
Still haven't taken the time to make new xvas to validate.

gduperrey (approved):
Tested on several new xva. It works well. Thanks for this update.
Sorry for the delay for my validation :)

PR #269, opened 2025-01-30
refactor: Use Dict for type annotation in PHASE_REPORT_KEY for better compatibility

### LINKED ISSUE #268: Unable to run pytest with platform linux -- Python 3.8.13, pytest-8.3.4, pluggy-1.5.0
Ran into an error as below

```
[root@06371cf799bb xcp-ng-tests]# pytest tests/misc/test_vm_basic_operations.py --hosts=x.x.x.x
ImportError while loading conftest '/root/xcp-ng-tests/conftest.py'.
conftest.py:122: in <module>
    PHASE_REPORT_KEY = pytest.StashKey[dict[str, pytest.CollectReport]]()
E   TypeError: 'type' object is not subscriptable
```

Made changes to `contest.py` for it to work
 
```
[root@06371cf799bb xcp-ng-tests]# git diff
diff --git a/conftest.py b/conftest.py
index 3c176a4..bc562c4 100644
--- a/conftest.py
+++ b/conftest.py
@@ -119,7 +119,8 @@ def pytest_collection_modifyitems(items, config):
 # FIXME we may have to move this into lib/ if fixtures in sub-packages
 # want to make use of this feature
 
-PHASE_REPORT_KEY = pytest.StashKey[dict[str, pytest.CollectReport]]()
+from typing import Dict
+PHASE_REPORT_KEY = pytest.StashKey[Dict[str, pytest.CollectReport]]()
 @pytest.hookimpl(wrapper=True, tryfirst=True)
 def pytest_runtest_makereport(item, call):
     # execute all other hooks to obtain the report object
```

The type hint `dict[str, pytest.CollectReport]` uses subscripting ([]), which requires Python 3.9 or later for built-in types like dict.

Ref commit : https://github.com/xcp-ng/xcp-ng-tests/commit/958e434927103b7e75771533aecd1e66535b0060

rushikeshjadhav:
- Use Dict for type annotation in PHASE_REPORT_KEY for better compatibility (Fixes #268)

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -4,6 +4,7 @@ import pytest
 import tempfile
 
 from packaging import version
+from typing import Dict
 
 import lib.config as global_config
 
@@ -119,7 +120,7 @@ def pytest_collection_modifyitems(items, config):
 # FIXME we may have to move this into lib/ if fixtures in sub-packages
 # want to make use of this feature
 
-PHASE_REPORT_KEY = pytest.StashKey[dict[str, pytest.CollectReport]]()
+PHASE_REPORT_KEY = pytest.StashKey[Dict[str, pytest.CollectReport]]()
 @pytest.hookimpl(wrapper=True, tryfirst=True)
 def pytest_runtest_makereport(item, call):
     # execute all other hooks to obtain the report object


stormi:
> Use Dict for type annotation in PHASE_REPORT_KEY for better compatibility 

Let's make it a separate PR.

rushikeshjadhav:
> > Use Dict for type annotation in PHASE_REPORT_KEY for better compatibility
> 
> Let's make it a separate PR.

Amended this PR.

ydirson (approved):

stormi:
Adding a new CLI parameter such as `--provisioning` is a last resort change, because it moves the responsibility for defining it to the caller, when we probably want that our tests, by default, provide both testing for thick and thin provisioning, as their *definition*. 

Couldn't we have test definitions that do both `thick` and `thin`, instead?

stormi (approved):

PR #270, opened 2025-02-05
feat(storage/linstor): Enhance multi-disk support and provisioning flexibility in Linstor SR tests

rushikeshjadhav:
- Added `sr_disks_for_all_hosts` fixture to support multiple disks, ensuring availability across all hosts and handling "auto" selection dynamically.
- Updated `lvm_disk` fixture to provision multiple disks collectively, refining vgcreate and pvcreate logic.
- Introduced `provisioning_type` and `storage_pool_name` fixtures to dynamically configure storage provisioning (thin or thick).
- Refactored Linstor SR test cases to use the new fixtures, improving test coverage across provisioning types.
- Optimized Linstor installation and cleanup using concurrent execution, reducing setup time.
- Enhanced validation and logging for disk selection.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -329,6 +329,40 @@ def sr_disk_for_all_hosts(pytestconfig, request, host):
         logging.info(f">> Disk or block device {disk} is present and free on all pool members")
     yield candidates[0]
 
+@pytest.fixture(scope='session')
+def sr_disks_for_all_hosts(pytestconfig, request, host):
+    disks = pytestconfig.getoption("sr_disk")
+    assert len(disks) > 0, "This test requires at least one --sr-disk parameter"
+    # Fetch available disks on the master host
+    master_disks = host.available_disks()
+    assert len(master_disks) > 0, "a free disk device is required on the master host"
+
+    if "auto" in disks:
+        candidates = list(master_disks)
+    else:
+        # Validate that all specified disks exist on the master host
+        for disk in disks:
+            assert disk in master_disks, \
+                f"Disk or block device {disk} is either not present or already used on the master host"
+        candidates = list(disks)
+
+    # Check if all disks are available on all hosts in the pool
+    for h in host.pool.hosts[1:]:
+        other_disks = h.available_disks()
+        candidates = [d for d in candidates if d in other_disks]
+
+    if "auto" in disks:
+        # Automatically select disks if "auto" is passed
+        assert len(candidates) > 0, \
+            f"Free disk devices are required on all pool members. Pool master has: {' '.join(master_disks)}."
+        logging.info(">> Using free disk device(s) on all pool hosts: %s.", candidates)
+    else:
+        # Ensure specified disks are free on all hosts
+        assert len(candidates) == len(disks), \
+            f"Some specified disks ({', '.join(disks)}) are not free or available on all hosts."
+        logging.info(">> Disk(s) %s are present and free on all pool members", candidates)
+    yield candidates
+
 @pytest.fixture(scope='module')
 def vm_ref(request):
     ref = request.param
--- a/tests/storage/linstor/conftest.py
+++ b/tests/storage/linstor/conftest.py
@@ -41,6 +41,14 @@ def lvm_disks(host, sr_disks_for_all_hosts, provisioning_type):
         for device in devices:
             host.ssh(['pvremove', device])
 
+@pytest.fixture(scope="package")
+def storage_pool_name(provisioning_type):
+    return GROUP_NAME if provisioning_type == "thick" else STORAGE_POOL_NAME
+
+@pytest.fixture(params=["thin", "thick"], scope="session")
+def provisioning_type(request):
+    return request.param
+
 @pytest.fixture(scope='package')
 def pool_with_linstor(hostA2, lvm_disks, pool_with_saved_yum_state):
     import concurrent.futures
@@ -79,11 +87,11 @@ def pool_with_linstor(hostA2, lvm_disks, pool_with_saved_yum_state):
         executor.map(remove_linstor, pool.hosts)
 
 @pytest.fixture(scope='package')
-def linstor_sr(pool_with_linstor):
+def linstor_sr(pool_with_linstor, provisioning_type, storage_pool_name):
     sr = pool_with_linstor.master.sr_create('linstor', 'LINSTOR-SR-test', {
-        'group-name': STORAGE_POOL_NAME,
+        'group-name': storage_pool_name,
         'redundancy': str(min(len(pool_with_linstor.hosts), 3)),
-        'provisioning': 'thin'
+        'provisioning': provisioning_type
     }, shared=True)
     yield sr
     sr.destroy()
--- a/tests/storage/linstor/conftest.py
+++ b/tests/storage/linstor/conftest.py
@@ -12,31 +12,34 @@ LINSTOR_RELEASE_PACKAGE = 'xcp-ng-release-linstor'
 LINSTOR_PACKAGE = 'xcp-ng-linstor'
 
 @pytest.fixture(scope='package')
-def lvm_disk(host, sr_disk_for_all_hosts):
-    device = '/dev/' + sr_disk_for_all_hosts
+def lvm_disks(host, sr_disks_for_all_hosts, provisioning_type):
+    devices = [f"/dev/{disk}" for disk in sr_disks_for_all_hosts]
     hosts = host.pool.hosts
 
     for host in hosts:
-        try:
-            host.ssh(['pvcreate', '-ff', '-y', device])
-        except commands.SSHCommandFailed as e:
-            if e.stdout.endswith('Mounted filesystem?'):
-                host.ssh(['vgremove', '-f', GROUP_NAME, '-y'])
+        for device in devices:
+            try:
                 host.ssh(['pvcreate', '-ff', '-y', device])
-            elif e.stdout.endswith('excluded by a filter.'):
-                host.ssh(['wipefs', '-a', device])
-                host.ssh(['pvcreate', '-ff', '-y', device])
-            else:
-                raise e
+            except commands.SSHCommandFailed as e:
+                if e.stdout.endswith('Mounted filesystem?'):
+                    host.ssh(['vgremove', '-f', GROUP_NAME, '-y'])
+                    host.ssh(['pvcreate', '-ff', '-y', device])
+                elif e.stdout.endswith('excluded by a filter.'):
+                    host.ssh(['wipefs', '-a', device])
+                    host.ssh(['pvcreate', '-ff', '-y', device])
+                else:
+                    raise e
 
-        host.ssh(['vgcreate', GROUP_NAME, device])
-        host.ssh(['lvcreate', '-l', '100%FREE', '-T', STORAGE_POOL_NAME])
+        host.ssh(['vgcreate', GROUP_NAME] + devices)
+        if provisioning_type == 'thin':
+            host.ssh(['lvcreate', '-l', '100%FREE', '-T', STORAGE_POOL_NAME])
 
-    yield device
+    yield devices
 
     for host in hosts:
         host.ssh(['vgremove', '-f', GROUP_NAME])
-        host.ssh(['pvremove', device])
+        for device in devices:
+            host.ssh(['pvremove', device])
 
 @pytest.fixture(scope='package')
 def pool_with_linstor(hostA2, lvm_disk, pool_with_saved_yum_state):
--- a/tests/storage/linstor/conftest.py
+++ b/tests/storage/linstor/conftest.py
@@ -42,15 +42,21 @@ def lvm_disks(host, sr_disks_for_all_hosts, provisioning_type):
             host.ssh(['pvremove', device])
 
 @pytest.fixture(scope='package')
-def pool_with_linstor(hostA2, lvm_disk, pool_with_saved_yum_state):
+def pool_with_linstor(hostA2, lvm_disks, pool_with_saved_yum_state):
+    import concurrent.futures
     pool = pool_with_saved_yum_state
-    for host in pool.hosts:
+
+    def check_linstor_installed(host):
         if host.is_package_installed(LINSTOR_PACKAGE):
             raise Exception(
                 f'{LINSTOR_PACKAGE} is already installed on host {host}. This should not be the case.'
             )
 
-    for host in pool.hosts:
+    with concurrent.futures.ThreadPoolExecutor() as executor:
+        executor.map(check_linstor_installed, pool.hosts)
+
+    def install_linstor(host):
+        logging.info(f"Installing {LINSTOR_PACKAGE} on host {host}...")
         host.yum_install([LINSTOR_RELEASE_PACKAGE])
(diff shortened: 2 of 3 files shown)


stormi:
Adding a new CLI parameter such as `--provisioning` is a last resort change, because it moves the responsibility for defining it to the caller, when we probably want that our tests, by default, provide both testing for thick and thin provisioning, as their *definition*. 

Couldn't we have test definitions that do both `thick` and `thin`, instead?

rushikeshjadhav:
> Couldn't we have test definitions that do both `thick` and `thin`, instead?

We can do that. Since this is XOSTOR level, one full cycle (create-test-destroy) with `thin` and then another full cycle with `thick` can be added. 

Can we retain the CLI option for those callers who want to selectively run in one of the mode?

stormi:
> Can we retain the CLI option for those callers who want to selectively run in one of the mode?

In theory pytest already offers enough selectors to execute a set of tests and not others.

tests/storage/linstor/test_linstor_sr.py
```diff
@@ -157,10 +157,17 @@ def test_diskless_kept(self, host, linstor_sr, vm_on_linstor_sr):
         for member in host.pool.hosts:
             controller_option += f"{member.hostname_or_ip},"
 
+        # Determine the correct group name based on provisioning type
+        group_mapping = {
+            "thick": "xcp-sr-linstor_group_device",
+            "thin": "xcp-sr-linstor_group_thin_device"
```
Wescoeur, line 163: Use conftest.py here, we already have these helpers:
https://github.com/xcp-ng/xcp-ng-tests/blob/134fd551c4d2b5827b159da067020ddef79ac643/tests/storage/linstor/conftest.py#L9-L10

Wescoeur: Also `xcp-sr-linstor_group_device` is not valid, it's just `xcp-sr-linstor_group_device` for thick provisioning.

tests/storage/linstor/test_linstor_sr.py
```diff
@@ -36,13 +36,13 @@ def test_create_sr_without_linstor(self, host, lvm_disk):
...
-    def test_create_and_destroy_sr(self, pool_with_linstor):
+    def test_create_and_destroy_sr(self, pool_with_linstor, provisioning_type, storage_pool_name):
         # Create and destroy tested in the same test to leave the host as unchanged as possible
         master = pool_with_linstor.master
         sr = master.sr_create('linstor', 'LINSTOR-SR-test', {
-            'group-name': STORAGE_POOL_NAME,
-            'redundancy': '1',
-            'provisioning': 'thin'
+            'group-name': storage_pool_name,
+            'redundancy': '2',
```
Wescoeur, line 44: Same question here.

tests/storage/linstor/test_linstor_sr.py
```diff
@@ -18,15 +18,15 @@ class TestLinstorSRCreateDestroy:
...
         # This test must be the first in the series in this module
         assert not host.is_package_installed('python-linstor'), \
             "linstor must not be installed on the host at the beginning of the tests"
         try:
             sr = host.sr_create('linstor', 'LINSTOR-SR-test', {
-                'group-name': STORAGE_POOL_NAME,
-                'redundancy': '1',
-                'provisioning': 'thin'
+                'group-name': storage_pool_name,
+                'redundancy': '2',
```
Wescoeur, line 28: Is there a specific reason to set 2 in your tests?

conftest.py
```diff
@@ -329,6 +329,41 @@ def sr_disk_for_all_hosts(pytestconfig, request, host):
         logging.info(f">> Disk or block device {disk} is present and free on all pool members")
     yield candidates[0]
 
+@pytest.fixture(scope='session')
+def sr_disks_for_all_hosts(pytestconfig, request, host):
+    disks = pytestconfig.getoption("sr_disk")
+    assert len(disks) > 0, "This test requires at least one --sr-disk parameter"
+    # Fetch available disks on the master host
+    master_disks = host.available_disks()
+    assert len(master_disks) > 0, "a free disk device is required on the master host"
```
Wescoeur, line 338: Maybe `A` instead of `a`  to uniformize.

tests/storage/linstor/conftest.py
```diff
@@ -12,57 +12,85 @@
 LINSTOR_PACKAGE = 'xcp-ng-linstor'
 
 @pytest.fixture(scope='package')
-def lvm_disk(host, sr_disk_for_all_hosts):
-    device = '/dev/' + sr_disk_for_all_hosts
+def lvm_disk(host, sr_disks_for_all_hosts, provisioning_type):
```
Wescoeur, line 15: Maybe `lvm_disks` if you change the logic here. :)

tests/storage/linstor/conftest.py
```diff
@@ -12,57 +12,85 @@
...
+        logging.info(f"Installing {LINSTOR_PACKAGE} on host {host}...")
         host.yum_install([LINSTOR_RELEASE_PACKAGE])
         host.yum_install([LINSTOR_PACKAGE], enablerepo="xcp-ng-linstor-testing")
         # Needed because the linstor driver is not in the xapi sm-plugins list
         # before installing the LINSTOR packages.
         host.ssh(["systemctl", "restart", "multipathd"])
         host.restart_toolstack(verify=True)
 
+    with concurrent.futures.ThreadPoolExecutor() as executor:
+        executor.map(install_linstor, pool.hosts)
```
Wescoeur, line 77: No particular remark, just: very good idea! :smiley:

Wescoeur (changes requested):

tests/storage/linstor/test_linstor_sr.py
rushikeshjadhav: Right, using `GROUP_NAME` is better.

conftest.py
rushikeshjadhav: Sure

tests/storage/linstor/conftest.py
rushikeshjadhav: Yes :)

tests/storage/linstor/conftest.py
rushikeshjadhav: I was running out of patience. 😃

tests/storage/linstor/test_linstor_sr.py
Wescoeur: I think we must handle this error in the driver itself. In theory a simple busy loop should fix that.

rushikeshjadhav: Thanks, will keep it as `1`.

rushikeshjadhav:
> Will this PR involve changes in linstor job definitions in `jobs.py` for the extended test coverage to be effective in Jenkins once this is merged?

After review of `jobs.py` it seems no changes are needed. Including a snippet for reference on "params" below. 

Earlier we had `--provisioning` but that was removed and now the tests will run for both `thin` and `thick` as both should be tested.

```
    "linstor-main": {
        "description": "tests the linstor storage driver, but avoids migrations and reboots",
        "requirements": [
            "A pool with at least 3 hosts.",
            "An additional free disk on every host.",
            "A small VM that can be imported on the SR.",
        ],
        "nb_pools": 1,
        "params": {
            "--vm": "single/small_vm",
            "--sr-disk": "auto",
        },
        "paths": ["tests/storage/linstor"],
        "markers": "(small_vm or no_vm) and not reboot and not quicktest",
        "name_filter": "not migration",
    },
    "linstor-migrations": {
        "description": "tests migrations with the linstor storage driver",
        "requirements": [
            "A pool with at least 3 hosts.",
            "An additional free disk on every host.",
            "A second pool with at least 1 host and a SR to receive VMs.",
            "A small VM that can be imported on the SRs.",
        ],
        "nb_pools": 2,
        "params": {
            "--vm": "single/small_vm",
            "--sr-disk": "auto",
        },
        "paths": ["tests/storage/linstor"],
        "markers": "",
        "name_filter": "migration",
    },
    "linstor-reboots": {
        "description": "linstor storage driver tests that involve rebooting hosts",
        "requirements": [
            "A pool with at least 3 hosts, whose master host can be rebooted (best if reboots fast).",
            "An additional free disk on every host.",
            "A small VM that can be imported on the SRs.",
        ],
        "nb_pools": 1,
        "params": {
            "--vm": "single/small_vm",
            "--sr-disk": "auto",
        },
        "paths": ["tests/storage/linstor"],
        "markers": "reboot",
    },
    "linstor-quicktest": {
        "description": "runs `quicktest` on the linstor storage driver`",
        "requirements": [
            "A pool with at least 3 hosts.",
            "An additional free disk on every host.",
        ],
        "nb_pools": 1,
        "params": {
            "--sr-disk": "auto",
        },
        "paths": ["tests/storage/linstor"],
        "markers": "quicktest",
    },
```

tests/storage/linstor/test_linstor_sr.py
```diff
@@ -65,7 +65,7 @@ def test_start_and_shutdown_VM(self, vm_on_linstor_sr):
         vm = vm_on_linstor_sr
         vm.start()
         vm.wait_for_os_booted()
-        vm.shutdown(verify=True)
+        vm.shutdown(verify=True, force_if_fails=True)
```
Nambrok, line 68: Why is `force_if_fails` added here? We might want to check that the VM normal shutdown can happen in the test.
Was it something that was discussed that I missed?

Nambrok (changes requested):

conftest.py
```diff
@@ -329,6 +329,41 @@ def sr_disk_for_all_hosts(pytestconfig, request, host):
...
+def sr_disks_for_all_hosts(pytestconfig, request, host):
+    disks = pytestconfig.getoption("sr_disk")
+    assert len(disks) > 0, "This test requires at least one --sr-disk parameter"
+    # Fetch available disks on the master host
+    master_disks = host.available_disks()
+    assert len(master_disks) > 0, "A free disk device is required on the master host"
+
+    if "auto" not in disks:
+        # Validate that all specified disks exist on the master host
+        for disk in disks:
```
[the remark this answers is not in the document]
rushikeshjadhav: Yes, we can update code and handle disks for `--sr-disk=nvme1n1,nvme2n1` scenario.  

However, it seemed `sr-disk` is used in other places as well and did not want to break/alter that for compatibility.
Also, if we are handling multiple disks in one option then the option will become `--sr-disks` instead of `--sr-disk` on same line as `hosts`.  And this will cause ripple effect at other places.. The current code will gracefully handle both cases, so can I put a different PR for this (`--sr-disks`) or you prefer to change in current one?

In `jobs.py`, `"--sr-disk": "auto"` is used which lets the current code pick multiple available disks and stay intact.

tests/storage/linstor/conftest.py
```diff
@@ -12,57 +12,85 @@
...
                 host.ssh(['pvcreate', '-ff', '-y', device])
-            elif e.stdout.endswith('excluded by a filter.'):
-                host.ssh(['wipefs', '-a', device])
-                host.ssh(['pvcreate', '-ff', '-y', device])
-            else:
-                raise e
+            except commands.SSHCommandFailed as e:
+                if e.stdout.endswith('Mounted filesystem?'):
+                    host.ssh(['vgremove', '-f', GROUP_NAME, '-y'])
+                    host.ssh(['pvcreate', '-ff', '-y', device])
```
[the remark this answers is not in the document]
rushikeshjadhav: The goal here was to `pvcreate` on device so that we can `vgcreate` at a later step L34. In its teardown you will see `pvremove`. I think this can further be optimised if #276 is to be taken care of. The idea is to ensure that the device is raw and available to use for making Linstor.

tests/storage/linstor/test_linstor_sr.py
rushikeshjadhav: Right, the test purpose is getting diluted. Will revert this.

conftest.py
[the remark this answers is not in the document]
Nambrok: It's mostly used in auto, it's just that if we are expected to give multiple disks to a test, using `--sr-disk` with comma separated value make more sense.
If we use multiple value, we expect the test to be run multiple time using the values one after the other like it is doing with `--vm`.
For this case, we don't have to go further than having comma-separated value working on `sr-disk` for your new fixture (and maybe) ignore the other value after the first disk for the already existing fixtures using `sr-disk` expecting only one disk.
For renaming the parameter completely, I'm summoning @Wescoeur and @stormi to give their idea on the matter.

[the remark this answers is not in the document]
ydirson: What's the benefit of parsing a comma-separated list, whereas `pytest` already gives us a way to specify `--sr-disk=nvme1n1 --sr-disk=nvme2n1` ?

ydirson (changes requested):
The `lvm_disks` part looks like a good candidate to move into a separate first commit

Nambrok (approved):

tests/storage/linstor/conftest.py
```diff
@@ -12,57 +12,85 @@
...
         host.ssh(['vgremove', '-f', GROUP_NAME])
-        host.ssh(['pvremove', device])
+        for device in devices:
+            host.ssh(['pvremove', device])
+
+@pytest.fixture(scope="package")
+def storage_pool_name(provisioning_type):
+    return GROUP_NAME if provisioning_type == "thick" else STORAGE_POOL_NAME
+
+@pytest.fixture(params=["thin", "thick"], scope="session")
```
ydirson, line 49: Won't the `session` scope here prevent to run both thick and thin tests in the same run?  Is that restriction necessary?

ydirson (changes requested):

conftest.py
```diff
@@ -329,6 +329,41 @@ def sr_disk_for_all_hosts(pytestconfig, request, host):
...
+            assert disk in master_disks, \
+                f"Disk or block device {disk} is either not present or already used on the master host"
+        master_disks = [disk for disk in disks if disk in master_disks]
```
ydirson, line 345: It would be more clear here to set `candidates` directly, and have the `candidates = list(master_disks)` be in an `else` statement (avoid overloading of `master_disks`)

tests/storage/linstor/conftest.py
```diff
@@ -12,57 +12,85 @@
...
+        executor.map(install_linstor, pool.hosts)
+
     yield pool
 
+    def remove_linstor(host):
+        logging.info(f"Cleaning up python-linstor from host {host}...")
+        host.yum_remove(["python-linstor"])
+
+    with concurrent.futures.ThreadPoolExecutor() as executor:
+        executor.map(remove_linstor, pool.hosts)
```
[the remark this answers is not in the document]
ydirson: OK, addressed in comment

ydirson:
Split the post-review changes into fixup commits as discussed offline

Wescoeur (approved):

rushikeshjadhav:
@stormi @ydirson can you review and help to close this?

stormi:
Both Damien and Ronan approved, and Yann won't be available in the coming days, so let's merge. Thanks!

stormi:
@gduperrey as soon as you find a slot, and as long as it doesn't delay anything 8.3-related, please run the linstor jobs from this branch to validate it.

stormi (approved):

PR #273, opened 2025-02-12
lib/vdi: Add VDI resize support and integrate into VM & linstor operations

rushikeshjadhav:
- Added `resize` method to `VDI` class to allow resizing of VDIs via `xe vdi-resize`.
- Introduced `vdi_resize` method in VM class to support 2x VDI resizing.
- Implemented `test_resize_vdi` in Linstor and basic VM tests, verifying VDI resize functionality and ensuring VM boot post-resize.
- Refactored VM operations to include `wait_for_vm_running`.

### DIFF
--- a/lib/vm.py
+++ b/lib/vm.py
@@ -190,6 +190,9 @@ class VM(BaseVM):
         self.wait_for_os_booted()
         wait_for(self.is_ssh_up, "Wait for SSH up")
 
+    def wait_for_vm_running(self):
+        wait_for(self.is_running)
+
     def ssh_touch_file(self, filepath):
         logging.info("Create file on VM (%s)" % filepath)
         self.ssh(['touch', filepath])
@@ -506,6 +509,24 @@ class VM(BaseVM):
             self.ssh(['test ! -f ' + filepath])
         finally:
             snapshot.destroy(verify=True)
+
+    def vdi_resize(self, vdi_uuid: str, new_size: Optional[int] = None):
+        # Reference
+        # MAX_VHD_SIZE = 2088960 * 1024 * 1024 # 2088960 MB
+        # MIN_VHD_SIZE = 1 * 1024 * 1024 # 1MB
+
+        # Following assert may be removed in future if we support online vdi resize
+        assert self.is_halted(), "Expected halted, got running"
+
+        vdi = VDI(vdi_uuid, sr=self.get_sr())
+        if new_size is None:
+            # Resize by double, ideally it should be within min and max limits reference
+            vdi_size = int(vdi.param_get("virtual-size"))
+            new_size = vdi_size * 2
+
+        vdi.resize(new_size)
+        resized_vdi_size = int(vdi.param_get("virtual-size"))
+        assert resized_vdi_size == new_size, f"Expected {new_size}, got {resized_vdi_size}"
 
     def get_messages(self, name):
         args = {
--- a/tests/storage/linstor/conftest.py
+++ b/tests/storage/linstor/conftest.py
@@ -100,14 +100,14 @@ def pool_with_linstor(
     import concurrent.futures
     pool = pool_with_saved_yum_state
 
-    def check_linstor_installed(host: Host):
+    def assert_linstor_not_installed(host):
         if host.is_package_installed(LINSTOR_PACKAGE):
             raise Exception(
                 f'{LINSTOR_PACKAGE} is already installed on host {host}. This should not be the case.'
             )
 
     with concurrent.futures.ThreadPoolExecutor() as executor:
-        executor.map(check_linstor_installed, pool.hosts)
+        executor.map(assert_linstor_not_installed, pool.hosts)
 
     def install_linstor(host: Host):
         logging.info(f"Installing {LINSTOR_PACKAGE} on host {host}...")
--- a/tests/storage/linstor/test_linstor_sr.py
+++ b/tests/storage/linstor/test_linstor_sr.py
@@ -88,6 +88,15 @@ class TestLinstorSR:
         finally:
             vm.shutdown(verify=True)
 
+    @pytest.mark.small_vm
+    @pytest.mark.big_vm
+    def test_resize_vdi(self, vm_on_linstor_sr):
+        vm = vm_on_linstor_sr
+        logging.info("VDI Resize started")
+        for vdi_uuid in vm.vdi_uuids():
+            vm.vdi_resize(vdi_uuid)
+        logging.info("VDI Resize completed")
+
     # *** tests with reboots (longer tests).
 
     @pytest.mark.reboot


lib/vm.py
```diff
@@ -361,6 +365,25 @@ def test_snapshot_on_running_vm(self):
         finally:
             snapshot.destroy(verify=True)
 
+    def vdi_resize(self, vdi_uuid, new_size=None):
+        # Should MIN MAX be handled or let it throw exception?
```
Wescoeur, line 369: I think we should raise here.
We will soon add QCOW2 tests and it could be an opportunity to test sizes larger than 2TB. In parallel we could have a VHD specific test to check that if we exceed the limit, it does not work.

Wescoeur (changes requested):

lib/vm.py
rushikeshjadhav: Done

lib/vm.py
```diff
@@ -361,6 +365,26 @@ def test_snapshot_on_running_vm(self):
         finally:
             snapshot.destroy(verify=True)
 
+    def vdi_resize(self, vdi_uuid, new_size=None):
+        # Reference
+        MAX_VHD_SIZE = 2088960 * 1024 * 1024 # 2088960 MB
+        MIN_VHD_SIZE = 1 * 1024 * 1024 # 1MB
+
+        if self.is_running():
+            self.shutdown(verify=True, force_if_fails=True)
```
stormi, line 374: I'm not sure about that. We've been tempted in the past to do smart things like that but we regretted it and came back to simple things, such as asserts. I think we should simply not call vdi_resize on a running VM at the moment, or, when we do, let it be so that we check that we get an appropriate error message.

Later, when we actually support resizing VDIs without turning the VMs off, we'd have to change this anyway.

tests/misc/test_basic_without_ssh.py
```diff
@@ -75,6 +75,22 @@ def test_checkpoint(self, imported_vm):
         finally:
             snapshot.destroy(verify=True)
 
+    def test_resize_vdi(self, imported_vm):
+        from lib.vdi import VDI
+
+        vm = imported_vm
+        if vm.is_running():
+            vm.shutdown(verify=True, force_if_fails=True)
```
stormi, line 83: Unless I'm mistaken, `imported_vm` is always  supposed to be a stopped VM. It's a global prerequisite in our tests. So let's not attempt to shut it down.

If there currently exists situations where imported_vm can be a running VM (when using the UUID of an existing VM maybe?), then maybe the `imported_vm` fixture should detect that and fail early.

tests/storage/linstor/test_linstor_sr.py
```diff
@@ -78,6 +78,17 @@ def test_snapshot(self, vm_on_linstor_sr):
...
+    @pytest.mark.big_vm
+    def test_resize_vdi(self, vm_on_linstor_sr):
+        vm = vm_on_linstor_sr
+        if vm.is_running():
+            vm.shutdown(verify=True, force_if_fails=True)
+        logging.info("* VDI Resize started *")
+        for vdi_uuid in vm.vdi_uuids():
+            vm.vdi_resize(vdi_uuid)
+        logging.info("* VDI Resize completed *")
+
```
stormi, line 91: This looks like a test we might want to add to tests of all other SRs (in a separate PR once this one is merged?)

lib/vm.py
rushikeshjadhav: Right, will update it.

tests/misc/test_basic_without_ssh.py
rushikeshjadhav: It was also an assumed prerequisite that VM to be turned off before calling vdi resize, hence it was added. However since we will handle it within the vdi resize function and in future may offer online vdi resize, I will drop this code.

tests/storage/linstor/test_linstor_sr.py
rushikeshjadhav: Yes, this one is generally applicable to all storage tests that we have.

stormi (approved):
LGTM but I already gave a quick look to the latest changes. Trusting Ronan or Damien with their complementary review.

tests/misc/test_basic_without_ssh.py
rushikeshjadhav: The `class TestBasicNoSSH` uses a `@pytest.mark.usefixtures("started_vm")` hence `imported_vm` is always running in these tests. There is no VM online VDI resize support for now, hence this code will have to stay. @stormi is that ok?

stormi: Yes, makes sense. However I'm not sure this test belongs to `test_basic_without_ssh` at all. It's not so basic, and it's dependent on the underlying SR. To me it belongs to every SR type test instead (EXT, LVM, NFS, etc.), so that it's tested with each SR type, in addition to linstor.

rushikeshjadhav: I have removed it from `test_basic_without_ssh` and will add to SR's later.

lib/vm.py
```diff
@@ -361,6 +365,24 @@ def test_snapshot_on_running_vm(self):
         finally:
             snapshot.destroy(verify=True)
 
+    def vdi_resize(self, vdi_uuid, new_size=None):
+        # Reference
+        # MAX_VHD_SIZE = 2088960 * 1024 * 1024 # 2088960 MB
+        # MIN_VHD_SIZE = 1 * 1024 * 1024 # 1MB
+
+        # Following assert may be removed in future if we support online vdi resize
+        assert not self.is_running(), f"Expected halted, got running"
```
ydirson, line 374: There are more states than just "running" and halted, that would probably need `assert self.is_halted()` here

lib/vm.py
```diff
@@ -361,6 +365,24 @@ def test_snapshot_on_running_vm(self):
...
+
+        vdi_obj = VDI(vdi_uuid, sr=self.get_sr())
```
ydirson, line 376: just `vdi` would be suitable, no need for `_obj`

lib/vm.py
```diff
@@ -361,6 +365,24 @@ def test_snapshot_on_running_vm(self):
...
+        if new_size is None:
+            # Resize by double, ideally it should be within min and max limits reference
+            vdi_size = int(vdi_obj.param_get("virtual-size"))
+            new_size = str(vdi_size * 2)
```
ydirson, line 380: It would be cleaner to use integers to manipulate size objects.  `Host.xe()` already handles the conversion to string anyway.
To make such things more visible, it would be great to add type hints to new methods, eg:
```python
def vdi_resize(self, vdi_uuid: str, new_size: Optional[int] = None) -> None:
    ...
```

lib/vm.py
```diff
@@ -361,6 +365,24 @@ def test_snapshot_on_running_vm(self):
...
+
+        vdi_obj.resize(new_size)
+        resized_vdi_size = vdi_obj.param_get("virtual-size")
```
ydirson, line 383: same idea, better convert to `int` directly here

tests/storage/linstor/test_linstor_sr.py
```diff
@@ -78,6 +78,15 @@ def test_snapshot(self, vm_on_linstor_sr):
         finally:
             vm.shutdown(verify=True)
 
+    @pytest.mark.small_vm
+    @pytest.mark.big_vm
+    def test_resize_vdi(self, vm_on_linstor_sr):
+        vm = vm_on_linstor_sr
+        logging.info("* VDI Resize started *")
```
ydirson, line 85: Do we really want `*` marker in the logs?

ydirson (changes requested):
Don't forget an empty line after the commit summary.  For some reason GitHub makes a wrong interpretation here, see how e.g. `git shortlog` interpret this *not* a paragraph break

Also, commit summary for last commit is wrong.

tests/storage/linstor/conftest.py
```diff
@@ -12,57 +12,85 @@
...
+def pool_with_linstor(hostA2, lvm_disks, pool_with_saved_yum_state):
+    import concurrent.futures
     pool = pool_with_saved_yum_state
-    for host in pool.hosts:
+
+    def is_linstor_installed(host):
         if host.is_package_installed(LINSTOR_PACKAGE):
             raise Exception(
                 f'{LINSTOR_PACKAGE} is already installed on host {host}. This should not be the case.'
             )
```
Millefeuille42, line 62: I'd expect a boolean return value with this name, not a raised exception. you could rename it `assert_linstor_not_installed`.

Millefeuille42 (changes requested):

lib/vm.py
rushikeshjadhav: Done

lib/vm.py
rushikeshjadhav: done

lib/vm.py
rushikeshjadhav: done

tests/storage/linstor/test_linstor_sr.py
rushikeshjadhav: Removed it.

tests/storage/linstor/conftest.py
```diff
@@ -12,57 +12,85 @@
...
+                    host.ssh(['pvcreate', '-ff', '-y', device])
+                elif e.stdout.endswith('excluded by a filter.'):
+                    host.ssh(['wipefs', '-a', device])
+                    host.ssh(['pvcreate', '-ff', '-y', device])
+                else:
+                    raise e
 
-        host.ssh(['vgcreate', GROUP_NAME, device])
-        host.ssh(['lvcreate', '-l', '100%FREE', '-T', STORAGE_POOL_NAME])
+        device_list = " ".join(devices)
```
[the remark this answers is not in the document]
rushikeshjadhav: Removed it

tests/storage/linstor/conftest.py
rushikeshjadhav: Done

Millefeuille42 (approved):

glehmann (approved):

stormi:
> > @Lankou66 I've rebased and updated the branch. This is old PR so it's still linked via my account instead of a direct branch on `xcp-ng-tests`. Is it ok or need to open a new PR with a dedicated branch for running tests? cc @gduperrey
> 
> @rushikeshjadhav Please push your branch on the main repository—it should be enough to trigger the CI

It's not possible, as far as I know, to retarget the PR to another branch. And this PR is 1 year old, from a time when we had not told Rishi to push to this repository yet. Are you suggesting to create another PR, or just to push the branch and go see the CI results there without touching this PR?

rushikeshjadhav:
> > Are you suggesting to create another PR, or just to push the branch and go see the CI results there without touching this PR?
> 
> Just pushing the branch on this repo would trigger the CI, and the result would appear in this PR, as long as this PR and the branch in this repo points to the same commit

I have pushed the branch to this repo - https://github.com/xcp-ng/xcp-ng-tests/tree/feat-storage-linstor-684

PR #277, opened 2025-02-21
Add Windows Secure Boot key upgrade tests + update tests for new varstored

dinhngtu:
As we're including more Secure Boot keys by default in varstored, add tests for upgrading guest keys from the old defaults to the new ones, using certificates provided by [microsoft/secureboot_objects](https://github.com/microsoft/secureboot_objects). Additionally, disable tests that aren't applicable any more with the variable changes.

Since the new SB variables each include multiple certificates, refactor the Certificate and EFIAuth classes to support them.

Finally, add some tweaks for running tests with VM UUIDs.


conftest.py
```diff
@@ -361,7 +361,7 @@ def imported_vm(host, vm_ref):
     else:
         vm_orig = host.import_vm(vm_ref, host.main_sr_uuid(), use_cache=CACHE_IMPORTED_VM)
 
-    if CACHE_IMPORTED_VM:
+    if CACHE_IMPORTED_VM or is_uuid(vm_ref):
```
ydirson, line 364: In what is it linked to SB tests?
I'm not sure this is a behaviour we want.  It will likely break users' expectations, and such changes should come with a rationale of why we'd want to do that.
Couldn't this usecase be fulfilled by the "import_vm: add clone:// and clone+start:// URIs" commit in the `install` branch?

lib/efi.py
```diff
@@ -20,6 +22,31 @@
...
+
+    def getfile(self, suffix=None, prefix=None):
+        fd, path = mkstemp(suffix=suffix, prefix=prefix, dir=self.get().name)
+        os.close(fd)
+        return path
+
+
+_tempdir = _EfiGlobalTempdir()
+
+
```
ydirson, line 50: What does this bring that a fixture could not achieve?

lib/efi.py
```diff
@@ -128,8 +155,11 @@ def get_secure_boot_guid(variable: str) -> GUID:
 def cert_to_efi_sig_list(cert):
     """Return an ESL from a PEM cert."""
     with open(cert, 'rb') as f:
-        pem = f.read()
-        cert = x509.load_pem_x509_certificate(pem)
+        cert_raw = f.read()
+        try:
+            cert = x509.load_pem_x509_certificate(cert_raw)
+        except ValueError:
+            cert = x509.load_der_x509_certificate(cert_raw)
```
ydirson, line 191: I'm not too much familiar with PEM vs DER, and don't see at first glance what this changes does and why.
Wouldn't that be useful to have in a separate commit, where the rationale for this change would be more easily written?

ydirson (commented):
I have very mixed feelings about this large "Rework Certificate and EFIAuth" commit.  There seem to be many things inside (type hints additions, API changes, API extension....).  This is already a part of the tests (and a feature set) I suspect few of us are familiar with.  We surely want to improve on the situation.  As it is, I cannot really give an opinion on this and that disturbs me.

conftest.py
dinhngtu: It's mostly for convenience. If the `install` branch already has this feature, I will remove it.

lib/efi.py
dinhngtu: I'll add a comment for this.

lib/efi.py
dinhngtu: `efi.py` is a standalone library that doesn't integrate with pytest at all, so fixtures can't be used here.

dinhngtu:
> I have very mixed feelings about this large "Rework Certificate and EFIAuth" commit. There seem to be many things inside (type hints additions, API changes, API extension....). This is already a part of the tests (and a feature set) I suspect few of us are familiar with. We surely want to improve on the situation. As it is, I cannot really give an opinion on this and that disturbs me.

The main goal of the `efi.py` changes is to support multi-cert variables, where one is an owner cert (i.e. contains a private key) and the others (e.g. MS-distributed ones) are not.

dinhngtu:
I've tried to break up the refactor into several commits.

lib/vm.py
```diff
@@ -524,12 +524,14 @@ def install_uefi_certs(self, auths):
...
-            ])
-            self.host.ssh(['rm', '-f', dest])
+            try:
+                self.host.scp(auth.auth(), dest)
+                self.host.ssh([
+                    'varstore-set', self.uuid, auth.guid.as_str(), auth.name,
+                    str(efi.EFI_AT_ATTRS), dest
+                ])
+            finally:
+                self.host.ssh(['rm', '-f', dest])
```
stormi, line 534: The addition of `try... finally` could be extracted to a separate commit, as it doesn't match the current commit description?

lib/efi.py
```diff
@@ -309,7 +309,15 @@ def __init__(self, name, is_null=False):
...
-        return os.path.exists(self.auth)
+        return self._auth_data is not None
+
+    def auth_data(self):
+        assert self.is_signed()
+        return self._auth_data
+
+    def auth(self):
+        assert self.is_signed()
+        return self._auth
```
stormi, line 434: I can't comment on the commit message, so I'm commenting here. 

The commit message says "For checking reasons, we don't want users to grab these variables
directly from the object."

I was not sure initially what this meant. After reading the code I assume this is so that we can run `assert self.is_signed()` every time we access them. Could you enrich the commit message with a description of the problem it solves?

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "contrib/secureboot_objects"]
+	path = contrib/secureboot_objects
+	url = https://github.com/microsoft/secureboot_objects.git
```
stormi, line 3: I was under the impression that git submodules were often a pain to initialize and maintain up to date. Also this adds a dependency towards a repository that we do not control. This wouldn't be the first, but maybe we could avoid it here.

Maybe talk with @glehmann about his other PR that adds the use of `mise` to setup the test environment. I think we could decorrelate the setup from the tests, and make it so `_SecureBootCertList._prefix` can be overridden by a setting in `data.py`?

tests/uefi_sb/test_varstored_sb.py
```diff
@@ -153,3 +156,102 @@ def test_key_exchanges(self, uefi_vm):
...
+        vm.execute_powershell_script("Enable-BitLocker $Env:SystemDrive -TpmProtector -UsedSpaceOnly")
+        # Confirm if PCR7 is bound.
+        assert vm.execute_powershell_script(
+            r"""Get-CimInstance -Namespace Root\CIMV2\Security\MicrosoftVolumeEncryption `
+-Query "select * from Win32_EncryptableVolume where VolumeType=0" |
+Invoke-CimMethod -MethodName GetSecureBootBindingState |
+Where-Object ReturnValue -eq 0 |
+Select-Object -ExpandProperty BindingState"""
+        ) == "3"  # Bound
+        vm.execute_powershell_script("Suspend-BitLocker $Env:SystemDrive")
```
stormi, line 249: What happens if we don't suspend bitlocker first? Would a test where we don't suspend it be worth it?

.gitmodules
dinhngtu: I'll vendor the certs directly inside the test repo. I'd rather not add another indirection and test config for this since it basically only changes once every 10 years.

tests/uefi_sb/test_varstored_sb.py
```diff
@@ -153,3 +156,102 @@ def test_key_exchanges(self, uefi_vm):
...
+        assert not vm.get_vtpm_uuid()
+        vm.create_vtpm()
+
+        PK, _, _, _ = self.install_old_certs(vm)
+        boot_and_check_sb_succeeded(vm)
+
+        vm.shutdown(verify=True)
+
+        self.install_new_certs(vm, PK)
+        boot_and_check_sb_succeeded(vm)
```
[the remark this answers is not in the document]
dinhngtu: That'd be a silent varstored failure, which IMO should be tested elsewhere rather than specifically here.

tests/uefi_sb/test_varstored_sb.py
dinhngtu: The BitLocker theatrics here are really just for testing Windows's PCR7 binding logic against PK changes. The suspend is required to unbind PCR7 and prepare for `install_new_certs`. Without it the test would simply fail every time.

dinhngtu:
Reformatted with `ruff`.

PR #278, opened 2025-02-27
Fix fixture vm_with_vbds not deleting VBDs

Nambrok:
It didn't delete the VBD so `vm.destroy()` would destroy the VDI that should have been cleanup by the fixture `vdis` instead.
The fixture `vdis` would then raise an exception when trying to delete inexisting VDIs.

### DIFF
--- a/tests/snapshot/conftest.py
+++ b/tests/snapshot/conftest.py
@@ -21,16 +21,20 @@ def vdis(host, local_sr_on_hostA1):
 
 @pytest.fixture(scope='module')
 def vm_with_vbds(host, vdis, imported_vm):
-    def _make_vbd(host, device, vdi):
-        host.xe('vbd-create', {
-            'vm-uuid': vm.uuid, 'mode': 'RW', 'type': 'Disk', 'device': device, 'vdi-uuid': vdi
-        })
-
+    vbds = []
     vm = imported_vm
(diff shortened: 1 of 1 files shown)


gduperrey (approved):
8.2 main :white_check_mark:

tests/snapshot/conftest.py
```diff
@@ -22,15 +22,24 @@ def _make_vdi(name):
 @pytest.fixture(scope='module')
 def vm_with_vbds(host, vdis, imported_vm):
     def _make_vbd(host, device, vdi):
-        host.xe('vbd-create', {
+        return host.xe('vbd-create', {
             'vm-uuid': vm.uuid, 'mode': 'RW', 'type': 'Disk', 'device': device, 'vdi-uuid': vdi
         })
 
+    def _remove_vbd(host, vbd_uuid):
```
Wescoeur, line 29: I don't understand why we use local functions with a "_" prefix.
Also, why do we use local functions that are called only once instead of directly executing block code?

Wescoeur (changes requested):

gduperrey (approved):
* 8.2 main: :white_check_mark:
* 8.3 main: :white_check_mark:

Wescoeur (approved):

PR #281, opened 2025-03-10
Rewrite generic storage tests to deduplicate

Nambrok:
Remove `tests/storage/{lvm,ext}` that were duplicated tests.
Add `tests/storage/generic` to replace them.
Add a parameter `--sr-type` to select which SR types to run generic tests on.
If the parameter is set to `all`, it will run on all members of the list defined in `tests/storage/generic/conftest.py:SR_TYPES` (containing lvm and ext SR types for the moment).
The jobs for ext and lvm only need to add the parameter in their job definition (I will add this soon).


conftest.py
```diff
@@ -75,6 +75,13 @@ def pytest_addoption(parser):
              "4KiB blocksize to be formatted and used in storage tests. "
              "Set it to 'auto' to let the fixtures auto-detect available disks."
     )
+    parser.addoption(
+        "--sr-type",
```
stormi, line 79: This is another added parameter to control which tests should be executed or not, which adds a layer between test definitions themselves and job definitions (jobs.py). I see a lot of PRs adding parameters, and this looks like added complexity for the ones running the tests.

How do you express "all storage types except largeblock and linstor" with this? (see job definitions in jobs.py that rely on this)

Nambrok: This only include ext and lvm since those were identical. I agree it's not optimal, I'm looking into a way that will involve giving a path to pytest.

stormi: Couldn't we use test class inheritance and still keep individual test files?

Nambrok: I'm gonna try new things for this PR, gonna see if we can do something better with inheritance or other mechanisms.

PR #285, opened 2025-03-13
wip: air gapped update

glehmann:


tests/air_gapped/test_air_gapped_update.py
```diff
@@ -0,0 +1,44 @@
...
+    tmppath = f'/tmp/{os.path.basename(url)}'
+    urlretrieve(url, tmppath)
+    host.scp(tmppath, tmppath)
+    os.remove(tmppath)
+
+
+def test_air_gapped_update(host: Host):
+    # get the required files and install the offline repositories
+    air_gapped_download(
+        'https://raw.githubusercontent.com/xcp-ng/xcp/refs/heads/master/scripts/setup_offline_xcpng_repos', host
```
rzr: I would create a variable or env var the can override the location, using floating master as default
but I believe a pinned version would make the test more reproductible

PR #286, opened 2025-03-14
Fixes mypy support

ydirson:
This PR:
* makes it possible to run `mypy` 1.15.0 on `lib/` and `tests/`, so we have a proper base for checking new type hints and catch regressions
* fixes hints of the EFI tests, I hope I got everything right (Bobby isn't with us any more to check that part of the code)

This leaves `mypy` only reprting errors only on `tests/guest_tools/win/`, I'll let @dinhngtu deal with those.


dinhngtu:
For the Windows guest tests please apply this patch: 

```patch
From 64564e1c9c8680ec6fd3c7966cfa92fc5571628a Mon Sep 17 00:00:00 2001
From: Tu Dinh <ngoc-tu.dinh@vates.tech>
Date: Fri, 14 Mar 2025 14:31:04 +0100
Subject: [PATCH] guest_tools/win: Fix type hints for test_xenclean

Signed-off-by: Tu Dinh <ngoc-tu.dinh@vates.tech>
---
 tests/guest_tools/win/test_xenclean.py | 4 ++--
 1 file changed, 2 insertions(+), 2 deletions(-)

diff --git a/tests/guest_tools/win/test_xenclean.py b/tests/guest_tools/win/test_xenclean.py
index 94719c8..a025e4e 100644
--- a/tests/guest_tools/win/test_xenclean.py
+++ b/tests/guest_tools/win/test_xenclean.py
@@ -1,6 +1,6 @@
 import logging
 from pathlib import PureWindowsPath
-from typing import Any, Dict
+from typing import Any, Dict, Tuple
 import pytest
 from lib.common import wait_for
 from lib.vm import VM
@@ -48,7 +48,7 @@ class TestXenClean:
         run_xenclean(vm, guest_tools_iso)
         assert vm.are_windows_tools_uninstalled()
 
-    def test_xenclean_with_other_tools(self, vm_install_other_drivers: VM, guest_tools_iso):
+    def test_xenclean_with_other_tools(self, vm_install_other_drivers: Tuple[VM, Dict], guest_tools_iso):
         vm, param = vm_install_other_drivers
         if param.get("vendor_device"):
             pytest.skip("Skipping XenClean with vendor device present")
-- 
2.43.0
```

dinhngtu (approved):

dinhngtu:
Could you also add a mypy job in the automated workflows?

ydirson:
> Could you also add a mypy job in the automated workflows?

That's the goal :)

glehmann (approved):

ydirson:
mypy job now active

glehmann (approved):

stormi (approved):

lib/efi.py
```diff
@@ -122,7 +122,7 @@ def get_secure_boot_guid(variable: str) -> GUID:
         'KEK': global_variable_guid,
         'db': image_security_database_guid,
         'dbx': image_security_database_guid,
-    }.get(variable)
+    }[variable]
 
 
 def cert_to_efi_sig_list(cert):
```
gthvn1, line 128: shouldn't we also annotate the return value here? As it is called by `certs_to_sig_db` that has a return value I expected to also see a return value here.

gthvn1 (approved):
Cool to see mypy in the flow :+1: 
I just have a remark about when function is annotated and when it is not. Otherwise it looks good to me.

lib/efi.py
gthvn1: Ok in fact I just understood that you fix existing one. Forget my remark :)

PR #288, opened 2025-03-26
add uv configuration

glehmann:
Then we just add `uv run` before the command we usually run, and it installs/updates the needed dependencies

~~~sh
❯ uv run pytest tests/misc/test_basic_without_ssh.py --hosts=gln-xcpng 
Using CPython 3.13.2 interpreter at: /usr/bin/python3.13
Creating virtual environment at: .venv
Installed 19 packages in 1.06s
…
~~~

It even installs Python if the local version doesn't match the required one.


glehmann:
`mise` and `uv` used together in #289

PR #289, opened 2025-03-27
setup mise

glehmann:
Today, configuring the test environment requires several manual steps, and the same is true when updating the test dependencies.
Moreover, the developers are not all using the same Python version, which may introduce differences in the test behavior.
Finally, with some dependencies from the operating system, most notably on Python, we are forced to use the oldest version available on our test environment, and can't benefit from the newest features.

`mise` helps solve most of this problem. It takes care of installing the dependencies independently of the languages they are written in. It ensures that the dependencies are at the exact same version on all the test environments that are set up. It can install a language like Python or Node at a specific version. It is easily installable as a single binary.

install `mise` once with

~~~sh
curl https://mise.run | sh
~~~

then just run

~~~sh
mise run test tests/misc/test_basic_without_ssh.py --hosts=gln-xcpng
~~~

to install all the required dependencies, including `xo-cli`, `python`, and required python modules, and run the selected tests.

If you prefer not to use `mise run`, you can tell mise to install the dependencies and then set up the shell environment to have them accessible in the shell:

~~~sh
mise install
mise activate | source
~~~

`mise` deactivates itself as soon as you leave the project directory, and reactivates when you enter a directory with a `mise` configuration, so you may even put that last command in your global shell configuration.


stormi:
IMO this kind of change lacks an essential introduction in the commit message(s), detailing what problem it solves and why it's the good way to solve it: see https://docs.xcp-ng.org/project/development-process/commit-message-conventions/

mise.toml
```diff
@@ -0,0 +1,11 @@
+[tools]
+"npm:xo-cli" = "0.32.1"
+python = "3.13"
+uv = "0.6"
```
ydirson, line 4: is it `mise` that relies on `uv` for the multiple python interpreter support?

pyproject.toml
```diff
@@ -0,0 +1,24 @@
+[project]
+name = "xcp-ng-tests"
+version = "0.1.0"
+description = "Add your description here"
+readme = "README.md"
+requires-python = ">=3.12"
```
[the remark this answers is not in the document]
glehmann: done

mise.toml
glehmann: `mise` use the same python binaries as `uv` but it doesn't require `uv` to install python.
`uv` is there to deal with the python dependencies

ydirson (commented):
It would be useful to have this detailed usage description available in `README.md`, to be easier to access and extend in the future.

pyproject.toml
```diff
@@ -0,0 +1,26 @@
+[project]
+name = "xcp-ng-tests"
+version = "0.1.0"
+description = "Add your description here"
```
ydirson, line 4: Are we forced to set a dummy `version` here?

`description` could benefit from a small enhancement :)

uv.lock
```diff
@@ -0,0 +1,856 @@
+version = 1
+revision = 1
+requires-python = ">=3.8, <4"
+resolution-markers = [
+    "python_full_version >= '3.11'",
+    "python_full_version == '3.10.*'",
+    "python_full_version == '3.9.*'",
+    "python_full_version < '3.9'",
+]
+
```
ydirson, line 10: The usage in `README.md` should also give some guidelines (even just tentative ones we can refine in the future) as to how we should manage this `uv.lock`.

Can dependabot be setup to notify us of availability of important/security upgrades?  Maybe in a future PR?

Maybe we need a more general policy on this.

pyproject.toml
glehmann: yes, unfortunately, for the `version`

~~~
error: Failed to parse: `pyproject.toml`
  Caused by: TOML parse error at line 1, column 1
  |
1 | [project]
  | ^^^^^^^^^
`pyproject.toml` is using the `[project]` table, but the required `project.version` field is neither set nor present in the `project.dynamic` list
~~~

agreed for the description :+1:

uv.lock
glehmann: Dependabot now has support to update `uv.lock`. I didn't have a chance to try it, though.

README.md
```diff
@@ -3,25 +3,36 @@
...
 
-```
-$ pip install -r requirements/base.txt
-```
+~~~sh
+curl https://mise.run | sh
+~~~
 
-Additionally, for dev dependencies (things like the linter / style checker):
+Optionally install the mise managed dependencies with `mise install`.
```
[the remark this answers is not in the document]
glehmann: The installation problem is fixed (or rather workaround, as it looks like a bug in mise).

I disagree about `xo-cli`: the goal here is to have all the required dependencies to run the tests, not to have to think about installing it when a test fails because' xo-cli' is not there.

.python-version
```diff
@@ -0,0 +1 @@
+3.8
```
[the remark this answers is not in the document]
glehmann: fixed :)

README.md
```diff
@@ -3,25 +3,36 @@
...
+This project uses [mise](https://github.com/jdx/mise) to deal with the dependencies.
+
+Extra test-specific requirements are documented in the test file "Requirements" header.
   * `netcat` is required on the machine running the tests.
 
-### Quick install (python requirements)
+### Quick install
 
-Install the python requirements using pip:
+Install `mise` with
```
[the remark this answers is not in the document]
glehmann: You don't have to explicitly run `mise trust`, but you have to trust the repository the first time you run a mise command.

README.md
```diff
@@ -3,25 +3,36 @@
...
-```
+~~~sh
+curl https://mise.run | sh
+~~~
 
-Additionally, for dev dependencies (things like the linter / style checker):
+Optionally install the mise managed dependencies with `mise install`.
+
+You can then use the dependencies managed by mise by prefixing running the commands documented in the next chapters with `mise run`.
+For example, `mise run pytest`, `mise run ./jobs.py`, …
```
ydirson, line 23: Points I noted that can be worth pointing out to help people chose wisely how to run `mise`:
- `mise run pytest` prevents the use of `pytest` shell completion - or maybe as part of its own shell completion
- `mise` does not provide shell completion out of the box, info in https://mise.jdx.dev/cli/completion.html.  Note I was unable to set that up:
  - installing in `~/.local/share/bash-completion/completions/` seems to be ignored by `bash-completion`, despite the path appearing in the latter
  - it does load when installed as `/etc/bash_completion.d/mise` but then
    ```
    Error: usage CLI not found. This is required for completions to work in mise.
    See https://usage.jdx.dev for more information.
    ```
    Real info is in https://usage.jdx.dev/cli/ (`cargo install usage-cli` - yes, "usage" was not just an english word in the error message :scream:)
  - even then it likely requires to run on a nightly build of a rolling release, thanks :crab: guys
    ```
    mise bash: _comp_initialize: command not found
    ```
  - ah, solution at https://github.com/jdx/mise/discussions/4744
    well, sort of, the thing crashes, you have to run it 2-3 times so it runs without crashing
    and then when attempting to complete
```
    $ mise mise ERROR [~/src/xcpng/tests/mise.toml] npm:xo-cli@0.32.2: No such file or directory (os error 2)
mise ERROR No such file or directory (os error 2)
mise ERROR Run with --verbose or MISE_VERBOSE=1 for more information
Error: xx::process (link)

  × exited with code 1
  │ sh -c mise tasks ls --complete
```

README.md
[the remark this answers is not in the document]
stormi: Also, I thought `mise` would be an optional helper, but the way the README is currently updated makes it look like it's the main way to use xcp-ng-tests and other ways are exceptions.

glehmann:
We can't activate the virtual environment for now with mise.
See https://github.com/jdx/mise/discussions/5301

We can try again mise later if we want xo-cli installed with the other dependencies

PR #290, opened 2025-04-02
feat(coalesce): Add tests for coalesce

Nambrok:
The tests create a VDI, connect it to Dom0 then use the tapdev to write random data somewhere in it.
It then compare the data of the VDI to the original data we have to see if it has changed.
The test create snapshot/clone before deleting them and waiting for the coalesce to have happened by observing `sm-config:vhd-parent` before checking the integrity of the data.

Also add `--vdi-type` parameter that default to `vhd` to prepare to tests multiple VDI types, i.e. qcow2.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -94,6 +94,13 @@ def pytest_addoption(parser):
              "4KiB blocksize to be formatted and used in storage tests. "
              "Set it to 'auto' to let the fixtures auto-detect available disks."
     )
+    parser.addoption(
+        "--image-format",
+        action="append",
+        default=[],
+        help="Format of VDI to execute tests on."
+        "Example: vhd,qcow2"
+    )
 
 def pytest_configure(config):
     global_config.ignore_ssh_banner = config.getoption('--ignore-ssh-banner')
@@ -105,6 +112,12 @@ def pytest_generate_tests(metafunc):
         if not vms:
             vms = [None] # no --vm parameter does not mean skip the test, for us, it means use the default
         metafunc.parametrize("vm_ref", vms, indirect=True, scope="module")
+
+    if "image_format" in metafunc.fixturenames:
+        image_format = metafunc.config.getoption("image_format")
+        if len(image_format) == 0:
+            image_format = ["vhd"] # Not giving image-format will default to doing tests on vhd
+        metafunc.parametrize("image_format", image_format, scope="session")
 
 def pytest_collection_modifyitems(items, config):
     # Automatically mark tests based on fixtures they require.
--- a/conftest.py
+++ b/conftest.py
@@ -304,6 +304,13 @@ def host_no_ipv6(host):
     if is_ipv6(host.hostname_or_ip):
         pytest.skip(f"This test requires an IPv4 XCP-ng")
 
+@pytest.fixture(scope="session")
+def shared_sr(host):
+    sr = host.pool.first_shared_sr()
+    assert sr, "No shared SR available on hosts"
+    logging.info(">> Shared SR on host present: {} of type {}".format(sr.uuid, sr.get_type()))
+    yield sr
+
 @pytest.fixture(scope='session')
 def local_sr_on_hostA1(hostA1):
     """ A local SR on the pool's master. """
@@ -311,7 +318,7 @@ def local_sr_on_hostA1(hostA1):
     assert len(srs) > 0, "a local SR is required on the pool's master"
     # use the first local SR found
     sr = srs[0]
-    logging.info(">> local SR on hostA1 present : %s" % sr.uuid)
+    logging.info(">> local SR on hostA1 present: {} of type {}".format(sr.uuid, sr.get_type()))
     yield sr
 
 @pytest.fixture(scope='session')
@@ -321,7 +328,7 @@ def local_sr_on_hostA2(hostA2):
     assert len(srs) > 0, "a local SR is required on the pool's second host"
     # use the first local SR found
     sr = srs[0]
-    logging.info(">> local SR on hostA2 present : %s" % sr.uuid)
+    logging.info(">> local SR on hostA2 present: {} of type {}".format(sr.uuid, sr.get_type()))
     yield sr
 
 @pytest.fixture(scope='session')
@@ -331,7 +338,7 @@ def local_sr_on_hostB1(hostB1):
     assert len(srs) > 0, "a local SR is required on the second pool's master"
     # use the first local SR found
     sr = srs[0]
-    logging.info(">> local SR on hostB1 present : %s" % sr.uuid)
+    logging.info(">> local SR on hostB1 present: {} of type {}".format(sr.uuid, sr.get_type()))
     yield sr
 
 @pytest.fixture(scope='session')
--- a/lib/basevm.py
+++ b/lib/basevm.py
@@ -1,6 +1,6 @@
 import logging
 
-from typing import TYPE_CHECKING, Any, Literal, Optional, overload
+from typing import TYPE_CHECKING, Any, List, Literal, Optional, overload
 
 if TYPE_CHECKING:
     import lib.host
@@ -59,7 +59,7 @@ class BaseVM:
     def _disk_list(self):
         raise NotImplementedError()
 
-    def vdi_uuids(self, sr_uuid=None):
+    def vdi_uuids(self, sr_uuid: Optional[str] = None) -> List[str]:
         output = self._disk_list()
         if output == '':
             return []
--- a/lib/host.py
+++ b/lib/host.py
@@ -67,6 +67,7 @@ class Host:
         self.uuid = self.inventory['INSTALLATION_UUID']
         self.xcp_version = version.parse(self.inventory['PRODUCT_VERSION'])
         self.xcp_version_short = f"{self.xcp_version.major}.{self.xcp_version.minor}"
+        self._dom0: Optional[VM] = None
 
     def __str__(self):
         return self.hostname_or_ip
@@ -711,3 +712,20 @@ class Host:
     def disable_hsts_header(self):
         self.ssh(['rm', '-f', f'{XAPI_CONF_DIR}/00-XCP-ng-tests-enable-hsts-header.conf'])
         self.restart_toolstack(verify=True)
+
+    def get_dom0_uuid(self):
+        return self.inventory["CONTROL_DOMAIN_UUID"]
+
+    def get_dom0_vm(self) -> VM:
+        if not self._dom0:
+            self._dom0 = VM(self.get_dom0_uuid(), self)
+        return self._dom0
+
+    def get_sr_from_vdi_uuid(self, vdi_uuid: str) -> Optional[SR]:
+        sr_uuid = self.xe("vdi-param-get", {
+            "param-name": "sr-uuid",
+            "uuid": vdi_uuid,
+        })
+        if not sr_uuid:
+            return None
+        return SR(sr_uuid, self.pool)
--- a/lib/sr.py
+++ b/lib/sr.py
@@ -1,5 +1,6 @@
 import logging
 import time
+from typing import Optional
 
 import lib.commands as commands
 from lib.common import prefix_object_name, safe_split, strtobool, wait_for, wait_for_not
@@ -11,6 +12,7 @@ class SR:
         self.pool = pool
         self._is_shared = None # cached value for is_shared()
         self._main_host = None # cached value for main_host()
+        self._type = None # cache value for get_type()
 
     def pbd_uuids(self):
         return safe_split(self.pool.master.xe('pbd-list', {'sr-uuid': self.uuid}, minimal=True))
@@ -152,13 +154,21 @@ class SR:
                                                             {'uuid': self.uuid, 'param-name': 'shared'}))
         return self._is_shared
 
-    def create_vdi(self, name_label, virtual_size=64):
+    def get_type(self) -> str:
+        if self._type is None:
+            self._type = self.pool.master.xe("sr-param-get", {"uuid": self.uuid, "param-name": "type"})
+        return self._type
+
+    def create_vdi(self, name_label: str, virtual_size: int = 64, image_format: Optional[str] = None) -> VDI:
         logging.info("Create VDI %r on SR %s", name_label, self.uuid)
-        vdi_uuid = self.pool.master.xe('vdi-create', {
+        args = {
             'name-label': prefix_object_name(name_label),
             'virtual-size': str(virtual_size),
-            'sr-uuid': self.uuid
-        })
+            'sr-uuid': self.uuid,
+        }
+        if image_format:
+            args["sm-config:image-format"] = image_format
+        vdi_uuid = self.pool.master.xe('vdi-create', args)
         return VDI(vdi_uuid, sr=self)
 
     def run_quicktest(self):
--- a/lib/sr.py
+++ b/lib/sr.py
@@ -1,10 +1,17 @@
 import logging
 import time
-from typing import Optional
 
 import lib.commands as commands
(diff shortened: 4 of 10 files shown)


conftest.py
```diff
@@ -75,6 +75,13 @@ def pytest_addoption(parser):
              "4KiB blocksize to be formatted and used in storage tests. "
              "Set it to 'auto' to let the fixtures auto-detect available disks."
     )
+    parser.addoption(
+        "--vdi-type",
```
Wescoeur, line 79: Maybe `image-format` or `vdi-image-format`.
In fact vdi-type is ambiguous in many contextes (sm, XAPI), it can be user/system for example, so not a format.

lib/host.py
```diff
@@ -606,3 +607,16 @@ def enable_hsts_header(self):
     def disable_hsts_header(self):
         self.ssh(['rm', '-f', f'{XAPI_CONF_DIR}/00-XCP-ng-tests-enable-hsts-header.conf'])
         self.restart_toolstack(verify=True)
+
+    def get_dom0_uuid(self):
+        output = self.ssh(["grep", "-e", "\"CONTROL_DOMAIN_UUID=\"", "/etc/xensource-inventory"])
+        return output.split("=")[1].replace("'", "")
+
+    def get_vdi_sr_uuid(self, vdi_uuid) -> Optional[SR]:
```
Wescoeur, line 615: It's not really a SR UUID here, you return an object.

Wescoeur (changes requested):

conftest.py
AnthoineB: I agree, I prefer `image-format ` too.

Nambrok: Changed then :)

Nambrok:
For the failing test: waiting for https://github.com/xcp-ng/xcp-ng-tests/issues/291to be resolved.

conftest.py
```diff
@@ -75,6 +75,13 @@ def pytest_addoption(parser):
              "4KiB blocksize to be formatted and used in storage tests. "
              "Set it to 'auto' to let the fixtures auto-detect available disks."
     )
+    parser.addoption(
+        "--image-format",
+        action="append",
+        default=[],
+        help="Format of VDI to execute tests on."
+        "Example: vhd,qcow2"
```
stormi, line 83: Making this a CLI parameter means you're deporting the test job definition outside pytest. When rishi wanted to do the same for thin vs thick in the context of XOSTOR tests, we asked him to rework the tests so they actually test both formats.

Nambrok: It's to allow to test only one type by hand, the objectives for tests is to keep the default which at the moment is `vhd` but will be `["vhd", "qcow2"]` eventually.

conftest.py
```diff
@@ -75,6 +75,13 @@ def pytest_addoption(parser):
              "4KiB blocksize to be formatted and used in storage tests. "
              "Set it to 'auto' to let the fixtures auto-detect available disks."
     )
+    parser.addoption(
+        "--image-format",
```
stormi, line 79: When you work in the context of storage everyday, maybe it's a clear name, but otherwise it might be too generic a name. Both `image` and `format` can be many things.

conftest.py
stormi: pytest already has test selection mechanisms:
* path to test files
* `-k`
* `-m`

conftest.py
Wescoeur: It's why I added `vdi-image-format` as suggestion.
But I wonder if we should not use the naming `volume-image-format` instead.

No preference.

stormi: And `vdi-type` looked consistent with `sm`'s terminology :thinking:

conftest.py
stormi: As long as the parameter is not necessary in `./jobs.py`, fine by me. We won't have SR types that only support VHD or only support QCOW2 at some point?

glehmann:
> For the failing test: waiting for #291 to be resolved.

I thought it would take a bit longer before it showed again!

conftest.py
Nambrok: The parameter default to VHD when not given meaning it's not needed to change jobs.py. But I will need to add the new coalesce tests in jobs.py either way and it's the only test at the moment that will use `vdi-type`
I'm not sure to see how the other selection mechanisms from pytest could work in this case.

conftest.py
Wescoeur:
```
# xe vdi-list params=type | sort | uniq
type ( RO)    : CBT metadata
type ( RO)    : HA statefile
type ( RO)    : Redo log
type ( RO)    : User
```

Not really consistent. In this example here, type is completly different.
- On the XAPI side, we have a `vdi_type` stored in the sm-config attribute. Here you can have the VHD/AIO value. However there are weird situations like: `sm-config (MRO)    : type: raw; vdi_type: aio`.
- Regarding the new QCOW2 format we discussed internally to use a new attribute image-format. `type` is completely ambiguous depending on what you are talking about and the context in which it is used.
- In the SMAPIv3, vdi type is completly removed (same for VDI, we use `volume` instead). And `image-format` is used.

AnthoineB (approved):

conftest.py
rushikeshjadhav: Works, I think we can move ahead with this merge.

conftest.py
```diff
@@ -98,6 +105,12 @@ def pytest_generate_tests(metafunc):
             vms = [None] # no --vm parameter does not mean skip the test, for us, it means use the default
         metafunc.parametrize("vm_ref", vms, indirect=True, scope="module")
 
+    if "image_format" in metafunc.fixturenames:
+        image_format = metafunc.config.getoption("image_format")
+        if len(image_format) == 0:
+            image_format = ["vhd"] # Not giving image-format will default to doing tests on vhd
```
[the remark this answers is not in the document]
Nambrok: Indeed, but it's because different input are made into a list so `--image-format=vhd --image-format=qcow2 => ["vhd", "qcow2"]`.
Some fixture are doing both but I didn't explicitely, we could do both.

lib/host.py
```diff
@@ -711,3 +712,20 @@ def enable_hsts_header(self):
...
+    def get_dom0_VM(self) -> VM:
+        if not self._dom0:
+            self._dom0 = VM(self.get_dom0_uuid(), self)
+        return self._dom0
+
+    def get_sr_from_vdi_uuid(self, vdi_uuid) -> Optional[SR]:
+        sr_uuid = self.xe("vdi-param-get",
+                          {"param-name": "sr-uuid",
+                           "uuid": vdi_uuid,
+                           })
```
Wescoeur, line 728:
```suggestion
        sr_uuid = self.xe("vdi-param-get", {
            "param-name": "sr-uuid",
            "uuid": vdi_uuid,
        })
```

lib/host.py
```diff
@@ -711,3 +712,20 @@ def enable_hsts_header(self):
...
+        if sr_uuid is None:
```
Wescoeur, line 729:
```suggestion
        if not sr_uuid:
```

I guess we can't have an empty string here, and it's the case it's invalid, right?

Wescoeur (approved):

tests/storage/coalesce/conftest.py
```diff
@@ -0,0 +1,62 @@
...
+
+@pytest.fixture(scope="module")
+def vdi_on_local_sr(host, local_sr_on_hostA1, image_format):
+    sr_uuid = local_sr_on_hostA1.uuid
+    vdi_uuid = host.xe("vdi-create",
+                       {"sr-uuid": sr_uuid,
+                        "name-label": "testVDI",
+                        "virtual-size": str(MAX_LENGTH),
+                        "sm-config:type": image_format,
+                        })
```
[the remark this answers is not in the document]
Nambrok: It's done in a commit after

tests/storage/coalesce/conftest.py
```diff
@@ -0,0 +1,62 @@
...
+                       {"vdi-uuid": vdi_uuid,
+                        "vm-uuid": dom0_uuid,
+                        "device": "autodetect",
+                        })
+    host.xe("vbd-plug", {"uuid": vbd_uuid})
+
+    yield vdi_on_local_sr
+
+    logging.info(f"<< Unplugging VDI {vdi_uuid} from Dom0")
+    host.xe("vbd-unplug", {"uuid": vbd_uuid})
```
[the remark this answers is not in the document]
Wescoeur: I'm a little afraid of potential side effects personally, but I have nothing against a warning log. On the other hand, we don't check whether the VDI to unplug is in the `self.vdis` list. :thinking:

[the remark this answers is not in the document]
ydirson: > I'm a little afraid of potential side effects

I'm myself a bit afraid of what can happen if we proceed while the system is not in the expected state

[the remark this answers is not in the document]
Nambrok: Expected state being VBD unplugged so we can then remove it. Calling it and it doing nothing is not a problem, if it errors it will be raised.

lib/basevm.py
```diff
@@ -18,6 +20,15 @@ def __init__(self, uuid: str, host: 'lib.host.Host'):
         logging.info("New %s: %s", type(self).__name__, uuid)
         self.uuid = uuid
         self.host = host
+        try:
+            self.vdis = [VDI(vdi_uuid, host=host) for vdi_uuid in self.vdi_uuids()]
+        except SSHCommandFailed as e:
+            # Doesn't work with Dom0 since `vm-disk-list` doesn't work on it so we create empty list
+            if e.stdout == "Error: No matching VMs found":
+                logging.info("Couldn't get disks list. We are Dom0. Continuing...")
```
ydirson, line 28: Maybe rather something like `"Couldn't get disks list (maybe we are dom0?). Continuing..."` ?

lib/basevm.py
```diff
@@ -59,7 +70,7 @@ def name(self) -> str:
     def _disk_list(self):
         raise NotImplementedError()
 
-    def vdi_uuids(self, sr_uuid=None):
+    def vdi_uuids(self, sr_uuid=None) -> List[str]:
```
ydirson, line 73: could also type `sr_uuid`

lib/vm.py
```diff
@@ -287,6 +288,30 @@ def checkpoint(self) -> Snapshot:
...
+        vbd_uuid = self.host.xe("vbd-create", {
+            "vdi-uuid": vdi.uuid,
+            "vm-uuid": self.uuid,
+            "device": device,
+        })
+        self.host.xe("vbd-plug", {"uuid": vbd_uuid})
+
+        self.vdis.append(vdi)
+
+        return vbd_uuid
```
ydirson, line 302: if `vbd-plug` fails we may want to remove the vbd?

It looks like 2 fixtures for create and plug would be a good fit, except that it would not allow to store `self.vdis`, but I don't see why we need to store it (note: this advocates for finer commit granularity, where we have a better opportunity to explain why we're adding such a feature)... and creating a VDI object requires at least on XAPI call, which can make this sensibly slower especially for a VM with many VDIs

tests/storage/coalesce/conftest.py
```diff
@@ -0,0 +1,53 @@
...
+    host.ssh(["rm", filename])
+
+@pytest.fixture(scope="module")
+def tapdev(local_sr_on_hostA1, vdi_with_vbd_on_dom0):
+    """
+    A tapdev is a blockdevice allowing access to a VDI from the Dom0.
+
+    It is usually used to give access to the VDI to Qemu for emulating devices
+    before PV driver are loaded in the guest.
+    """
```
ydirson, line 50: That does not really explain how it is useful for testing.

Maybe naming it like `tapdev_on_vdi_on_local_sr` would be sufficient.

tests/storage/coalesce/test_coalesce.py
```diff
@@ -0,0 +1,42 @@
...
+        offset = 0
+
+        logging.info("Copying data to tapdev")
+        copy_data_to_tapdev(host, data_file_on_host, tapdev, offset, length)
+
+        logging.info("Comparing data to tapdev")
+        assert compare_data(host, tapdev, data_file_on_host, offset, length)
+
+    @pytest.mark.parametrize("vdi_op", ["snapshot", "clone"])
+    def test_coalesce(self, host: "Host", tapdev: str, vdi_with_vbd_on_dom0: "VDI", data_file_on_host: str, vdi_op):
```
ydirson, line 25: Having to use `vdi_with_vbd_on_dom0` to get access to the VDI behind `tapdev` is confusing.
Maybe a `TapDev` object to encapsulate all relevant info would improve this?

ydirson (changes requested):

stormi (approved):
Almost good to merge for me, as we need to move on and I don't think remaining unadressed comments add too much technical debt.

However, I'd prefer if, before the merge, some of the remaining comments were addressed.

And this has to be run on CI before the merge.

tests/storage/coalesce/conftest.py
glehmann: It should be mentioned somewhere that tapdev is used for a better test performance, to avoid instantiating a new VM to access the vdi

tests/storage/coalesce/test_coalesce.py
glehmann, line 25: We could avoid generating a file on the dom0 if we:
* generate a file with random content on the vdi directly;
* compute the check sum and retrieve it locally
* trigger the coalesce
* recompute the check sum and verify it's the same as before

lib/sr.py
```diff
@@ -152,13 +154,21 @@ def is_shared(self):
                                                             {'uuid': self.uuid, 'param-name': 'shared'}))
         return self._is_shared
 
-    def create_vdi(self, name_label, virtual_size=64):
+    def get_type(self) -> str:
+        if self._type is None:
```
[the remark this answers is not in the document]
ydirson: There is `@functools.cache` in this case :)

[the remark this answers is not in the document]
Wescoeur: IMHO I think having own decorator to handle 0 arg (const_cache?) or continuing with an explicit private variable is better. `@functools.cache` uses a dict in its implementation to deal with arguments. (I think I don't like scripted languages with a tendency to consume more resources where we wouldn't do that in a binary :p )

lib/vm.py
```diff
@@ -338,6 +331,18 @@ def destroy_vdi(self, vdi_uuid: str) -> None:
                 super().destroy_vdi(vdi_uuid)
                 break
 
+    def create_vdis_list(self) -> None:
```
ydirson, line 334: `_create_vdis_list` would make it more obvious this is an internal method

lib/vm.py
```diff
@@ -34,15 +34,7 @@ def __init__(self, uuid: str, host: 'Host'):
...
-        try:
-            self.vdis = [VDI(vdi_uuid, host=host) for vdi_uuid in self.vdi_uuids()]
-        except commands.SSHCommandFailed as e:
-            # Doesn't work with Dom0 since `vm-disk-list` doesn't work on it so we create empty list
-            if e.stdout == "Error: No matching VMs found":
-                logging.info("Couldn't get disks list. We are Dom0. Continuing...")
-                self.vdis = []
-            else:
-                raise
+        self.create_vdis_list()
```
ydirson: Since this is a fix of the previous commit, with the essential change being moving this new code into a separate method, it would be worth squashing?

Now maybe for clarity that previous commit could be split, but it may be small enough that we prefer merging as is

lib/vm.py
ydirson: It's still not for users of the object, just like `_disk_list`.  One leading underscore is the conventional equivalent to C++' `protected`, there two of them are the `private` equivalent.

Issue #291, opened 2025-04-02
Drop the DWIM behavior on bools is in `xe()`


ydirson:
The problem is the DWIM behavior on bools is in `xe()`, which clearly seems a bad place for this.  Do we even have a clue if it is only `get_param` using this?

> I think I like the cast() version more, though.

You only had to add a single one because of how few type hints we have today.  When the majority of the calls would have to get an explicit cast, IMO it can only be a wrong solution.

And anyway, how are things like this behaving with this hidden cast to bool?

```
lib/vdi.py:        return self.param_get("read-only") == "true"
```

This implicit cast only created a mess.

_Originally posted by @ydirson in https://github.com/xcp-ng/xcp-ng-tests/pull/284#discussion_r2018197924_

PR #292, opened 2025-04-03
Adapt return types of xe and param_get

Nambrok:
Simplify the return type of `xe`, remove bool being a implicit cast for `xe`.
Adapt some call that were expecting a bool to use `strtobool` explicitely instead.
Adapted `_param_get` to new types for xe.

Resolve https://github.com/xcp-ng/xcp-ng-tests/issues/291

### DIFF
--- a/lib/basevm.py
+++ b/lib/basevm.py
@@ -1,6 +1,6 @@
 import logging
 
-from typing import Any, Optional, TYPE_CHECKING, Union
+from typing import Any, Literal, Optional, overload, TYPE_CHECKING
 
 import lib.commands as commands
 if TYPE_CHECKING:
@@ -13,14 +13,25 @@ class BaseVM:
     """ Base class for VM and Snapshot. """
 
     xe_prefix = "vm"
+    uuid: str
 
-    def __init__(self, uuid, host: 'lib.host.Host'):
+    def __init__(self, uuid: str, host: 'lib.host.Host'):
         logging.info("New %s: %s", type(self).__name__, uuid)
         self.uuid = uuid
         self.host = host
 
+    @overload
+    def param_get(self, param_name: str, key: Optional[str] = ...,
+                  accept_unknown_key: Literal[False] = ...) -> str:
+        ...
+
+    @overload
+    def param_get(self, param_name: str, key: Optional[str] = ...,
+                  accept_unknown_key: Literal[True] = ...) -> Optional[str]:
+        ...
+
     def param_get(self, param_name: str, key: Optional[str] = None,
-                  accept_unknown_key: bool = False) -> Union[str, bool, None]:
+                  accept_unknown_key: bool = False) -> Optional[str]:
         return _param_get(self.host, self.xe_prefix, self.uuid,
                           param_name, key, accept_unknown_key)
 
--- a/lib/host.py
+++ b/lib/host.py
@@ -207,9 +207,9 @@ class Host:
             finally:
                 self.ssh(['rm', '-f', script.name])
 
-    def _get_xensource_inventory(self):
+    def _get_xensource_inventory(self) -> Dict[str, str]:
         output = self.ssh(['cat', '/etc/xensource-inventory'])
-        inventory: dict[str, str] = {}
+        inventory: Dict[str, str] = {}
         for line in output.splitlines():
             key, raw_value = line.split('=')
             inventory[key] = raw_value.strip('\'')
--- a/lib/host.py
+++ b/lib/host.py
@@ -6,14 +6,14 @@ import tempfile
 import uuid
 
 from packaging import version
-from typing import TYPE_CHECKING, Dict, List, Literal, Optional, Union, overload
+from typing import Dict, List, Literal, Optional, overload, TYPE_CHECKING, Union
 
 import lib.commands as commands
 import lib.pif as pif
 if TYPE_CHECKING:
     import lib.pool
 
-from lib.common import _param_add, _param_clear, _param_get, _param_remove, _param_set
+from lib.common import _param_add, _param_clear, _param_get, _param_remove, _param_set, strtobool
 from lib.common import safe_split, strip_suffix, to_xapi_bool, wait_for, wait_for_not
 from lib.common import prefix_object_name
 from lib.netutil import wrap_ip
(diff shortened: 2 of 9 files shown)


Nambrok:
I'm also pondering dropping the implicit cast from `bool` to `str` using `to_xapi_bool` in `xe` for `args` parameter that end up making us needing to cast dict in `Dict[str, Union[str, bool]]`

glehmann (changes requested):

lib/basevm.py
```diff
@@ -13,14 +13,30 @@ class BaseVM:
...
+        ...
+
+    @overload
+    def param_get(self, param_name: str, key: Optional[str] = None,
+                  accept_unknown_key: Literal[True] = True) -> Optional[str]:
+        ...
+
+    @overload
+    def param_get(self, param_name: str, key: Optional[str],
+                  accept_unknown_key: bool) -> Optional[str]:
```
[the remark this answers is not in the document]
Nambrok: You are right, I was forced to add the `bool` one last because having both Literal False and True wasn't enough to recognize a bool.

lib/basevm.py
```diff
@@ -13,14 +13,30 @@ class BaseVM:
...
-    def __init__(self, uuid, host: 'lib.host.Host'):
+    def __init__(self, uuid: str, host: 'lib.host.Host'):
         logging.info("New %s: %s", type(self).__name__, uuid)
         self.uuid = uuid
         self.host = host
 
+    @overload
     def param_get(self, param_name: str, key: Optional[str] = None,
-                  accept_unknown_key: bool = False) -> Union[str, bool, None]:
+                  accept_unknown_key: Literal[False] = False) -> str:
```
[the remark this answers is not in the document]
Nambrok: I can do it in this PR since I already touch up `xe` type definition if you want @glehmann

lib/basevm.py
```diff
@@ -13,14 +13,25 @@ class BaseVM:
...
         self.host = host
 
+    @overload
+    def param_get(self, param_name: str, key: Optional[str] = ...,
+                  accept_unknown_key: Literal[False] = ...) -> str:
+        ...
+
+    @overload
+    def param_get(self, param_name: str, key: Optional[str] = ...,
+                  accept_unknown_key: bool = ...) -> Optional[str]:
```
ydirson, line 30: shouldn't that rather be `accept_unknown_key: Literal[True]`?

lib/basevm.py
```diff
@@ -13,14 +13,25 @@ class BaseVM:
...
+        ...
+
     def param_get(self, param_name: str, key: Optional[str] = None,
-                  accept_unknown_key: bool = False) -> Union[str, bool, None]:
+                  accept_unknown_key: bool = False) -> Optional[str]:
```
ydirson, line 34: There shouldn't be a need to keep any type hints here, since everything is in the overloads now, ... or is there?

lib/host.py
```diff
@@ -328,7 +324,7 @@ def import_iso(self, uri, sr: SR):
 
         download_path = None
         try:
-            params = {'uuid': vdi_uuid}
+            params: Dict[str, Union[str, bool]] = {'uuid': vdi_uuid}
```
ydirson, line 327: That one feels strange, as we're only using `str` values, but that seems to be required because `xe()` wants that type.  It looks like that constraint only makes sense since `xe()` can in theory modify its `params` argument, and that the constraint would go away if we could annotate `params` as input-only argument, but I can't find if we can do that.

ydirson (changes requested):

lib/host.py
Nambrok: It's indeed because `xe` `args` parameter is broken: https://github.com/xcp-ng/xcp-ng-tests/pull/292#issuecomment-2778002055
It's not really in the scope of the PR indeed, I will remove it.

lib/basevm.py
Nambrok: No, it doesn't happen because `_param_get` has all of the overload needed. But having `Literal[False]` and `Literal[True]` is not enough to be recognized as `bool`.
https://mypy.readthedocs.io/en/stable/literal_types.html#literal-types

lib/basevm.py
glehmann: I tried to remove them but then pyright is not detecting the type properly in the IDE (while not complaining when running from the command line)

ydirson: oh. That could be an issue to raise to the pyright project?

lib/basevm.py
ydirson: Oh.  But what I understand of this is that in the general case we need 3 overloads, 2 for the literal values, and one for bool variables, when we cannot tell.  It happens that in our case the union of the return types is the return type for the `Literal[False]` case, but then it would at least warrant a comment explaining that (or just provide the 3 overloads).

Nambrok: Indeed, that was I did at first: https://github.com/xcp-ng/xcp-ng-tests/pull/292#discussion_r2027746839 ^^

lib/basevm.py
Nambrok: Ok, it should be better, removed the type hint on the real function, the `lib/common.py::_param_get` has a fall back for bool that is working on my own editor pyright.

lib/host.py
Nambrok: I can't remove it without making pyright unhappy

lib/basevm.py
glehmann: your pyright is able to find the type of `param_name`, where it fails to do that in the screenshot above?

Nambrok: Nope, not this, I misunderstood your problem.

glehmann: ok, too bad, it would have been better with just the annotations on the overloads :)
We can maybe work on Pyright to make it work, but in the meantime, please keep the type annotations on the functions args: that's definitely some useful info!

lib/basevm.py
```diff
@@ -13,14 +13,25 @@ class BaseVM:
...
+        ...
+
+    @overload
+    def param_get(self, param_name: str, key: Optional[str] = ...,
+                  accept_unknown_key: Literal[True] = ...) -> Optional[str]:
+        ...
+
     def param_get(self, param_name: str, key: Optional[str] = None,
-                  accept_unknown_key: bool = False) -> Union[str, bool, None]:
+                  accept_unknown_key: bool = False) -> Optional[str]:
```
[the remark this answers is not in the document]
stormi: When you have several overloads for the same function, I find it nice to opt for a compromise between absolute readability and vertical space. To some, more information in one window, without having to scroll, is beneficial (without falling into the adverse extreme where everything would be made ultra compact at the expense of readability).

Personnally, I find that in this example, what has been done is a good compromise.

lib/sr.py
```diff
@@ -41,8 +41,10 @@ def unplug_pbds(self, force=False):
     def all_pbds_attached(self):
         all_attached = True
         for pbd_uuid in self.pbd_uuids():
-            all_attached = all_attached and self.pool.master.xe('pbd-param-get', {'uuid': pbd_uuid,
-                                                                'param-name': 'currently-attached'})
+            all_attached = all_attached and strtobool(self.pool.master.xe('pbd-param-get',
+                                                                          {'uuid': pbd_uuid,
+                                                                           'param-name': 'currently-attached',
+                                                                           }))
```
glehmann, line 47:
```suggestion
            attached = self.pool.master.xe('pbd-param-get', {'uuid': pbd_uuid, 'param-name': 'currently-attached'})
            all_attached = all_attached and strtobool(attached)
```

might be easier to read

glehmann (approved):

lib/basevm.py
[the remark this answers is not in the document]
glehmann: When I look at the overloads, I want to be able to compare the types of the overloads easily. In that regard, I find them both not very easy to read. A much easier version for me would be:

~~~python
    @overload
    def param_get(self, param_name: str, key: Optional[str] = ...,  accept_unknown_key: Literal[False] = ...) -> str:
        ...
    @overload
    def param_get(self, param_name: str, key: Optional[str] = ...,  accept_unknown_key: Literal[True] = ...)  -> Optional[str]:
        ...
    def param_get(self, param_name: str, key: Optional[str] = None, accept_unknown_key: bool = False)         -> Optional[str]:
~~~

but then I'm breaking quite a few pep8 rules…

Coding style discussions are always fun ;-)

Wescoeur (approved):

PR #293, opened 2025-04-04
Typing fixes

ydirson:
- fixes a mistake we missed in #284 that only shows with older python and would likely bomb out on current Jenkins setup
- adds a few method hints to deal with mypy reporting `annotation-unchecked` "notes" since #284

### DIFF
--- a/lib/host.py
+++ b/lib/host.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import logging
 import os
 import shlex
--- a/lib/vm.py
+++ b/lib/vm.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import logging
 import os
 import subprocess


lib/vm.py
```diff
@@ -12,8 +12,11 @@
 from lib.snapshot import Snapshot
 from lib.vif import VIF
 
+if TYPE_CHECKING:
+    from lib.host import Host
```
glehmann, line 16:
```suggestion
    import lib.host
```

lib/vm.py
```diff
@@ -12,8 +12,11 @@
...
+
 class VM(BaseVM):
-    def __init__(self, uuid, host):
+    def __init__(self, uuid: str, host: 'Host'):
```
glehmann, line 19:
```suggestion
    def __init__(self, uuid: str, host: 'lib.Host'):
```

glehmann (changes requested):

lib/vm.py
ydirson: Why?  I noticed you did this in the previous branch, but it deviates from the use we use the classes outside of typing context, and if makes the hints themselves IMO-unnecessarily longer.

lib/host.py
```diff
@@ -201,7 +201,7 @@ def execute_script(self, script_contents, shebang='sh', simple_output=True):
             finally:
                 self.ssh(['rm', '-f', script.name])
 
-    def _get_xensource_inventory(self):
+    def _get_xensource_inventory(self) -> dict[str, str]:
```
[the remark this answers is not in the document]
ydirson: Duh, right

ydirson:
The second commit is unrelated with fixing the CI, and in fact clashes with #292, so I'll drop it from this PR

glehmann (approved):

Nambrok (approved):

PR #295, opened 2025-04-06
fix broken xo_cli's bytes return type

glehmann:
and the the invalid call that was added because of it


gduperrey:
* 8.3 - ipv6 - xo :white_check_mark: 
* 8.2 - xo :white_check_mark:

lib/host.py
```diff
@@ -242,7 +242,7 @@ def xo_server_add(self, username, password, label=None, unregister_first=True):
                 'allowUnauthorized': 'true',
                 'label': label
             }
-        ).decode()
+        )
```
Nambrok, line 245: Should this be done by `xo_cli` itself? There other users of `xo_cli` that seem to expect a `str` like here: https://github.com/xcp-ng/xcp-ng-tests/pull/295/files#diff-979b2af8ef5ed828b12e562799c4ae87c3eeeaadafa3ae0074ec6962b6a3f799R213

stormi (approved):

lib/xo.py
```diff
@@ -17,9 +17,9 @@ def xo_cli(action: str, args: Dict[str, str] = {}, *, check: bool = True, simple
     ...
 @overload
 def xo_cli(action: str, args: Dict[str, str] = {}, *, check: bool = True, simple_output: bool = True,
-           use_json: bool = False) -> Union[subprocess.CompletedProcess, Any, bytes]:
+           use_json: bool = False) -> Union[subprocess.CompletedProcess, Any, str]:
     ...
-def xo_cli(action, args={}, check=True, simple_output=True, use_json=False):
+def xo_cli(action, args: Dict[str, str] = {}, check=True, simple_output=True, use_json=False):
```
Nambrok, line 22: Why the type hint on the function when there is overload for this with the hint already?

glehmann: to have the type annotation in the IDE (with pyright at least).

Nambrok: But then it might be better to get the one with all type on the function and remove the generic overload?

glehmann: there is only one type missing for `action`, the others being infered with the default value.
But I agree, it's better to merge the last overload and the main function signature :+1:

Nambrok (approved):

PR #299, opened 2025-04-15
lint python code with ruff

glehmann:
Ruff (or any linter) helps catch mistakes and messy code before they cause headaches. It points out things like unused imports, style issues, or potential bugs, so we don’t have to hunt them down later. It also keeps our code consistent and easy to read, which is great when working in team or with external contributors.

Ruff is very fast, easy to set up in the CI and in our IDEs.

I've used the same rules as in the xcp repository, but we probably want to remove some of them or add some more.
Look at the commit that fixes errors to get an idea of what you may or may not like.
All the available rules are documented at https://docs.astral.sh/ruff/rules/

### DIFF
--- a/.github/workflows/code-checkers.yml
+++ b/.github/workflows/code-checkers.yml
@@ -32,3 +32,18 @@ jobs:
       run: cp data.py-dist data.py
     - name: Check with pyright
       run: uv run pyright lib/ conftest.py pkgfixtures.py # tests/
+
+  ruff:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Install uv
+      uses: astral-sh/setup-uv@v6
+      with:
+        version: "0.7.x"
+    - name: Install dependencies
+      run: uv sync --frozen
+    - name: Create a dummy data.py
+      run: cp data.py-dist data.py
+    - name: Check with ruff
+      run: uv run ruff check lib/ tests/
--- a/conftest.py
+++ b/conftest.py
@@ -1,11 +1,13 @@
+from typing import Dict
+
+import pytest
+
 import itertools
 import logging
 import os
 import tempfile
-from typing import Dict
 
 import git
-import pytest
 from packaging import version
 
 import lib.config as global_config
--- a/conftest.py
+++ b/conftest.py
@@ -1,5 +1,3 @@
-from typing import Dict
-
 import pytest
 
 import itertools
@@ -32,6 +30,8 @@ from lib.xo import xo_cli
 # then import them in individual packages to fix the buggy package scope handling by pytest, we also
 # need to import them in the global conftest.py so that they are recognized as fixtures.
 from pkgfixtures import formatted_and_mounted_ext4_disk, sr_disk_wiped
+
+from typing import Dict
 
 # Do we cache VMs?
 try:
--- a/conftest.py
+++ b/conftest.py
@@ -1,19 +1,25 @@
 import itertools
-import git
 import logging
 import os
-import pytest
 import tempfile
-
-from packaging import version
 from typing import Dict
 
-import lib.config as global_config
+import git
+import pytest
+from packaging import version
 
+import lib.config as global_config
 from lib import pxe
-from lib.common import callable_marker, shortened_nodeid, prefix_object_name
-from lib.common import wait_for, vm_image, is_uuid
-from lib.common import setup_formatted_and_mounted_disk, teardown_formatted_and_mounted_disk
+from lib.common import (
+    callable_marker,
+    is_uuid,
+    prefix_object_name,
+    setup_formatted_and_mounted_disk,
+    shortened_nodeid,
+    teardown_formatted_and_mounted_disk,
+    vm_image,
+    wait_for,
+)
 from lib.netutil import is_ipv6
 from lib.pool import Pool
 from lib.sr import SR
--- a/lib/basevm.py
+++ b/lib/basevm.py
@@ -1,8 +1,6 @@
 import logging
 from typing import TYPE_CHECKING, Any, Literal, Optional, overload
 
-import lib.commands as commands
-
 if TYPE_CHECKING:
     import lib.host
 
--- a/lib/basevm.py
+++ b/lib/basevm.py
@@ -85,10 +85,7 @@ class BaseVM:
         return True
 
     def all_vdis_on_sr(self, sr) -> bool:
-        for vdi_uuid in self.vdi_uuids():
-            if self.host.pool.get_vdi_sr_uuid(vdi_uuid) != sr.uuid:
-                return False
-        return True
+        return all(self.host.pool.get_vdi_sr_uuid(vdi_uuid) == sr.uuid for vdi_uuid in self.vdi_uuids())
 
     def get_sr(self):
         # in this method we assume the SR of the first VDI is the VM SR
--- a/lib/basevm.py
+++ b/lib/basevm.py
@@ -1,8 +1,8 @@
 import logging
-
-from typing import Any, Literal, Optional, overload, TYPE_CHECKING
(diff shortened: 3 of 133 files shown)


glehmann:
no hurry on this review, it's there to trigger a discussion :-)

lib/commands.py
```diff
@@ -74,10 +74,7 @@ def _ssh(hostname_or_ip, cmd, check, simple_output, suppress_fingerprint_warning
         opts.append('-o "LogLevel ERROR"')
         opts.append('-o "UserKnownHostsFile /dev/null"')
 
-    if isinstance(cmd, str):
-        command = cmd
-    else:
-        command = " ".join(cmd)
+    command = cmd if isinstance(cmd, str) else ' '.join(cmd)
```
stormi, line 77: Why does ruff care at all about whether we use a simple if/else or a ternary?

lib/pool.py
```diff
@@ -1,14 +1,15 @@
...
-from typing import Any, Dict, Optional, cast
+from typing import Any, Dict, Optional
 
 from packaging import version
 
 import lib.commands as commands
 from lib.common import _param_get, _param_set, safe_split, wait_for, wait_for_not
 from lib.host import Host
 from lib.sr import SR
+import contextlib
```
stormi, line 12: Is it the right place for this import? I'd expect it grouped with system imports.

tests/security/test_tls.py
```diff
@@ -25,11 +25,11 @@ def test_tls_disabled(host: str, protocol_name: str):
 
     with pytest.raises(ssl.SSLError):
         context = ssl.SSLContext(protocol)
-        with socket.create_connection((str(host), PORT), timeout=10) as sock:
-            with context.wrap_socket(sock, server_hostname=str(host)) as ssock:
-                ssock.do_handshake()
-                # If we reach this point, the protocol is enabled (test should fail)
-                pytest.fail(f"Protocol {protocol} should be disabled but connection succeeded")
+        with socket.create_connection((str(host), PORT), timeout=10) as sock, \
+             context.wrap_socket(sock, server_hostname=str(host)) as ssock:
```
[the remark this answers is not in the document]
stormi: Are we on python 3.10 yet as the minimum requirement?

tests/storage/__init__.py
```diff
@@ -1 +1,7 @@
-from .storage import *
```
stormi, line 1: That's bad practice even in `__init__.py`?

tests/storage/iso/conftest.py
```diff
@@ -26,10 +26,7 @@ def copy_tools_iso_to_iso_sr(host, sr, location=None):
     # copy the ISO file to the right location
     iso_path = host.ssh(['find', '/opt/xensource/packages/iso/', '-name', '"*.iso"'])
     iso_new_name = sr.uuid + "_test.iso"
-    if location is not None:
-        iso_new_path = f"{location}/{iso_new_name}"
-    else:
-        iso_new_path = f"/run/sr-mount/{sr.uuid}/{iso_new_name}"
+    iso_new_path = f'{location}/{iso_new_name}' if location is not None else f'/run/sr-mount/{sr.uuid}/{iso_new_name}'
```
stormi, line 29: Mostly less readable because of the length of the paths. Not a blocker, but maybe that's not a rule that is really important to enforce.

image

tests/storage/linstor/test_linstor_sr.py
```diff
@@ -6,6 +6,7 @@
 from lib.commands import SSHCommandFailed
 from lib.common import wait_for, vm_image
 from tests.storage import vdi_is_open
+import contextlib
```
stormi, line 9: To be grouped with system imports?

tests/security/test_tls.py
```diff
@@ -25,11 +25,11 @@ def test_tls_disabled(host: str, protocol_name: str):
 
     with pytest.raises(ssl.SSLError):
         context = ssl.SSLContext(protocol)
-        with socket.create_connection((str(host), PORT), timeout=10) as sock:
-            with context.wrap_socket(sock, server_hostname=str(host)) as ssock:
-                ssock.do_handshake()
-                # If we reach this point, the protocol is enabled (test should fail)
-                pytest.fail(f"Protocol {protocol} should be disabled but connection succeeded")
+        with socket.create_connection((str(host), PORT), timeout=10) as sock, \
```
stormi, line 28: What's the rationale? Are nested `with` considered bad practice?

ydirson: less indentation, likely?

lib/commands.py
glehmann: not sure, but we don't have to use that specific rule

lib/pool.py
glehmann: yes, it also shows that there should be a rule for that :)

tests/security/test_tls.py
stormi: Ok, a really minor improvement then, and in this case, in my view, defeated by the need to go multiline. My personnal code scanning system (eyes + brains) catch successive short with a lot better than they read wrapped lines.

tests/security/test_tls.py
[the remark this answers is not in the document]
glehmann: We should switch to 3.13 right away (when we have uv).

lib/basevm.py
```diff
@@ -1,14 +1,13 @@
...
-from typing import Any, Literal, Optional, overload, TYPE_CHECKING
-
-import lib.commands as commands
 if TYPE_CHECKING:
     import lib.host
 
 from lib.common import _param_add, _param_clear, _param_get, _param_remove, _param_set
 from lib.sr import SR
 
+
```
stormi, line 10: Can ruff be made to reorder imports without enforcing these two lines? So many modified files...

.github/workflows/code-checkers.yml
```diff
@@ -38,3 +38,21 @@ jobs:
       run: cp data.py-dist data.py
     - name: Check with pyright
       run: pyright lib/ # tests/
+
+  ruff:
```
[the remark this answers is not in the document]
stormi: Then maybe my definition of static code checker was wrong/outdated.µ

But then... Isn't pycodestyle a static code checker too?

tests/security/test_tls.py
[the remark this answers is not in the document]
ydirson: Well, switching to 3.13 would be like forcing everyone but Gentoo and Arch users to use `uv`, maybe we don't want to take that step :)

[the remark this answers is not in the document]
glehmann: Fedora has it since version 41 :-)

At least let's require 3.11, which is the version shipped by debian stable, which was released two years ago.

BTW I'm ok with making people use `uv` if they can't get the dependencies, as long as it is simple and reliable, which is the case with `uv`.

stormi (approved):

conftest.py
```diff
@@ -1,19 +1,25 @@
...
-import git
 import logging
 import os
-import pytest
 import tempfile
+from typing import Dict
 
+import git
+import pytest
 from packaging import version
```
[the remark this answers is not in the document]
glehmann: feel free to give it a try and push a commit with the changes in that PR :+1: 

Once the configuration is ready, you should be able to fix all the files with `ruff check --select I001 --fix`.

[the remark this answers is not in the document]
ydirson: > Then yes, that wouldn't be exactly that order.
> Is that a problem?

I though `ruff` would complain, but then apparently not?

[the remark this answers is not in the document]
glehmann: ruff is as happy as it could :)

tests/security/test_tls.py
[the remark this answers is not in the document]
ydirson: We know have 3.11, so maybe we can get rid of the explicit continuation?

Issue #276, opened 2025-02-20
lib/host: disk_is_available function only validates mounted disks and not others


rushikeshjadhav:
In following case, disk_is_available('sdb') returns True which should not be the case.

https://github.com/xcp-ng/xcp-ng-tests/blob/358665a7bf3bfdd96777da3635af7d5c304fa5f2/lib/host.py#L467

```
[18:26 xcp-ng-xs1 ~]# lsblk 
NAME                                                                                     MAJ:MIN  RM   SIZE RO TYPE MOUNTPOINT
sdb                                                                                        8:16    0   500G  0 disk 
├─linstor_group-thin_device_tmeta                                                        253:1     0   128M  0 lvm  
│ └─linstor_group-thin_device-tpool                                                      253:3     0 499.8G  0 lvm  
│   ├─linstor_group-thin_device                                                          253:4     0 499.8G  0 lvm  
│   └─linstor_group-xcp--persistent--database_00000                                      253:5     0     1G  0 lvm  
│     └─drbd1000                                                                         147:1000  0     1G  0 disk 
└─linstor_group-thin_device_tdata                                                        253:2     0 499.8G  0 lvm  
  └─linstor_group-thin_device-tpool                                                      253:3     0 499.8G  0 lvm  
    ├─linstor_group-thin_device                                                          253:4     0 499.8G  0 lvm  
    └─linstor_group-xcp--persistent--database_00000                                      253:5     0     1G  0 lvm  
      └─drbd1000                                                                         147:1000  0     1G  0 disk 
sr0                                                                                       11:0     1  1024M  0 rom  
sdc                                                                                        8:32    0   500G  0 disk 
sda                                                                                        8:0     0    50G  0 disk 
├─sda4                                                                                     8:4     0   512M  0 part 
├─sda2                                                                                     8:2     0    18G  0 part 
├─sda5                                                                                     8:5     0     4G  0 part /var/log
├─sda3                                                                                     8:3     0   8.5G  0 part 
│ └─XSLocalEXT--8900cca8--706a--d7f2--447b--bff4129333a3-8900cca8--706a--d7f2--447b--bff4129333a3
                                                                                         253:0     0   8.5G  0 lvm  /run/sr-mount/8900cca8-706a-d7f2-447b-bff4129333a3
├─sda1                                                                                     8:1     0    18G  0 part /
└─sda6                                                                                     8:6     0     1G  0 part [SWAP]
```

[...]

PR #301, opened 2025-04-22
lib/host: Improved `disk_is_available()` using `blkid` for reliable disk checks.

rushikeshjadhav:
An issue ([https://github.com/xcp-ng/xcp-ng-tests/issues/276](https://github.com/xcp-ng/xcp-ng-tests/issues/276)) occurred while expanding a pool using/finding free disks of the hosts (PR [https://github.com/xcp-ng/xcp-ng-tests/pull/282](https://github.com/xcp-ng/xcp-ng-tests/pull/282)), hence fixing with this PR.
`blkid` returns 2 if there is nothing on the disk e.g. LVM, gpt, etc. hence marking `disk_is_available`.


stormi:
I fear regressions here. We use and reuse the same disks over and over in CI. After a test, they do have contents. We don't wipe them fresh everytime we unmount them.

rushikeshjadhav:
> I fear regressions here. We use and reuse the same disks over and over in CI. After a test, they do have contents. We don't wipe them fresh everytime we unmount them.

Do you mean that the content is "required to be kept" when unmounting or it's just not cleaned in those CI? May be a `wipefs` would do during a teardown.

Else we could introduce a different function `disk_is_raw` and handle it using `blkid` wherever suitable.

stormi:
It's just not cleaned, and we didn't have a reason to enforce that the disk be void of anything before starting tests. The SR creation takes care of overwriting what exists. That's why an available disk is defined as a disk that is not mounted. 

If XOSTOR has different requirements, maybe it should use a separate fixture?

rushikeshjadhav:
In XOSTOR, both used and unused disks are not mounted, so while finding a free disk on server that has "used" disk(s) during pool expansion, the current  `disk_is_available` marks both used and unused disks as available. Proposing to introduce `raw=True`, does it fit?

stormi (approved):
Works for me. To be seen if creating a new `raw_disk_is_available` function would be preferred by others.

lib/host.py
```diff
@@ -516,10 +516,12 @@ def disks(self):
         disks.sort()
         return disks
 
-    def disk_is_available(self, disk):
+    def disk_is_available(self, disk, disk_raw=False):
```
Wescoeur, line 519: I'm not sure this new param is explicit for everyone.
I prefer something like: `check_unformatted`, `check_no_fs` or `check_wiped`.

Wescoeur (changes requested):

lib/host.py
Nambrok: I also find its naming unclear

rushikeshjadhav: We can rename the parameter from `disk_raw` to `check_raw`, else for more elaborate use `type` option can be introduced. Where `type` can be `raw`, `LVM2_member`, `ext3`, `part` etc. and it can be found from `blkid` or suitable disk management command based on `type`.

e.g `disk_is_available(self, disk, type='raw')`

rushikeshjadhav: I have added `raw_disk_is_available()` and kept the current function untouched.

lib/host.py
```diff
@@ -516,6 +516,9 @@ def disks(self):
         disks.sort()
         return disks
 
+    def raw_disk_is_available(self, disk):
+        return self.ssh_with_result(['blkid', '/dev/' + disk]).returncode == 2
+
```
ydirson, line 521: A docstring explaining what this new method wrt `disk_is_available` would be useful, as well as one for the latter for easier comparison.

It could also be good to add type hints on such new functions.

ydirson (changes requested):

lib/host.py
rushikeshjadhav: Done

Nambrok (approved):

rushikeshjadhav:
@ydirson  @stormi  Kindly check this if its ready for merge.

stormi (approved):

stormi:
Checks did not run so pydocstyle failed after the merge to master.

@rushikeshjadhav please make sure to open future PRs from branches coming from this repository in which you have write rights IIRC, so that the style checks are run automatically.

PR #304, opened 2025-05-07
xen: Add test-ring0 tests

dinhngtu:
Run the test-ring0 tests that are straightforward to run.
The ones that are not (e.g. legacy tests, tests that crash the host or leak resources infinitely) are skipped, but included for completeness.

`test_xst_memory_leak` requires a custom kernel with `CONFIG_DEBUG_KMEMLEAK` set.

### DIFF
--- a/tests/xen/test_ring0.py
+++ b/tests/xen/test_ring0.py
@@ -1,0 +1,159 @@
+import logging
+import pytest
+import secrets
+import time
+from typing import Generator, Optional
+
+from lib.host import Host
+from lib.vm import VM
+
+# Requirements:
+# From --hosts parameter:
+# - host: XCP-ng host >= 8.2, reboots after test
+# From --vm parameter:
+# - A VM to import
+# Only XST tests with pass/fail results, and that don't crash the host were included.
+
+
+@pytest.fixture(scope="package")
+def host_with_ring0_tests(
+    host_with_saved_yum_state: Host,
+) -> Generator[Host, None, None]:
+    host = host_with_saved_yum_state
+    host.yum_install(["test-ring0"])
+    yield host
+    # clean up the loaded test modules and test states at the end
+    host.reboot(verify=True)
+
+
+@pytest.fixture
+def host_without_livepatch_loaded(host: Host) -> Host:
+    if host.ssh_with_result("lsmod | grep -wq livepatch_tester").returncode == 0:
+        pytest.fail("livepatch_tester already loaded, host needs reboot")
+    return host
+
+
+def do_execute_xst(host: Host, modname: str, testname: Optional[str] = None) -> None:
+    if testname is None:
+        testname = modname
+    host.ssh(f"modprobe xst_{modname}")
+    try:
+        host.ssh(f"echo 1 > /sys/kernel/debug/xst/{testname}/run")
+        host.ssh(f"grep -q 'status: pass' /sys/kernel/debug/xst/{testname}/results")
+    finally:
+        host.ssh(f"modprobe -r xst_{modname}", check=False)
+
+
+@pytest.mark.reboot  # host_with_ring0_tests
+@pytest.mark.usefixtures("host_with_ring0_tests")
+class TestRing0Tests:
+    def test_privcmd_restrict(self, host: Host):
+        host.ssh("/usr/bin/privcmd-restrict_test")
+
+    def test_xst_alloc_balloon(self, host: Host):
+        do_execute_xst(host, "alloc_balloon")
+
+    def test_xst_big_module(self, host: Host):
+        do_execute_xst(host, "big_module")
+
+    def test_xst_evtchn_latency(self, host: Host):
+        do_execute_xst(host, "evtchn_latency", "evtchn_lat")
+
+    @pytest.mark.skip("only makes sense for 2l evtchn")
+    def test_xst_evtchn_limit(self, host: Host):
+        do_execute_xst(host, "evtchn_limit")
+
+    def test_xst_evtchn_stress(self, host: Host):
+        do_execute_xst(host, "evtchn_stress")
+
+    @pytest.mark.skip("leaks event channels infinitely")
+    def test_xst_evtchn_unbind(self, host: Host):
+        do_execute_xst(host, "evtchn_unbind")
+
+    def test_xst_get_user_pages(self, host: Host):
+        host.ssh("modprobe xst_get_user_pages")
+        try:
+            host.ssh("/usr/bin/gup_test")
+        finally:
+            host.ssh("modprobe -r xst_get_user_pages", check=False)
+
+    def test_xst_grant_copy_perf(self, host: Host):
+        do_execute_xst(host, "grant_copy_perf", "gntcpy_perf")
+
+    @pytest.mark.small_vm
+    def test_xst_ioemu_msi(self, host: Host, running_unix_vm: VM):
+        # TODO: validate MSI reception in guest
+        vm = running_unix_vm
+        domid = vm.param_get("dom-id")
+        host.ssh("modprobe xst_ioemu_msi")
+        try:
+            host.ssh(f"echo {domid} > /sys/kernel/debug/xst/ioemu_msi/domid")
+            host.ssh("echo 1 > /sys/kernel/debug/xst/ioemu_msi/data")
+            host.ssh("echo 1 > /sys/kernel/debug/xst/ioemu_msi/run")
+            host.ssh("grep -q 'status: pass' /sys/kernel/debug/xst/ioemu_msi/results")
+        finally:
+            host.ssh("modprobe -r xst_ioemu_msi", check=False)
+
+    @pytest.mark.usefixtures("host_at_least_8_3")
+    def test_xst_livepatch(self, host_without_livepatch_loaded: Host):
+        """
+        This test loads a `livepatch_testee` module, and triggers the test
+        function `test_function_default` by writing to
+        `/proc/livepatch_testee/cmd`. This function is then updated by loading
+        `livepatch_tester` and retested. Distinguish between unpatched and
+        patched functions using printk output, delimited by a random marker.
+
+        Strangely enough, the patch in `livepatch_tester` causes the patched
+        `test_function_crash` to crash the host (instead of the other way
+        around). So don't test that.
+        """
+        host = host_without_livepatch_loaded
+        try:
+            host.ssh("modprobe livepatch_testee")
+
+            marker = secrets.token_hex()
+            logging.debug(f"using pre-patch marker {marker}")
+            host.ssh(f"echo {marker} > /dev/kmsg")
+            host.ssh("echo 1 > /proc/livepatch_testee/cmd")
+            host.ssh(f"dmesg | grep -A 9999 {marker} | grep -q test_function_default_old")
+
+            host.ssh("modprobe livepatch_tester")
+
+            marker = secrets.token_hex()
+            logging.debug(f"using post-patch marker {marker}")
+            host.ssh(f"echo {marker} > /dev/kmsg")
+            host.ssh("echo 1 > /proc/livepatch_testee/cmd")
+            host.ssh(f"dmesg | grep -A 9999 {marker} | grep -q test_function_default_new")
+        finally:
+            host.ssh("modprobe -r livepatch_testee", check=False)
+
+    def test_xst_memory_leak(self, host: Host):
+        if not host.file_exists("/sys/kernel/debug/kmemleak"):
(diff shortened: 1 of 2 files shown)


tests/xen/conftest.py
```diff
@@ -3,6 +3,7 @@
 
 from packaging import version
 
+from lib.host import Host
```
stormi, line 6: I think this ideally belongs to the line just before `from packaging import version`

tests/xen/test_ring0.py
```diff
@@ -0,0 +1,135 @@
+import logging
+import secrets
+import time
+from typing import Optional
+import pytest
+
+from lib.commands import SSHCommandFailed
+from lib.host import Host
+from lib.vm import VM
```
stormi, line 9: Needs to be reordered. Straight imports first, alphabetically. Then a blank line. Then "from xxx" imports, alphabetically.

tests/xen/test_ring0.py
```diff
@@ -0,0 +1,135 @@
...
+            host.ssh("echo 1 > /sys/kernel/debug/xst/ioemu_msi/data")
+            host.ssh("echo 1 > /sys/kernel/debug/xst/ioemu_msi/run")
+            host.ssh("grep -q 'status: pass' /sys/kernel/debug/xst/ioemu_msi/results")
+        finally:
+            host.ssh("modprobe -r xst_ioemu_msi", check=False)
+
+    def test_xst_livepatch(self, host: Host):
+        try:
+            host.ssh("lsmod | grep -wq livepatch_tester")
+            pytest.skip("livepatch_tester already loaded, needs reboot")
```
stormi, line 83: We avoid skipping tests for such reasons. One might falsely think that the test was executed, in CI, as no one checks all skips one by one. What we do instead is check for prerequisites in a fixture and fail the test if the prerequisite is not met.

ydirson (changes requested):

tests/xen/conftest.py
ydirson: Not necessarily, as it is an import from a private module, while `packaging` is an external one.

tests/xen/test_ring0.py
```diff
@@ -0,0 +1,135 @@
...
+# Only XST tests with pass/fail results, and that don't crash the host were included.
+
+
+def do_execute_xst(host: Host, modname: str, testname: Optional[str] = None):
+    if testname is None:
+        testname = modname
+    host.ssh(f"modprobe xst_{modname}")
+    try:
+        host.ssh(f"echo 1 > /sys/kernel/debug/xst/{testname}/run")
+        host.ssh(f"grep -q 'status: pass' /sys/kernel/debug/xst/{testname}/results")
```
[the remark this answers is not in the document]
dinhngtu: Yes, the tests are synchronous.

tests/xen/test_ring0.py
stormi, line 1: As it is, the test will be executed as part of the `xen` job. This job is currently very short and runs at the beginning of our test pipeline, on real hardware. However, with a reboot, we'll go from 1 minute to close to 10 minutes. 

Are the tests themselves long to execute? 

Our options:
- keep as is
- move to a separate job that we'll run later in the pipeline
- split between fast tests that would not require a reboot (if any) and the rest.

tests/xen/conftest.py
stormi: Oh, right. Forget about my comment then :)

tests/xen/test_ring0.py
```diff
@@ -0,0 +1,135 @@
...
+    finally:
+        host.ssh(f"modprobe -r xst_{modname}", check=False)
```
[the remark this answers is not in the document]
dinhngtu: Mostly since failure to remove test modules has no bearing on the test results, and the host is cleaned up afterwards by rebooting anyway. Should I make removal failures deadly?

tests/xen/test_ring0.py
stormi: As Yann indicated above, I forgot about the split between external and internal libs, too.

tests/xen/test_ring0.py
dinhngtu: The tests themselves are quite short. I can split it to a separate job if desired.

tests/xen/test_ring0.py
ydirson: I would not use `fail`, that's rather what `error` is for

tests/xen/test_ring0.py
stormi: We can keep them in the same job for now.

stormi: Please update the `xen` job description in jobs.py to indicate that it will reboot the master host.

tests/xen/test_ring0.py
dinhngtu: I couldn't find `pytest.error` anywhere.

tests/xen/test_ring0.py
stormi: Stack Overflow most voted answer pushes for sorting, and that's also the convention we tend to use.

https://stackoverflow.com/questions/20762662/whats-the-correct-way-to-sort-python-import-x-and-from-x-import-y-statement

tests/xen/test_ring0.py
stormi: By moving the prerequisite test to a fixture as I suggested (and that's one of the things fixtures are here for in pytest), it would be obvious that the failure is in the prerequisites and not in the test itself.

tests/xen/test_ring0.py
ydirson: I thought you were originally refering to `pytest` not being sorted :wink:

tests/xen/test_ring0.py
ydirson: Yes, my bad, there does not seem to be another mechanism than fixtures to generate error rather than failure (https://docs.pytest.org/en/stable/explanation/fixtures.html#fixture-errors seems to be where they talk about it, but it does not spell out precisely that errors are solely coming from fixtures)

ydirson: Next question would be, do we want to do this in a `host_without_livepatch_tester` fixture to be able to get an error rather than a failure?

stormi: That's precisely what I'm talking about from the beginning :D (though in other fixtures we do use pytest.fail too at the moment)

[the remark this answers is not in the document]
dinhngtu: Fixed with `host_without_livepatch_loaded` fixture.

tests/xen/test_ring0.py
```diff
@@ -0,0 +1,135 @@
...
+            host.ssh("grep -q 'status: pass' /sys/kernel/debug/xst/ioemu_msi/results")
+        finally:
+            host.ssh("modprobe -r xst_ioemu_msi", check=False)
+
+    def test_xst_livepatch(self, host: Host):
+        try:
+            host.ssh("lsmod | grep -wq livepatch_tester")
+            pytest.skip("livepatch_tester already loaded, needs reboot")
+        except SSHCommandFailed:
+            pass
```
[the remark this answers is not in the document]
dinhngtu: Fixed.

tests/xen/test_ring0.py
```diff
@@ -0,0 +1,135 @@
...
+        host.ssh("modprobe xst_ioemu_msi")
+        try:
+            host.ssh(f"echo {domid} > /sys/kernel/debug/xst/ioemu_msi/domid")
+            host.ssh("echo 1 > /sys/kernel/debug/xst/ioemu_msi/data")
+            host.ssh("echo 1 > /sys/kernel/debug/xst/ioemu_msi/run")
+            host.ssh("grep -q 'status: pass' /sys/kernel/debug/xst/ioemu_msi/results")
+        finally:
+            host.ssh("modprobe -r xst_ioemu_msi", check=False)
+
+    def test_xst_livepatch(self, host: Host):
```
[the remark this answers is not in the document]
dinhngtu: Comment added.

tests/xen/test_ring0.py
dinhngtu: Fixed.

tests/xen/test_ring0.py
dinhngtu: Fixed with the suggested fixture.

tests/xen/test_ring0.py
dinhngtu: Fixed.

stormi:
LGTM. As this adds a reboot to the xen test, we need to test it in CI conditions to see how it behaves, before merging.

CC @gduperrey

stormi (approved):

tests/xen/test_ring0.py
```diff
@@ -0,0 +1,146 @@
...
+
+# Requirements:
+# From --hosts parameter:
+# - host: XCP-ng host >= 8.2, reboots after test
+# From --vm parameter:
+# - A VM to import
+# Only XST tests with pass/fail results, and that don't crash the host were included.
+
+
+def do_execute_xst(host: Host, modname: str, testname: Optional[str] = None):
```
ydirson, line 18: type hint `-> None`

tests/xen/test_ring0.py
```diff
@@ -0,0 +1,146 @@
...
+    host.ssh(f"modprobe xst_{modname}")
+    try:
+        host.ssh(f"echo 1 > /sys/kernel/debug/xst/{testname}/run")
+        host.ssh(f"grep -q 'status: pass' /sys/kernel/debug/xst/{testname}/results")
+    finally:
+        host.ssh(f"modprobe -r xst_{modname}", check=False)
+
+
+@pytest.fixture
+def host_without_livepatch_loaded(host: Host):
```
ydirson, line 30: `-> Generator[Host, None, None]`

tests/xen/conftest.py
```diff
@@ -65,3 +66,11 @@ def host_with_dom0_tests(host_with_saved_yum_state):
     host = host_with_saved_yum_state
     host.yum_install(['xen-dom0-tests'])
     yield host
+
+@pytest.fixture(scope="package")
+def host_with_ring0_tests(host_with_saved_yum_state: Host):
```
ydirson, line 71: What's the rationale of having this fixture in `conftest` ?  That seems really specific to `test_ring0` (which incidentally includes another fixture)

tests/xen/test_ring0.py
```diff
@@ -0,0 +1,146 @@
...
+    if host.ssh_with_result("lsmod | grep -wq livepatch_tester").returncode == 0:
+        pytest.fail("livepatch_tester already loaded, host needs reboot")
```
ydirson, line 32: The goal of the fixture is to get an error rather than a failure.  Even if `pytest.fail` achieves that it looks awkward, wouldn't a simple `assert` do the job?

tests/xen/test_ring0.py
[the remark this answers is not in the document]
ydirson: maybe add that the printk's are triggered by the `cmd` file?  Not that hard to understand, but that still required a few back-and-forth

ydirson (changes requested):
A few cosmetic remarks, otherwise lgtm

tests/xen/test_ring0.py
dinhngtu: Done.

tests/xen/conftest.py
dinhngtu: Moved fixture to `test_ring0`.

gduperrey:
2025-06-19:
* 8.2 xen ✅
* 8.3 xen ✅

tests/xen/test_ring0.py
ydirson: Looks like it can even be a good idea, as the `Failed` exception raised by `pytest.fail` does not derive from `Exception` and should not be silenced by error by `except Exception` statements.

tests/xen/test_ring0.py
```diff
@@ -0,0 +1,158 @@
...
+# From --hosts parameter:
+# - host: XCP-ng host >= 8.2, reboots after test
+# From --vm parameter:
+# - A VM to import
+# Only XST tests with pass/fail results, and that don't crash the host were included.
+
+
+@pytest.fixture(scope="package")
+def host_with_ring0_tests(host_with_saved_yum_state: Host) \
+        -> Generator[Host, None, None]:
```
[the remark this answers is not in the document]
dinhngtu: On the contrary, this style is used just about everywhere in our tests.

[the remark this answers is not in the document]
dinhngtu: Applied formatter.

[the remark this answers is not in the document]
ydirson: well, the additional newline may not be necessary, but I can live with it :)

PR #306, opened 2025-05-19
lib/vm: Work around Windows XS guest agents not updating version after resume

dinhngtu:
Resuming after a suspend wipes out xenstore information including `PV-drivers-version`. Check for both the agent version or the xenbus version to work around an issue where the Windows agent sometimes fails to update its version after a resume.

### DIFF
--- a/lib/vm.py
+++ b/lib/vm.py
@@ -148,7 +148,11 @@ class VM(BaseVM):
             return False
 
     def is_management_agent_up(self):
-        return self.param_get('PV-drivers-version', 'major', accept_unknown_key=True) is not None
+        return (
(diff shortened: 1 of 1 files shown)


ydirson (approved):
no objection, at the very least it will confirm whether this is the case that happens so often these days

dinhngtu:
That might be better since we haven't completely shipped the WinPV driver version reporting fix yet. Still, `wait_for_os_booted` is the only place where we use `is_management_agent_up` so we might just be able to delete `is_management_agent_up` altogether (`try_get_and_store_ip` should be enough to prove that OS booted right?)

stormi (approved):

PR #310, opened 2025-05-21
Add test for rolling pool update with VM availability check on updated host

rushikeshjadhav:
Added `test_linstor_sr_pool_update` for rolling pool update with VM availability check.

It performs
- A rolling update and reboot of all hosts in a LINSTOR SR pool (starting with the master).
- Verifies VM can start and shutdown successfully on each host after update.
- Ensures SR remains usable throughout the process.

### DIFF
--- a/tests/storage/linstor/test_linstor_sr.py
+++ b/tests/storage/linstor/test_linstor_sr.py
@@ -131,6 +131,56 @@ class TestLinstorSR:
             if not linstor_installed:
                 host.yum_install([LINSTOR_PACKAGE])
 
+    @pytest.mark.reboot
+    @pytest.mark.small_vm
+    @pytest.mark.upgrade_test
+    def test_linstor_sr_pool_update(self, linstor_sr, vm_on_linstor_sr):
+        """
+        Perform update on the Linstor SR pool hosts while ensuring VM availability.
+        1. Identify all hosts in the SR pool and order them with the master first.
+        2. Update all hosts if updates are available.
+        3. Reboot all hosts.
+        4. Sequentially ensure that the VM can start on all hosts.
+        """
+        import concurrent.futures, threading
+
+        sr = linstor_sr
+        vm = vm_on_linstor_sr
+        updates_applied = []
+        updates_lock = threading.Lock()
+
+        # Sort hosts so that pool master is first (optional)
+        hosts = sorted(sr.pool.hosts, key=lambda h: h != sr.pool.master)
+
+        # RPU is disabled for pools with XOSTOR SRs.
+        # LINSTOR expects that we always use satellites and controllers with the same version on all hosts.
+        def install_updates_on(host):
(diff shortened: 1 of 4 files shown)


tests/storage/linstor/test_linstor_sr.py
stormi, line 1: The tricky part here is ensuring that we do have packages to test the rolling pool update. That means starting with a pool which has an older version of the packages compared to the ones we want to test. That also likely means moving this kind of test to a specific job, or have an orchestrator test (or fixtures) that prepares a nested pool for this test.

See with @ydirson, as I think there are common challenges with the installation and upgrade tests.

tests/storage/linstor/test_linstor_sr.py
```diff
@@ -131,6 +131,32 @@ def test_linstor_missing(self, linstor_sr, host):
...
+
+        # Ensure that VM is able to start on all post applying available updates on hosts, starting with master
+        for h in hosts:
+            logging.info("Checking on host %s", h.hostname_or_ip)
+            if h.has_updates():
+                h.install_updates()
+                h.reboot(verify=True)
+                vm.start(on=h.uuid)
+                vm.wait_for_os_booted()
+                vm.shutdown(verify=True)
```
stormi, line 156: This doesn't look like the proper way to upgrade a pool with linstor. They don't support rolling upgrade, so you have to update all satellites and restart them, at some point. @Wescoeur has this documented somewhere.

tests/storage/linstor/test_linstor_sr.py
rushikeshjadhav: So should we limit this to `xcp-ng-linstor-testing` packages to keep it in linstor tests? Along side of main update tests, this can be specific for linstor related testing package tests.

Larger installation and upgrade tests can happen in different job.

```
# repoquery -a --repoid=xcp-ng-linstor-testing
linstor-common-0:1.29.2-1.el7_9.noarch
linstor-controller-0:1.29.2-1.el7_9.noarch
linstor-satellite-0:1.29.2-1.el7_9.noarch
```

```
# repoquery -a --repoid=xcp-ng-linstor
linstor-common-0:1.29.0-1.el7_9.noarch
linstor-controller-0:1.29.0-1.el7_9.noarch
linstor-satellite-0:1.29.0-1.el7_9.noarch
```

tests/storage/linstor/test_linstor_sr.py
rushikeshjadhav: Yes, it's https://docs.xcp-ng.org/management/updates/#rolling-pool-update-rpu I have updated the code considering above.

ydirson (changes requested):
In addition, please format the commit message [as per the standard converntions](https://www.baeldung.com/ops/git-commit-messages#bd-1-structure-for-good-commit-messages)

PR #313, opened 2025-05-21
Add host failure test to verify VM resiliency and SR stability

rushikeshjadhav:
Added `test_linstor_sr_fail_host` to simulate a crash of a non-master host.
- Chooses a host within a LINSTOR SR pool and simulate crash using sysrq-trigger.
- Verifies VM boot and shutdown on all remaining hosts during the outage, and confirms recovery of the failed host for VM placement post-reboot.
- Ensures SR scan consistency post-recovery.

### DIFF
--- a/tests/storage/linstor/test_linstor_sr.py
+++ b/tests/storage/linstor/test_linstor_sr.py
@@ -131,6 +131,48 @@ class TestLinstorSR:
             if not linstor_installed:
                 host.yum_install([LINSTOR_PACKAGE])
 
+    @pytest.mark.reboot
+    @pytest.mark.small_vm
+    def test_linstor_sr_fail_host(self, linstor_sr, host, vm_on_linstor_sr):
+        """
+        Fail non master host from the same pool Linstor SR.
+        Ensure that VM is able to boot and shutdown on all hosts.
+        """
+        import random
+        sr = linstor_sr
+        vm = vm_on_linstor_sr
+        # Ensure that its a single host pool and not multi host pool
+        assert len(host.pool.hosts) > 2, "This test requires Pool to have more than 2 hosts"
+
+        # Remove master from hosts list to avoid xapi calls failure
+        hosts = list(sr.pool.hosts)
+        hosts.remove(sr.pool.master)
+        # Evacuate the node to be deleted
+        try:
+            random_host = random.choice(hosts) # TBD: Choose Linstor Diskfull node
+            logging.info("Working on %s", random_host.hostname_or_ip)
+            random_host.ssh(['echo', 'c', '>', '/proc/sysrq-trigger'])
+        except Exception as e:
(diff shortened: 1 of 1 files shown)


tests/storage/linstor/test_linstor_sr.py
stormi, line 1: If the test fails while the SR is in a failed state, could this leave the pool in a bad state?

rushikeshjadhav: If the test fails and SR goes bad, I don't see an easy way to recover from it. There could be scenario based troubleshooting required before cleaning up. However, principally single host failure in XOSTOR should be tolerable and this test should catch in case failures.

We probably should use nested tests for this so that in case the pool goes bad, we can just wipe and start with a new clean one.

ydirson: Well if we use those block devices for nothing else than this test, we can easily blank them to restart from scratch.  We do that for all local SRs, and in some way a Linstor SR is "local to the pool".  If our test pool is the sole user, it looks throw-away to me.

tests/storage/linstor/test_linstor_sr.py
```diff
@@ -133,6 +134,47 @@ def test_linstor_missing(self, linstor_sr, host):
...
+
+        # Ensure that VM is able to start on all hosts except failed one
+        for h in sr.pool.hosts:
+            logging.info("Checking VM on host %s", h.hostname_or_ip)
+            if h.hostname_or_ip != random_host.hostname_or_ip:
+                vm.start(on=h.uuid)
+                vm.wait_for_os_booted()
+                vm.shutdown(verify=True)
+
+        # Wait for radom_host to come online
```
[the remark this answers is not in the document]
rushikeshjadhav: @glehmann ^^ does it address?

tests/storage/linstor/test_linstor_sr.py
```diff
@@ -133,6 +134,47 @@ def test_linstor_missing(self, linstor_sr, host):
...
+    @pytest.mark.reboot
+    @pytest.mark.small_vm
+    def test_linstor_sr_fail_host(self, linstor_sr, host, vm_on_linstor_sr):
+        """
+        Fail non master host from the same pool Linstor SR.
+        Ensure that VM is able to boot and shutdown on all hosts.
+        """
+        sr = linstor_sr
+        vm = vm_on_linstor_sr
+        # Ensure that its a single host pool and not multi host pool
```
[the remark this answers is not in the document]
stormi: @rushikeshjadhav does this still need to be addressed?

Wescoeur:
@Lankou66 Since you need to rebase, could you base your work on Erwan's code to standardize the helpers (HA two nodes)? For instance, regarding `/proc/sysrq-trigger`? Thanks.

We might even need to wait until the HA PR has been fully reviewed first.

PR #314, opened 2025-05-23
Add TLS_Test & XOA deployment

LucienLassalle:
This pull request introduces two new test suites for verifying the functionality and security of the XCP-ng environment, along with a minor update to the test paths in `jobs.py`. The most important changes include adding tests to access on some website, ensuring TLS protocol configurations are correct.

### New Test Suites

* [`tests/misc/test_access_links.py`]: Added tests to verify the accessibility of some website (Especially the XOA deployment script) using both `curl` and `wget`. These tests ensure the script responds correctly by comparing the checksum between the local & remote

* [`tests/security/test_tls.py`]: Added tests to validate TLS protocol configurations on the XCP-ng host. Tests ensure that insecure protocols (`tls1` and `tls1_1`) are disabled and that `tls1_2` is enabled.

### DIFF
--- a/tests/misc/test_access_links.py
+++ b/tests/misc/test_access_links.py
@@ -1,0 +1,47 @@
+import pytest
+import subprocess
+import hashlib
+
+# This test is designed to verify the accessibility of the XOA deployment script
+#
+# Requirements:
+# - an XCP-ng host with xcp-ng-release >= 8.3.0.29
+
+@pytest.mark.parametrize("command_id", ["curl", "wget"])
+@pytest.mark.parametrize("url_id", [
+    "xoa",
+    "xcp-ng",
+    "vates"
+])
+def test_access_links(host, command_id, url_id):
+    """
+    Verifies that the specified URL responds correctly via the specified command
+    and compares the checksum of the downloaded content between local and remote.
+    """
+    command = {"curl": "curl -fsSL",
+               "wget": "wget -qO-"}[command_id]
+    url = {
+        "xoa": "https://xoa.io/deploy",
+        "xcpng": "https://updates.xcp-ng.org/trace",
+        "vates": "https://repo.vates.tech/README.txt"
+    }[url_id]
+    COMMAND = f"{command} '{url}'"
+
+    # Download from remote host
+    remote_result = host.ssh(COMMAND)
+
+    # Verify the download worked by comparing with local download
+    # This ensures the content is accessible and identical from both locations
+    local_result = host.local_cmd(COMMAND)
+
+    assert local_result.returncode == 0, (
+        f"Failed to fetch URL locally: {local_result.stderr}"
+    )
+
+    # Extract checksums
+    local_checksum = hashlib.sha256(local_result.stdout.split()[0].encode('utf-8')).hexdigest()
+    remote_checksum = hashlib.sha256(remote_result.split()[0].encode('utf-8')).hexdigest()
+
+    assert local_checksum == remote_checksum, (
+        f"Checksum mismatch: local ({local_checksum}) != remote ({remote_checksum})"
+    )
--- a/tests/security/test_tls.py
+++ b/tests/security/test_tls.py
@@ -1,0 +1,55 @@
+import pytest
+import ssl
+import socket
+import logging
+
+# This test is designed to verify that TLS connections is secured
+#
+# Requirements:
+# - An XCP-ng host
+
+@pytest.mark.parametrize("protocol_name", ["TLSv1", "TLSv1.1"])
+def test_tls_disabled(host: str, protocol_name: str):
+    """
+    Verifies that specified TLS protocols are disabled on the XCP-ng host.
+    Uses the ssl library directly. Should raise SSLError.
+    """
+    PORT = 443
+
+    protocol = {
+        "TLSv1": ssl.PROTOCOL_TLSv1,
+        "TLSv1.1": ssl.PROTOCOL_TLSv1_1
+    }[protocol_name]
+
+    logging.info(f"Testing if protocol {protocol_name} is disabled on host {host}")
+
+    with pytest.raises(ssl.SSLError):
+        context = ssl.SSLContext(protocol)
+        with socket.create_connection((str(host), PORT), timeout=10) as sock:
+            with context.wrap_socket(sock, server_hostname=str(host)) as ssock:
+                ssock.do_handshake()
+                # If we reach this point, the protocol is enabled (test should fail)
+                pytest.fail(f"Protocol {protocol} should be disabled but connection succeeded")
+
+@pytest.mark.parametrize("protocol_name", ["TLSv1.2"])
+def test_enabled(host: str, protocol_name: str):
+    """
+    Verifies that TLSv1.2 is enabled on the XCP-ng host.
+    Uses the ssl library directly.
+    """
+    PORT = 443
+
+    protocol = {
+        "TLSv1.2": ssl.PROTOCOL_TLSv1_2
+    }[protocol_name]
+
+    logging.info(f"Testing if protocol {protocol_name} is enabled on host {host}")
+
+    try:
+        context = ssl.SSLContext(protocol)
+        with socket.create_connection((str(host), PORT), timeout=10) as sock:
+            with context.wrap_socket(sock, server_hostname=str(host)) as ssock:
+                ssock.do_handshake()
+                assert ssock.version()
+    except ssl.SSLError as e:
+        pytest.fail(f"{protocol_name} should be enabled, but got SSLError: {e}")
--- a/jobs.py
+++ b/jobs.py
@@ -23,6 +23,7 @@ JOBS = {
         },
         "paths": [
             "tests/misc",
+            "tests/security",
             "tests/migration",
             "tests/network",
             "tests/snapshot",


tests/security/test_tls.py
```diff
@@ -0,0 +1,34 @@
...
+# - An XCP-ng host
+
+@pytest.mark.parametrize("protocol", ["tls1", "tls1_1"])
+def test_tls_disabled(host, protocol):
+    """
+    Verifies that specified TLS protocols are disabled on the XCP-ng host.
+    Uses the openssl command via SSH. (Should raise an error)
+    """
+    try:
+        host.ssh(f"openssl s_client -connect localhost:443 -{protocol}")
```
stormi, line 16: does connecting to localhost test the same thing as connecting to the management IP address?

stormi: Why run openssl from the host? We could run it from the test runner and see what the host answers to outside requests. I think it's acceptable to add openssl to the list of requirements for the test runner. 

We can also likely testing using Python's openssl bindings or client implementation, directly from the test code.

tests/security/test_tls.py
```diff
@@ -0,0 +1,34 @@
...
+    except SSHCommandFailed as e:
```
stormi, line 17: See `pytest.raises`

stormi (changes requested):

tests/misc/test_xoa_deploy.py
```diff
@@ -0,0 +1,20 @@
...
+    """
+    result = host.ssh("curl -fsSL https://xoa.io/deploy")
+    assert result.startswith("#!/bin/bash"), "The XOA deployment theoretically impossible. (Issue with curl?)"
+
+def test_xoa_deploy_with_wget(host):
+    """
+    Verifies that the URL https://xoa.io/deploy responds correctly via wget.
+    """
+    result = host.ssh("wget -qO- https://xoa.io/deploy")
+    assert result.startswith("#!/bin/bash"), "The XOA deployment theoretically impossible. (Issue with wget?)"
```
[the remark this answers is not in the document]
ydirson: What other URLs are you thinking about?

tests/misc/test_xoa_deploy.py
```diff
@@ -0,0 +1,20 @@
+import pytest
+
+# This test is designed to verify the accessibility of the XOA deployment script
+#
+# Requirements:
+# - an XCP-ng host with xcp-ng-release >= 8.3.0.29
+
+def test_xoa_deploy_with_curl(host):
```
ydirson, line 8: to avoid copypaste, see `pytest.mark.parametrize`

ydirson (changes requested):

tests/misc/test_xoa_deploy.py
stormi: parametrize + test code that changes based on parameter sounds like not the best use of pytest's parametrization. I'd use helper functions instead.

tests/misc/test_xoa_deploy.py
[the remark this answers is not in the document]
LucienLassalle: I don't really understand. Currently, some tests also use curl to retrieve information, etc. Grouping them together seems odd to me.

And the idea behind this test is to verify that XOA is deployable on XCP-ng using either curl or wget (hence the shebang is used to verify that this is feasible).

(And perhaps there will be other tests here to actually ensure that xoa is deployable (for now, this is only theoretical)

tests/security/test_tls.py
LucienLassalle: it should be better now

tests/misc/test_xoa_deploy.py
LucienLassalle: I used parametrize, but I can change if you want

tests/misc/test_xoa_deploy.py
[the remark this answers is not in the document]
stormi: curl's test suite doesn't check that with the current accepted ciphers, important URLs work

At the very least we should try `updates.xcp-ng.org` and `repo.vates.tech`, as we download updates from there (not directly using curl, but via yum using libcurl, but I think it's good to also check that downloading directly from curl works). 

Checking the shebang seems unnecessary. You don't need to check the contents of the file, you need to make sure it was downloaded without an error. We're not testing that the deploy works, here, anyway. We just check that we can download the script. If we wanted to check the deployment, we would have to actually deploy the XOA VM and perform a few checks on it.

That's why I suggest requalifying the test as "let's check that curl can successfully download from a chosen set of URLs (including the XOA deploy script).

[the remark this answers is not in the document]
stormi: yum update doesn't read curlrc, to my knowledge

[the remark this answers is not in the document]
LucienLassalle: And I agree with you, but then what's the point of testing this? If we're already doing a yum update and the urls are only for a yum update?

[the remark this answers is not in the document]
stormi: The point is testing that curl remains able to discuss with important web sites, whenever we update its configuration in the future.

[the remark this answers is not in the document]
LucienLassalle: Okay, I'll make the changes :)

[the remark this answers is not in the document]
ydirson: > Checking the shebang seems unnecessary. You don't need to check the contents of the file, you need to make sure it was downloaded without an error.

Checking it was downloaded without error includes that the correct data was downloaded.  I already saw successful downloads which end up being HTML content - those would be caught by checking the shebang.  But we might want a more thorough test, like verifying a checksum.

tests/misc/test_access_links.py
```diff
@@ -0,0 +1,45 @@
...
+# Requirements:
+# - an XCP-ng host with xcp-ng-release >= 8.3.0.29
+
+@pytest.mark.parametrize("command", ["curl -fsSL", "wget -qO-"])
+@pytest.mark.parametrize("url", [
+    "https://xoa.io/deploy",
+    "https://updates.xcp-ng.org/trace",
+    "https://repo.vates.tech/README.txt",
+    "https://repository.moosefs.com/RPM-GPG-KEY-MooseFS",
+    "http://repository.moosefs.com/MooseFS-3-el7.repo",
```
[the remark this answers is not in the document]
LucienLassalle: No worries, anyway, we can modify the tests in the future according to our needs, note that the links I indicated are present in the storage part, so they will be tested anyway. But later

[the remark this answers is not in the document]
stormi: > note that the links I indicated are present in the storage part, so they will be tested anyway

Maybe not for long, as we don't know any user using this kind of SR. Next time there's an issue with the website, they'll probably go.

tests/misc/test_access_links.py
```diff
@@ -0,0 +1,43 @@
...
+
+    assert local_result.returncode == 0, (
+        f"Failed to fetch URL locally: {local_result.stderr}"
+    )
+
+    # Extract checksums
+    local_checksum = hashlib.sha256(local_result.stdout.split()[0].encode('utf-8')).hexdigest()
+    remote_checksum = hashlib.sha256(remote_result.split()[0].encode('utf-8')).hexdigest()
+
+    assert local_checksum == remote_checksum, (
```
ydirson, line 41: I was thinking about a more conventional "get checksums file from server" - but well here we don't have any so I guess that's about as much as we can test.

I'd still recommend to:
- first do the download that's the subject of the test
- then group the check, of which the local download is part, and label the whole part with an explanation of what we're doing

tests/security/test_tls.py
```diff
@@ -0,0 +1,44 @@
...
+
+@pytest.mark.parametrize("protocol", ["tls1", "tls1_1"])
+def test_tls_disabled(host, protocol):
+    """
+    Verifies that specified TLS protocols are disabled on the XCP-ng host.
+    Uses the openssl command via subprocess. Should raise CalledProcessError.
+    """
+    PORT = 443
+
+    with pytest.raises(subprocess.CalledProcessError) as execInfo:
```
ydirson, line 17: `execInfo` breaks the convention of using snake-case for vars

tests/security/test_tls.py
```diff
@@ -0,0 +1,44 @@
...
+        subprocess.run(
+            ["openssl", "s_client", "-connect", f"{host}:{PORT}", f"-{protocol}"],
+            capture_output=True,
+            text=True,
+            check=True
+        )
+    assert (
+        "no protocols available" in execInfo.value.stderr
+    ), f"{protocol} should be disabled, but got: {execInfo.value.stderr}"
```
ydirson, line 26: Are we OK with ignoring stdout here?  Is stderr giving sufficient info when such a protocol is accepted?

tests/misc/test_access_links.py
```diff
@@ -0,0 +1,43 @@
...
+#
+# Requirements:
+# - an XCP-ng host with xcp-ng-release >= 8.3.0.29
+
+@pytest.mark.parametrize("command", ["curl -fsSL", "wget -qO-"])
+@pytest.mark.parametrize("url", [
+    "https://xoa.io/deploy",
+    "https://updates.xcp-ng.org/trace",
+    "https://repo.vates.tech/README.txt"
+])
```
ydirson, line 15: Using long strings directly as parameters gives funky test names:
```
$ pytest tests/misc/test_access_links.py --co -q
tests/misc/test_access_links.py::test_xoa_deploy[https://xoa.io/deploy-curl -fsSL]
tests/misc/test_access_links.py::test_xoa_deploy[https://xoa.io/deploy-wget -qO-]
tests/misc/test_access_links.py::test_xoa_deploy[https://updates.xcp-ng.org/trace-curl -fsSL]
tests/misc/test_access_links.py::test_xoa_deploy[https://updates.xcp-ng.org/trace-wget -qO-]
tests/misc/test_access_links.py::test_xoa_deploy[https://repo.vates.tech/README.txt-curl -fsSL]
tests/misc/test_access_links.py::test_xoa_deploy[https://repo.vates.tech/README.txt-wget -qO-]
```

Something like this will be more palatable:
```python
@pytest.mark.parametrize("command_id", ("curl", wget"))
...
def test_xoa_deploy(host, command_id, url_id):
    command = {"curl": "curl -fsSL",
               "wget": "wget -qO-"}[command_id]
    url = {...}[url_id]
...
```

ydirson (changes requested):

tests/misc/test_access_links.py
```diff
@@ -0,0 +1,43 @@
...
+    COMMAND = f"{command} '{url}'"
+
+    # Download
+    local_result = subprocess.run(
+        COMMAND,
+        shell=True,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        text=True,
+    )
```
ydirson, line 30: We have `lib.commands.local_cmd()` to wrap `subprocess` and ensure the command we run appears in the log

tests/security/test_tls.py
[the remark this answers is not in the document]
LucienLassalle: Fixed :)

tests/misc/test_access_links.py
LucienLassalle: Thanks, I update it

tests/security/test_tls.py
```diff
@@ -0,0 +1,46 @@
...
+import socket
+import logging
+
+# This test is designed to verify that TLS connections is secured
+#
+# Requirements:
+# - An XCP-ng host
+
+@pytest.mark.parametrize("protocol", [ssl.PROTOCOL_TLSv1, ssl.PROTOCOL_TLSv1_1])
+def test_tls_disabled(host: str, protocol):
```
ydirson, line 12: those constants are numeric, that makes both test naming and logging hard to understand:

```
tests/security/test_tls.py::test_tls_disabled[3] 
...
May 28 13:34:11.502 INFO Testing if protocol 3 is disabled on host 10.1.15.40
```

tests/misc/test_access_links.py
```diff
@@ -0,0 +1,47 @@
...
+
+# This test is designed to verify the accessibility of the XOA deployment script
+#
+# Requirements:
+# - an XCP-ng host with xcp-ng-release >= 8.3.0.29
+
+@pytest.mark.parametrize("command_id", ["curl", "wget"])
+@pytest.mark.parametrize("url_id", [
+    "xoa",
+    "xcp-ng",
```
ydirson, line 13: `-` is a bit special, being used as separator between parameters when naming the test, compare:
```
tests/misc/test_access_links.py::test_access_links[xoa-wget]
tests/misc/test_access_links.py::test_access_links[xcp-ng-curl]
```

Let's avoid this when possible :)

ydirson (changes requested):
Other than those last points, lgtm

stormi (approved):
I'm leaving the rest of the review to Yann, with an early approval. Thanks Lucas.

ydirson (approved):

PR #319, opened 2025-06-11
Installer features prerequisites

ydirson:
This PR collects all the small improvements that ended up in #311 and #316, and are independent enough to be useful to merge before those get settled.


stormi (approved):
LGTM. Before merging we likely need to run at least some testsuites to avoid surprise unwanted failures in CI.

PR #320, opened 2025-06-11
Fix systemd test with xostor

AnthoineB:
When xostor is activated, drbd-reactor.service pulls polkit dependency.
As polkit.service doesn't exist, the command `systemd-analyze verify default.target` prints the message:
`Cannot add dependency job for unit polkit.service, ignoring: Unit not found.`
and `test_systemd.py` fails.

Whitelist polkit message to allow `test_systemd.py` to pass.

### DIFF
--- a/tests/system/test_systemd.py
+++ b/tests/system/test_systemd.py
@@ -28,7 +28,7 @@ def test_verify_default_target(host: Host) -> None:
     polkit_msg = "Cannot add dependency job for unit polkit.service, ignoring: Unit not found."
     for line in analyse.splitlines():
         if line == polkit_msg:
-            pytest.xfail(f"drbd-reactor package should be fix to remove dep to polkit: {polkit_msg}")
+            pytest.xfail(f"drbd-reactor package must be fixed to remove dep to polkit: {polkit_msg}")
         if line not in white_list_issues:
             logging.error(f"{line}")
             err = True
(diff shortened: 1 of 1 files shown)


stormi (commented):
I would rather make the test XFAIL when this string is found, and in the same time queue a work item for a fix.

rzr (commented):
This suggestion should be trivial, but the proposed patch looks valid since the error whitelist was previously added.

May we create a task to remove it and turn all those workarounds into failures (if possible)?

I looks like upstream introduced the dep to polkit and dbus in the service as an example:

https://github.com/LINBIT/drbd-reactor/commit/f0742a80174266cf983be897650c4057b8721bfa

while not installed we have polkit rules packages

https://github.com/xcp-ng-rpms/systemd/blob/master/SPECS/systemd.spec#L1519

tests/system/test_systemd.py
```diff
@@ -25,7 +25,10 @@ def test_failed_units(host: Host) -> None:
 def test_verify_default_target(host: Host) -> None:
     analyse = host.ssh('systemd-analyze verify default.target')
     err = False
+    polkit_msg = "Cannot add dependency job for unit polkit.service, ignoring: Unit not found."
     for line in analyse.splitlines():
+        if line == polkit_msg:
+            pytest.xfail(f"drbd-reactor package should be fix to remove dep to polkit: {polkit_msg}")
```
stormi:
```suggestion
            pytest.xfail(f"drbd-reactor package must be fixed to remove dep to polkit: {polkit_msg}")
```

glehmann (approved):

stormi (approved):

PR #321, opened 2025-06-12
setup uv

glehmann:
install `uv` once with one of these commands:

  pip install uv
  pipx install uv
  curl -LsSf https://astral.sh/uv/install.sh | sh

then just run

~~~sh
  uv run pytest tests/misc/test_basic_without_ssh.py --hosts=gln-xcpng
~~~

to install all the required python dependencies and run the selected
tests.

If you prefer not to use `uv run`, you can tell uv to install the
dependencies and then set up the shell environment to have them
accessible in the shell:

~~~sh
  uv sync
  source .venv/bin/activate
~~~

Run `deactivate` to deactivate the virtual environment.


.python-version
```diff
@@ -0,0 +1 @@
+3.8.20
```
ydirson, line 1: Don't we just want `3.8` instead?

Note for others: this version is used by `uv venv`, but not by `uv sync`

pyproject.toml
```diff
@@ -24,6 +24,7 @@ dev = [
     "packaging>=20.7",
     "pluggy>=1.1.0",
     "pycodestyle>=2.6.0",
+    "pydocstyle>=6.3.0",
```
ydirson, line 27: could be squashed with mypy/pyright addition

ydirson (changes requested):

.python-version
ydirson: > > Don't we just want `3.8` instead?
> 
> You mean that the strict python version could be defined in the lock file? I haven't checked that.

It does not seem to be set there (that puzzles me a bit)

> > Note for others: this version is used by `uv venv`, but not by `uv sync`
> 
> It is used by both

I mean that changing/removing that file and running `uv sync` does not change the interpreter in the venv, whereas `uv venv` does.

glehmann:
> > Note that the pydocstyle test is failing because of the usage of `match` in `xcpng-fs-diff.py`, which was only introduced in Python 3.10.
> 
> wow, how come this was not flagged earlier by the CI running 3.8? 🤔

The job installs python without specifying a version:

~~~yaml
    - name: Set up Python
      uses: actions/setup-python@v2
~~~

.python-version
ydirson: Maybe I drew conclusions a bit too fast, but nevertheless:

```
(tests) user@work-vates:tests (gln/uv-setup-tests-uzqs =)$ uv run python --version
Python 3.8.20
(tests) user@work-vates:tests (gln/uv-setup-tests-uzqs =)$ rm .python-version 
(tests) user@work-vates:tests (gln/uv-setup-tests-uzqs =)$ uv sync
Resolved 53 packages in 0.89ms
Audited 38 packages in 0.03ms
(tests) user@work-vates:tests (gln/uv-setup-tests-uzqs =)$ uv run python --version
Python 3.8.20

(tests) user@work-vates:tests (gln/uv-setup-tests-uzqs =)$ uv venv
Using CPython 3.10.18
Creating virtual environment at: .venv
Activate with: source .venv/bin/activate
(tests) user@work-vates:tests (gln/uv-setup-tests-uzqs =)$ uv run python --version
Installed 38 packages in 263ms
Python 3.10.18
```

glehmann:
see also https://git.vates.tech/jenkins-ci/ci-xcp-ng/pulls/5

README.md
stormi, line 1: Could we also keep a possible setup that doesn't imply using a venv? On a recent enough system, `pip` was largerly  enough and completely transparent. I'd like us to keep offering such an option for those who prefer to use system packages, or user-level packages rather than venv-level ones. Maybe there's simply an option of `uv sync` to install the packages globally for the user?

glehmann: I've updated the documentation with how to work with a specific python version and how to work without uv at all.

glehmann: And there is also a tip about direnv

.python-version
glehmann: We want to pin a version, in order to create an environment as reproducible as possible.
Note that it doesn't prevent us to use another python version, with or without uv, as now documented in the readme

ydirson: The question is, what extra level of reproducibility do you expect from pinning to maintenance releases?  It's not like a compiler where we care about bit-by-bit reproducibility of build artifacts, or like a checker that constantly adds more checks. What problems could arise from using the latest maintenance release of a supported Python version?  Here we clearly add more work work on our side just to get the latest fixes when they are out (and we won't want to do extensive non-ref testing anyway when we do so): to me it just looks like the extra burden is not worth the gains.

README.md
```diff
@@ -3,25 +3,127 @@
...
-Install the python requirements using pip:
+Install `uv` with one of these commands
 
-```
-$ pip install -r requirements/base.txt
-```
+~~~sh
+pip install uv
+pipx install uv
+curl -LsSf https://astral.sh/uv/install.sh | sh
```
ydirson, line 19: Some more guided info could help
```suggestion
pip install uv                                      # for old distros and Fedora-based ones
pipx install uv                                     # for newer distros
curl -LsSf https://astral.sh/uv/install.sh | sh     # if you prefer an installer modifying your environment
```

xcp_ng_tests.egg-info/SOURCES.txt
```diff
@@ -0,0 +1,31 @@
...
+lib/vif.py
+lib/vm.py
+lib/xo.py
+requirements/base.txt
+requirements/dev.txt
+xcp_ng_tests.egg-info/PKG-INFO
+xcp_ng_tests.egg-info/SOURCES.txt
+xcp_ng_tests.egg-info/dependency_links.txt
+xcp_ng_tests.egg-info/requires.txt
+xcp_ng_tests.egg-info/top_level.txt
```
ydirson: the removal for this was likely squashed into the wrong commit

ydirson (changes requested):

PR #327, opened 2025-06-16
host_with_perf: Set to 8.3 hosts only

tescande:
Make perf_probe test to run on 8.3 hosts only.


tests/packages/linux_image/conftest.py
```diff
@@ -6,7 +6,7 @@
 from pkgfixtures import host_with_saved_yum_state
 
 @pytest.fixture(scope="package")
-def host_with_perf(host_with_saved_yum_state):
+def host_with_perf(host_with_saved_yum_state, host_at_least_8_3):
```
stormi, line 9: TIP: invert the fixtures and if no other constraints reorder fixture initialization, this should verify the version before attempting to save yum state.

PR #328, opened 2025-06-16
a-posteriori review and fixes of Linstor tests

ydirson:
First version is only driven by the `TestLinstorSRCreateDestroy` class of tests.  I propose fixes for the problems I see with obvious solutions (and some changes that are merely improvements), and insert FIXME's in the places I see as potentially problematic.

### DIFF
--- a/tests/storage/linstor/conftest.py
+++ b/tests/storage/linstor/conftest.py
@@ -43,16 +43,29 @@ def lvm_disks(host, sr_disks_for_all_hosts, provisioning_type):
 
 @pytest.fixture(scope="package")
 def storage_pool_name(provisioning_type):
+    # FIXME: this needs an explanation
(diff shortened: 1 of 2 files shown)


tests/storage/linstor/conftest.py
```diff
@@ -43,17 +43,30 @@ def lvm_disks(host, sr_disks_for_all_hosts, provisioning_type):
 
 @pytest.fixture(scope="package")
 def storage_pool_name(provisioning_type):
+    # FIXME: this needs an explanation
     return GROUP_NAME if provisioning_type == "thick" else STORAGE_POOL_NAME
 
+# FIXME why having this feature of session scope?  Shouldn't it make
+# it impossible to run both thin and thick tests in the same session?
```
rushikeshjadhav, line 50: Since all linstor tests were going to be repeated with just one change of SR type, pytest `params` technique was used.

PR #330, opened 2025-06-18
Disk fixtures rework

ydirson:
This replaces the existing disk fixtures, which rely on the assumption that we can take a dom0 device name on the pool master, and expect to find it across the whole pool with the same meaning.  This was causing issues for GlusterFS and Linstor tests when some hosts in the test pool get their disk device names swapped after a reboot.

The old mechanism used a mandatory `--sr_disk` flag to control the behavior of various `sr_disk*` fixtures, with little code-sharing.  The new one is based on a central `disks` fixture to properly introspect the pool's disks, which comes with its own `--disks` flag to filter which disks to expose to the tests, but by default exposes all of them.  It then uses additional fixtures to filter disks using different criteria (for now: availability and blocksize of the disk), and then some pool-level fixtures allowing testing of distributed SR without risking to (try to) use the wrong disk.

Builds on:
- #323

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -129,7 +129,6 @@ def pytest_collection_modifyitems(items, config):
         'windows_vm',
         'hostA2',
         'hostB1',
-        'sr_disk',
         'sr_disk_4k',
         'unused_512B_disks',
     ]
@@ -403,30 +402,6 @@ def unused_512B_disks(disks: dict[Host, list[Host.BlockDeviceInfo]]
            }
     logging.debug("available disks collected: %s", {host.hostname_or_ip: value for host, value in ret.items()})
     return ret
-
-@pytest.fixture(scope='session')
-def sr_disk(pytestconfig, host: Host) -> Generator[DiskDevName]:
-    """
-    Disk DEVICE NAME available on FIRST POOL MASTER.
-
-    Abort if not exactly one --sr_disk.  If --sr_disk=auto take any, else
-    return requested device (abort if not present).
-    """
-    disks = pytestconfig.getoption("sr_disk")
-    if len(disks) != 1:
-        pytest.fail("This test requires exactly one --sr-disk parameter")
-    disk = disks[0]
-    if disk == "auto":
-        logging.info(">> Check for the presence of a free disk device on the master host")
-        disks = host.available_disks()
-        assert len(disks) > 0, "a free disk device is required on the master host"
-        disk = disks[0]
-        logging.info(f">> Found free disk device(s) on hostA1: {' '.join(disks)}. Using {disk}.")
-    else:
-        logging.info(f">> Check that disk or block device {disk} is available on the master host")
-        assert disk in host.available_disks(), \
-            f"disk or block device {disk} is either not present or already used on master host"
-    yield disk
 
 @pytest.fixture(scope='session')
 def sr_disk_4k(pytestconfig, host: Host) -> Generator[DiskDevName]:
--- a/conftest.py
+++ b/conftest.py
@@ -15,6 +15,7 @@ from lib import pxe
 from lib.common import (
     callable_marker,
     DiskDevName,
+    HostAddress,
     is_uuid,
     prefix_object_name,
     setup_formatted_and_mounted_disk,
@@ -35,7 +36,7 @@ from lib.xo import xo_cli
 # need to import them in the global conftest.py so that they are recognized as fixtures.
 from pkgfixtures import formatted_and_mounted_ext4_disk, sr_disk_wiped
 
-from typing import Dict, Generator
+from typing import Dict, Generator, Iterable
 
 # Do we cache VMs?
 try:
@@ -82,6 +83,14 @@ def pytest_addoption(parser):
         action="store",
         default=20,
         help="Max lines to output in a ssh log (0 if no limit)"
+    )
+    parser.addoption(
+        "--disks",
+        action="append",
+        default=[],
+        help="HOST:DISKS to authorize for use by tests. "
+             "DISKS is a possibly-empty comma-separated list. "
+             "No mention of a given host authorizes use of all its disks."
     )
     parser.addoption(
         "--sr-disk",
@@ -229,6 +238,14 @@ def hosts(pytestconfig) -> Generator[list[Host]]:
     cleanup_hosts()
 
 @pytest.fixture(scope='session')
+def pools_hosts_by_name_or_ip(hosts: list[Host]) -> Generator[dict[HostAddress, Host]]:
+    """All hosts of all pools, each indexed by their hostname_or_ip."""
+    yield {host.hostname_or_ip: host
+           for pool_master in hosts
+           for host in pool_master.pool.hosts
+           }
+
+@pytest.fixture(scope='session')
 def registered_xo_cli():
     # The fixture is not responsible for establishing the connection.
     # We just check that xo-cli is currently registered
@@ -340,11 +357,38 @@ def local_sr_on_hostB1(hostB1):
     yield sr
 
 @pytest.fixture(scope='session')
-def disks(hosts: list[Host]) -> dict[Host, list[Host.BlockDeviceInfo]]:
+def disks(pytestconfig, pools_hosts_by_name_or_ip: dict[HostAddress, Host]
+          ) -> dict[Host, list[Host.BlockDeviceInfo]]:
     """Dict identifying names of all disks for on all hosts of first pool."""
-    ret = {host: host.disks()
-           for pool_master in hosts
-           for host in pool_master.pool.hosts
+    def _parse_disk_option(option_text: str) -> tuple[HostAddress, list[DiskDevName]]:
+        parsed = option_text.split(sep=":", maxsplit=1)
+        assert len(parsed) == 2, f"--disks option {option_text!r} is not <host>:<disk>[,<disk>]*"
+        host_address, disks_string = parsed
+        devices = disks_string.split(',') if disks_string else []
+        return host_address, devices
+
+    cli_disks = dict(_parse_disk_option(option_text)
+                     for option_text in pytestconfig.getoption("disks"))
+
+    def _host_disks(host: Host, hosts_cli_disks: list[DiskDevName] | None) -> Iterable[Host.BlockDeviceInfo]:
+        """Filter host disks according to list from `--cli` if given."""
+        host_disks = host.disks()
+        # no disk specified = allow all
+        if hosts_cli_disks is None:
+            yield from host_disks
+            return
+        # check all disks in --disks=host:... exist
+        for cli_disk in hosts_cli_disks:
+            for disk in host_disks:
+                if disk['name'] == cli_disk:
+                    yield disk
+                    break # names are unique, don't expect another one
+            else:
+                raise Exception(f"no {cli_disk!r} disk on host {host.hostname_or_ip}, "
+                                f"has {','.join(disk['name'] for disk in host_disks)}")
+
+    ret = {host: list(_host_disks(host, cli_disks.get(host.hostname_or_ip)))
+           for host in pools_hosts_by_name_or_ip.values()
            }
     logging.debug("disks collected: %s", {host.hostname_or_ip: value for host, value in ret.items()})
     return ret
--- a/conftest.py
+++ b/conftest.py
@@ -121,7 +121,8 @@ def pytest_collection_modifyitems(items, config):
         'hostA2',
         'hostB1',
         'sr_disk',
-        'sr_disk_4k'
+        'sr_disk_4k',
+        'unused_512B_disks',
     ]
 
     for item in items:
@@ -339,6 +340,27 @@ def local_sr_on_hostB1(hostB1):
     yield sr
 
 @pytest.fixture(scope='session')
+def disks(hosts: list[Host]) -> dict[Host, list[Host.BlockDeviceInfo]]:
+    """Dict identifying names of all disks for on all hosts of first pool."""
+    ret = {host: host.disks()
+           for pool_master in hosts
+           for host in pool_master.pool.hosts
+           }
+    logging.debug("disks collected: %s", {host.hostname_or_ip: value for host, value in ret.items()})
+    return ret
+
+@pytest.fixture(scope='session')
+def unused_512B_disks(disks: dict[Host, list[Host.BlockDeviceInfo]]
+                      ) -> dict[Host, list[Host.BlockDeviceInfo]]:
+    """Dict identifying names of all 512-bytes-blocks disks for on all hosts of first pool."""
+    ret = {host: [disk for disk in host_disks
+                  if disk["log-sec"] == "512" and host.disk_is_available(disk["name"])]
+           for host, host_disks in disks.items()
+           }
+    logging.debug("available disks collected: %s", {host.hostname_or_ip: value for host, value in ret.items()})
+    return ret
+
+@pytest.fixture(scope='session')
 def sr_disk(pytestconfig, host: Host) -> Generator[DiskDevName]:
     """
     Disk DEVICE NAME available on FIRST POOL MASTER.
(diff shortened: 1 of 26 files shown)


ydirson:
Last push:
- adds a first `--disk` implementation
- fixes `unused_disks` ignoring any filtering done in `disks`
- extends a bit the use of the type aliases
- notes in WIP that `pool_with_unused_disk` only applies to pool A, we'll need
  - to make that more explicit
  - something better for linstor pool-join

Nambrok (changes requested):

conftest.py
```diff
@@ -149,7 +161,7 @@ def pytest_runtest_makereport(item, call):
 # fixtures
 
 @pytest.fixture(scope='session')
-def hosts(pytestconfig):
+def hosts(pytestconfig) -> Generator[list[Host], None, None]:
```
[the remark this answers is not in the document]
ydirson: done

conftest.py
```diff
@@ -327,103 +347,60 @@ def local_sr_on_hostB1(hostB1):
...
-            f"disk or block device {disk} was not found to be present and free on all hosts"
-        logging.info(f">> Disk or block device {disk} is present and free on all pool members")
-    yield candidates[0]
+def unused_4k_disks(unused_disks: dict[Host, list[DiskDevName]]
+                    ) -> Generator[dict[Host, list[DiskDevName]]]:
+    ret = {host: [disk for disk in host_disks if host.disk_is_4k(disk)]
+           for host, host_disks in unused_disks.items()
+           }
+    logging.debug("available 4k disks collected: %s", {host.hostname_or_ip: value for host, value in ret.items()})
+    yield ret
```
ydirson, line 396: There is one thing here to be noted: unlike `sr_disk` which ended up only getting access to 512B-blocks disks, `unused_disks` gets all disks regardless of blocksize, and `unused_4k_disks` is just an additional filter on top of the former.
This means that in theory a test that would need both a standard disk and a 4k one could see the additional one get allocated the 4k disk of the host, and when requesting a 4k one there would be none left, if the fixtures are specified in that order.  Specifying them in the "most specific to less specific order" naturally works around that issue.

Do we want to do better?  Like making sure `unused_disks` picks 512B disks first? I think it might not be worth the added complexity, but do I miss some cases?

conftest.py
```diff
@@ -215,6 +227,14 @@ def cleanup_hosts():
 
     cleanup_hosts()
 
+@pytest.fixture(scope='session')
+def pools_hosts_by_name_or_ip(hosts: list[Host]) -> Generator[dict[HostAddress, Host]]:
+    """All hosts of all pools, each indexed by their hostname_or_ip."""
+    yield {host.hostname_or_ip: host
+           for pool_master in hosts
+           for host in pool_master.pool.hosts
+           }
```
Wescoeur, line 236: Waiting for a code formatter to have this :angel::
```python
    yield {
        host.hostname_or_ip: host
        for pool_master in hosts
        for host in pool_master.pool.hosts
    }
```

lib/host.py
```diff
@@ -534,43 +536,53 @@ def management_pif(self):
...
-        blk_output = self.ssh(['lsblk', '-nd', '-I', '8,259', '--output', 'NAME,LOG-SEC']).splitlines()
-        for line in blk_output:
-            line = line.split()
+        blk_output = self.ssh(['lsblk', '--noheadings', '--nodeps',
+                               '-I', '8,259', # limit to: sd, blkext
+                               '--output', 'NAME,LOG-SEC']).splitlines()
+        for line_str in blk_output:
+            line = line_str.split()
             disk = line[0]
             sec_size = line[1]
```
Wescoeur, line 587:
```suggestion
        for line in blk_output:
            disk, sec_size = line.split()
```

conftest.py
ydirson: :thinking: what would it bring, aside from wasted vertical space?

Wescoeur: Well IMHO:
- This fixes the alignment asymmetry of `{` and `}`
- Fix the wasted horizontal space :wink: 
- Fix the indentation of the code, not a multiple of 4

conftest.py
```diff
@@ -327,103 +347,60 @@ def local_sr_on_hostB1(hostB1):
...
+        parsed = option_text.split(sep=":", maxsplit=1)
+        assert len(parsed) == 2, f"--disks option {option_text!r} is not <host>:<disk>[,<disk>]*"
+        host_address, disks_string = parsed
+        devices = disks_string.split(',') if disks_string else []
+        return host_address, devices
+
+    cli_disks = dict(_parse_disk_option(option_text)
+                     for option_text in pytestconfig.getoption("disks"))
+
+    def _host_disks(host: Host, hosts_cli_disks: list[DiskDevName] | None) -> Iterable[DiskDevName]:
```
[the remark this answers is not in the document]
Nambrok: I think this function should be named more sensibly, `_compare_cli_disks_to_host_disks` or `_filter_host_disks_to_cli_disks`

conftest.py
Nambrok: The previous `sr_disk` gave only 512B disks because a 4KiB in ext SR would fail.

conftest.py
ydirson: > I think explicit for loops would be more readable here

IIRC they would be less efficient

conftest.py
Nambrok: The existing SR would work for 4KiB only with QCOW2 images, trying to create a VHD would still fail.

lib/host.py
ydirson: Separate optimization, but yes. Not strictly equivalent if `lsblk` ever breaks the assumption, but should be safe.

conftest.py
ydirson: OK, looks like a case for having linked parameters, as in
```
(("512B", "vhd"),
 ("512B", "qcow"),
 ("4k", "qcow")
)
```

Those would parametrize `vdi_on_*_sr` and the tests explicitly creating a VDI or importing a VM, we should be able to pass the proper parameter values to the SR-creation fixture.

ydirson: This makes this change and #335 somewhat linked

conftest.py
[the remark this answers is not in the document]
ydirson: > I like to use `Optional` in this kind of case to carry the meaning that the parameter should be None but it would mean needing to import it

There are several issues with `Optional`:
- the name clashes with the "optional parameters" of a func (ie. those with default values), and that would not be that much of a problem if a defaulted value was always `None`, which is not the case.  So yes, in some particular cases `X | None` is really "an optional X", but for all non-`None` default values `Optional[X]` is just wrong.  So banning it could even be an option to preserve sanity now that we have `| None` (and indeed I thought I had read that it was being deprecated, though I did not find any evidence for this).
- here the function parameter is not not even optional, we should really pass either a list or a plain `None`

lib/host.py
```diff
@@ -534,43 +536,53 @@ def management_pif(self):
         uuid = self.xe('pif-list', {'management': True, 'host-uuid': self.uuid}, minimal=True)
         return pif.PIF(uuid, self)
 
-    def disks(self):
+    def disks(self) -> list[DiskDevName]:
         """ List of SCSI disks, e.g ['sda', 'sdb', 'nvme0n1']. """
```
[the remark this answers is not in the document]
ydirson: fixed

lib/host.py
ydirson: applied

conftest.py
ydirson: > OK, looks like a case for having linked parameters, as in
> 
> ```
> (("512B", "vhd"),
>  ("512B", "qcow"),
>  ("4k", "qcow")
> )
> ```
> 
> Those would parametrize `vdi_on_*_sr` and the tests explicitly creating a VDI or importing a VM, we should be able to pass the proper parameter values to the SR-creation fixture.

I do have a prototype working, based on this fixture in toplevel `conftest.py`:
```python
@pytest.fixture(params=(512, 4096), scope='session')
def disk_blocksize(request) -> Generator[int]:
    return request.param
```

 overridable in narrower scopes with a subset of the values, eg. for `largeblock/conftest.py`:
```python
@pytest.fixture(params=(4096,), scope='session')
def disk_blocksize(request) -> Generator[int]:
    return request.param
```

**however** making this a test parameter, rather than a restriction on disk selection, has a nasty side-effect: we wouldn't be able to use both 512B and 4k SRs in a single test, preventing e.g. migration tests between the 2 kinds of disk.  Thus I'm withdrawing this part of the proposal, and will get something closer to what we have today.

ydirson: > Ah, then maybe we should introduce `unused_512B_disks` to get things right and make this visible.

Went this way

> But at some point those SR currently limited to 512B will be able to use larger blocksize, and we'll want to test them on both 512B and 4k. Maybe this should indeed be anticipated and use blocksize as a test parameter already, so it will be trivial to add those new tests when time comes?

At that point we should be able to introduce a `blocksize` param on the test, and pull either `unused_512B_disks` or `unused_4k_disks` depending on the param value. @rushikeshjadhav I guess

tests/storage/linstor/conftest.py
```diff
@@ -12,11 +12,21 @@
...
+    DEVICE NAME exists ACROSS THE WHOLE POOL. Then make a VG out of
+    all those, then a LV taking up the whole VG space.
+
+    Return the list of device node paths for that list of devices
+    used in all hosts.
+    """
+    hosts = pool_with_unused_512B_disk.hosts
 
     for host in hosts:
+        devices = [f"/dev/{disk}" for disk in unused_512B_disks[host][0:1]]
```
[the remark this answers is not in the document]
ydirson: Limiting the number of disks used as PV to 1.  The docstring should be updated at the same time since the logic is changing

conftest.py
```diff
@@ -327,103 +347,66 @@ def local_sr_on_hostB1(hostB1):
...
-    else:
-        logging.info(f">> Check that 4KiB block device {disk} is available on the master host")
-        assert disk in host.available_disks(4096), \
-            f"4KiB block device {disk} must be available for use on master host"
-    yield disk
+def unused_512B_disks(disks: dict[Host, list[DiskDevName]]
+                      ) -> Generator[dict[Host, list[DiskDevName]]]:
+    """Dict identifying names of all 512-bytes-blocks disks for on all hosts of first pool."""
+    ret = {host: [disk for disk in host_disks
+                  if host.disk_is_available(disk) and not host.disk_is_4k(disk)]
```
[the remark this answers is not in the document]
Nambrok: That's what was there before, you gave a blocksize as parameter with a default parameter being 512

conftest.py
```diff
@@ -71,6 +75,14 @@ def pytest_addoption(parser):
         default=20,
         help="Max lines to output in a ssh log (0 if no limit)"
     )
+    parser.addoption(
+        "--disks",
+        action="append",
+        default=[],
+        help="HOST:DISKS to authorize for use by tests. "
+             "DISKS is a possibly-empty comma-separated list. "
+             "No mention of a given host authorizes use of all its disks."
```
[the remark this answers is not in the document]
ydirson: Not really, the `--disks` flag acts on purpose on the `disks` feature, then naturally all fixtures filtering further based on such criteria as "unused" inherits from this, and benefits from this filtering.
We may want in the future to have a fixture to use a disk already sporting a SR, and the same single `--disks` flag will impact it as well.

ydirson:
... and rebased

lib/common.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
```
glehmann, line 1: We no longer need that now that we require Python 3.11

conftest.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
```
glehmann, line 1: We no longer need that now that we require Python 3.11

lib/host.py
```diff
@@ -52,6 +53,18 @@ def host_data(hostname_or_ip):
 class Host:
     xe_prefix = "host"
 
+    # Must be kept in sync. Data extraction is automatic, no conversion from str is done.
+    BLOCK_DEVICES_FIELDS = "NAME,KNAME,PKNAME,SIZE,LOG-SEC,TYPE"
```
glehmann, line 58: maybe replace the explicit list with
```python
','.join(k.upper() for k in BlockDeviceInfo.__annotations__.keys())
```
?

glehmann (approved):

lib/host.py
ydirson: Good point, I was too lazy to dig that line at the time, thx :)

conftest.py
ydirson: That's not strictly true, see https://docs.python.org/3/library/__future__.html#id2

Such annotation is also useful to avoid extra runtime imports that are really only needed for typechecking (without resorting to enclosing the type inside quotes).

Additionally, for this particular one, the annotation seems to be what allows to omit the trailing `, None, None` for the `Generator`:
```
user@work-vates:tests (disks-rework >)$ uv run pytest --hosts=10.1.132.1 tests/storage/ext/test_ext_sr.py --co -q
ImportError while loading conftest '/home/user/src/xcpng/tests/conftest.py'.
conftest.py:171: in <module>
    def hosts(pytestconfig) -> Generator[list[Host]]:
                               ^^^^^^^^^^^^^^^^^^^^^
../../../.local/share/uv/python/cpython-3.11.11-linux-x86_64-gnu/lib/python3.11/typing.py:379: in inner
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
../../../.local/share/uv/python/cpython-3.11.11-linux-x86_64-gnu/lib/python3.11/typing.py:1592: in __getitem__
    _check_generic(self, params, self._nparams)
../../../.local/share/uv/python/cpython-3.11.11-linux-x86_64-gnu/lib/python3.11/typing.py:293: in _check_generic
    raise TypeError(f"Too {'many' if alen > elen else 'few'} arguments for {cls};"
E   TypeError: Too few arguments for typing.Generator; actual 1, expected 3
```

glehmann (approved):

ydirson:
Last push squashes fixups from fixing issues discovered during CI runs.

Wescoeur (approved):

conftest.py
```diff
@@ -335,103 +341,71 @@ def local_sr_on_hostB1(hostB1):
...
-        assert len(disks) > 0, "a free 4KiB block device is required on the master host"
-        disk = disks[0]
-        logging.info(f">> Found free 4KiB block device(s) on hostA1: {' '.join(disks)}. Using {disk}.")
-    else:
-        logging.info(f">> Check that 4KiB block device {disk} is available on the master host")
-        assert disk in host.available_disks(4096), \
-            f"4KiB block device {disk} must be available for use on master host"
-    yield disk
+def unused_512B_disks(disks: dict[Host, list[Host.BlockDeviceInfo]]
+                      ) -> Generator[dict[Host, list[Host.BlockDeviceInfo]]]:
```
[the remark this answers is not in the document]
ydirson: fixed

conftest.py
```diff
@@ -335,103 +341,71 @@ def local_sr_on_hostB1(hostB1):
...
-        assert len(candidates) > 0, \
-            f"a free disk device is required on all pool members. Pool master has: {' '.join(master_disks)}."
-        logging.info(f">> Found free disk device(s) on all pool hosts: {' '.join(candidates)}. Using {candidates[0]}.")
-    else:
-        assert len(candidates) > 0, \
-            f"disk or block device {disk} was not found to be present and free on all hosts"
-        logging.info(f">> Disk or block device {disk} is present and free on all pool members")
-    yield candidates[0]
+def unused_4k_disks(disks: dict[Host, list[Host.BlockDeviceInfo]]
+                    ) -> Generator[dict[Host, list[Host.BlockDeviceInfo]]]:
```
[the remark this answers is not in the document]
ydirson: fixed

lib/host.py
```diff
@@ -546,21 +563,32 @@ def management_pif(self):
...
-
-    def raw_disk_is_available(self, disk: str) -> bool:
+    def rescan_block_devices_info(self) -> None:
         """
-        Check if a raw disk (without any identifiable filesystem or partition label) is available.
+        Initalize static informations about the disks.
 
-        It suggests the disk is "raw" and likely unformatted thus available.
+        Despite those being static, it can be necessary to rescan,
+        when we test how XCP-ng reacts to changes of hardware (or
```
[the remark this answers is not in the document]
ydirson: added

Nambrok (approved):

tests/storage/largeblock/test_largeblock_sr.py
```diff
@@ -16,9 +23,10 @@ class TestLARGEBLOCKSRCreateDestroy:
     def test_create_sr_with_missing_device(self, host):
         try_to_create_sr_with_missing_device('largeblock', 'LARGEBLOCK-local-SR-test', host)
 
-    def test_create_and_destroy_sr(self, host, sr_disk_4k):
+    def test_create_and_destroy_sr(self, host: Host, unused_4k_disks: dict[Host, list[Host.BlockDeviceInfo]]) -> None:
         # Create and destroy tested in the same test to leave the host as unchanged as possible
-        sr = host.sr_create('largeblock', "LARGEBLOCK-local-SR-test", {'device': '/dev/' + sr_disk_4k}, verify=True)
+        sr_disk = unused_4k_disks[host][0]["name"]
```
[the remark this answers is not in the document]
ydirson: Not sure I get what you mean, but that sounds like an improvement that can be done later?

PR #333, opened 2025-06-30
feat(image_format): support image_format

Nambrok:
vdi.py: add getting the image_format if available
vdi.py: add get parent from smconfig if available
sr.py: add choosing a image_format on vdi creation sr.py add get_type to get the type of the SR

### DIFF
--- a/lib/sr.py
+++ b/lib/sr.py
@@ -1,5 +1,6 @@
 import logging
 import time
+from typing import Optional
 
 import lib.commands as commands
 
@@ -12,6 +13,7 @@ class SR:
         self.pool = pool
         self._is_shared = None # cached value for is_shared()
         self._main_host = None # cached value for main_host()
+        self._type = None # cache value for get_type()
 
     def pbd_uuids(self):
         return safe_split(self.pool.master.xe('pbd-list', {'sr-uuid': self.uuid}, minimal=True))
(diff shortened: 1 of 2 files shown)


lib/sr.py
```diff
@@ -153,13 +155,21 @@ def is_shared(self):
                                                             {'uuid': self.uuid, 'param-name': 'shared'}))
         return self._is_shared
 
-    def create_vdi(self, name_label, virtual_size=64):
+    def get_type(self) -> str:
+        if self._type is None:
+            self._type = self.pool.master.xe("sr-param-get", {"uuid": self.uuid, "param-name": "type"})
+        return self._type
+
+    def create_vdi(self, name_label: str, virtual_size: int = 64, image_format: Optional[str] = None) -> VDI:
```
Wescoeur, line 163:
```suggestion
    def create_vdi(self, name_label: str, virtual_size: int = 64, image_format: str = '') -> VDI:
```

lib/sr.py
```diff
@@ -1,5 +1,6 @@
 import logging
 import time
+from typing import Optional
```
Wescoeur, line 3: Not required if you think that my next comment is correct.

Wescoeur (changes requested):

PR #324, opened 2025-06-13
feat(tests): add disk performance benchmark tests

Millefeuille42:

PR #334, opened 2025-06-30
lib/host: Avoid using identical source and destination paths in execute_script scp

rushikeshjadhav:
Using the same path for both source and destination may cause scp to fail, especially when the source path (e.g. a temp path on macOS) does not exist on the remote system. Thus, resorting to use `/tmp/` and random name from `sctipt.name` on destination.


lib/host.py
```diff
@@ -201,13 +201,13 @@ def execute_script(self, script_contents, shebang='sh', simple_output=True):
             script.write('#!/usr/bin/env ' + shebang + '\n')
             script.write(script_contents)
             script.flush()
-            self.scp(script.name, script.name)
+            self.scp(script.name, "/tmp/" + script.name.split('/')[-1])
```
ydirson, line 204: better use `os.path.basneame`

ydirson (changes requested):

lib/host.py
rushikeshjadhav: thanks, made the change.

stormi: Ping @ydirson

ydirson: I like this fixture :)

Millefeuille42: Straight up borrowed from https://github.com/xcp-ng/xcp-ng-tests/pull/324 😁
It could be great to have this kind of generic fixtures available for all tests 👀 (eventually platform agnostic unlike the one I suggested)

rushikeshjadhav: Thanks, please see the latest update if it works.

Millefeuille42: it's all good for me

lib/host.py
```diff
@@ -231,13 +231,15 @@ def execute_script(self, script_contents, shebang='sh', simple_output=True):
             script.write('#!/usr/bin/env ' + shebang + '\n')
             script.write(script_contents)
             script.flush()
-            self.scp(script.name, script.name)
+            remote_path = self.ssh("mktemp").strip()
+            self.scp(script.name, remote_path)
+            self.ssh(['chmod', '0755', remote_path])
```
Wescoeur, line 236: Should probably be moved in try scope in case for some reason a call fails even if unlikely.

Wescoeur (changes requested):

lib/host.py
rushikeshjadhav: Sure, have moved them.

glehmann (approved):

PR #336, opened 2025-07-02
tests/storage/linstor: Added `daemon-reload` post package installation

rushikeshjadhav:
Sometimes the `systemctl` may complain about service reload.

```
SM: [64335] linstor-manager:prepare_sr error: Failed to stop drbd-reactor:  Failed to stop drbd-reactor.service: Unit drbd-reactor.service not loaded.
```

### DIFF
--- a/tests/storage/linstor/conftest.py
+++ b/tests/storage/linstor/conftest.py
@@ -70,6 +70,8 @@ def pool_with_linstor(hostA2, lvm_disks, pool_with_saved_yum_state):
         # Needed because the linstor driver is not in the xapi sm-plugins list
         # before installing the LINSTOR packages.
         host.ssh(["systemctl", "restart", "multipathd"])
+        # Needed because sometimes the systemctl may complain about service reload.
+        host.ssh(["systemctl", "daemon-reload"])
         host.restart_toolstack(verify=True)
 
     with concurrent.futures.ThreadPoolExecutor() as executor:


tests/storage/linstor/conftest.py
```diff
@@ -70,6 +70,8 @@ def install_linstor(host):
         # Needed because the linstor driver is not in the xapi sm-plugins list
         # before installing the LINSTOR packages.
         host.ssh(["systemctl", "restart", "multipathd"])
+        # Needed because sometimes the systemctl may complain about service reload.
+        host.ssh("systemctl", "daemon-reload")
```
Wescoeur, line 74:
```suggestion
        host.ssh(["systemctl", "daemon-reload"])
```

rushikeshjadhav: right, changed to list.

gduperrey (approved):
8.3 complete linstor :white_check_mark:

rushikeshjadhav:
> > > As I said on mattermost, the proper fix will have to be done in packaging. RPMs that add or remove services must reload systemd. I would like to avoid such workarounds that will remain who knows how much time in the test suite. CC @ydirson for Platform&Release team.
> > 
> > 
> > Can we link this ~issue~ PR to the specific RPM issue and then undo this once RPM is fixed?
> 
> The proper fix is probably not that complicated. If we add a workaround, we lose an incentive to fix it properly. Let's make sure we fix it shortly, instead.

Done, will close this one.

PR #337, opened 2025-07-02
tests/storage/linstor: Add retry logic to _get_diskful_hosts to handle transient failures

rushikeshjadhav:
Added a retry loop with up to 3 attempts and 5-second delay in _get_diskful_hosts to handle transient SSHCommandFailed exceptions during 'linstor resource list' query.

### DIFF
--- a/tests/storage/linstor/test_linstor_sr.py
+++ b/tests/storage/linstor/test_linstor_sr.py
@@ -136,17 +136,28 @@ class TestLinstorSR:
 # --- Test diskless resources --------------------------------------------------
 
 def _get_diskful_hosts(host, controller_option, volume_name):
-    # Find host where volume is diskless
-    # | {volume_name} | {host} | 7017 | Unused | Ok    |   UpToDate | 2023-10-24 18:52:05 |
-    lines = host.ssh([
-        "linstor", controller_option, "resource", "list",
-        "|", "grep", volume_name, "|", "grep", "UpToDate"
-    ]).splitlines()
-    diskfuls = []
-    for line in lines:
-        hostname = line.split('|')[2].strip()
-        diskfuls += hostname
-    return diskfuls
+    # TODO: If any resource is in a temporary creation state or unknown, then need to wait intelligently.
+    attempt = 0
+    retries = 3
+    sleep_sec = 5
+    while attempt < retries:
+        try:
+            # Find host where volume is UpToDate
+            # | {volume_name} | {host} | 7017 | Unused | Ok    |   UpToDate | 2023-10-24 18:52:05 |
+            lines = host.ssh([
+                "linstor", controller_option, "resource", "list",
+                "|", "grep", volume_name, "|", "grep", "UpToDate"
+            ]).splitlines()
+            diskfuls = []
+            for line in lines:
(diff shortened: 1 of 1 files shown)


tests/storage/linstor/test_linstor_sr.py
```diff
@@ -136,17 +136,27 @@ def test_linstor_missing(self, linstor_sr, host):
...
+                "linstor", controller_option, "resource", "list",
+                "|", "grep", volume_name, "|", "grep", "UpToDate"
+            ]).splitlines()
+            diskfuls = []
+            for line in lines:
+                hostname = line.split('|')[2].strip()
+                diskfuls += hostname
+            return diskfuls
+        except SSHCommandFailed:
+            attempt += 1
```
Wescoeur, line 156: I think it's not entirely ok in this case: is it acceptable to return only one diskful when there is, for example, another replica with a temporary creation status? Should we wait for a status change to potentially return two diskfuls?

rushikeshjadhav: That can be a TODO.

In a situation where some node(s) might have rebooted, there can be status of `unknown` too, so perhaps more awaiting to change. Since this function was specific to return `UpToDate`, I think it's ok to retry SSH and return currently available `UpToDate`.

We can look at writing more mature function that can preempt certain status and wait intelligently. Do you prefer to do it now or after this is through?

Nambrok: @rushikeshjadhav UP, can you add the todo? I would like to have this merged.

gduperrey:
We needed this change to run the `8.3 main-linstor`. We were therefore able to validate it to a copy of this branch on the main repository (`dtt-linstorretryuptodate`).

This allowed us to successfully run it after four failed runs.

We were also able to run the `8.3 complete linstor` on this copy of this branch :)

Wescoeur (approved):

tests/storage/linstor/test_linstor_sr.py
```diff
@@ -136,17 +136,28 @@ def test_linstor_missing(self, linstor_sr, host):
...
+        if attempt >= retries:
+            raise
```
Nambrok, line 159: There is a little problem here, the empty raise isn't in a except, meaning it will error with `RuntimeError: No active exception to reraise` instead of re-raising the error.

tests/storage/linstor/test_linstor_sr.py
```diff
@@ -136,17 +136,28 @@ def test_linstor_missing(self, linstor_sr, host):
...
+                "linstor", controller_option, "resource", "list",
+                "|", "grep", volume_name, "|", "grep", "UpToDate"
+            ]).splitlines()
+            diskfuls = []
+            for line in lines:
+                hostname = line.split('|')[2].strip()
+                diskfuls += hostname
+            return diskfuls
+        except SSHCommandFailed:
+            attempt += 1
```
Nambrok, line 157: Could you also add a logging of the exception with the attempt number?

rushikeshjadhav: Done

PR #340, opened 2025-07-10
Implement fistpoint for smapi in tests

Nambrok:
This PR implement the usage of fistpoint for testing some errors in smapi.
A fistpoint is an action that you can enable in the smapi to for example add a sleep at some point or raise an exception.
For example:
```py
        with FistPoint(vm.host, "blktap_activate_inject_failure"):
            try:
                vm.start()
                vm.shutdown(force=True)
            except Exception as e:
                logging.info(f"Expected failure {e}")
```
Activating the fistpoint `blktap_activate_inject_failure` mean that the VDI activation will fail.
This fistpoint always raise an exception but most fistpoint just add a sleep in the code.
Using the fixture `exit_on_fistpoint` make all fistpoint raise an exception instead by enabling a special fistpoint called `fist_LVHDRT_exit`.

### DIFF
--- a/lib/fistpoint.py
+++ b/lib/fistpoint.py
@@ -1,5 +1,6 @@
 import logging
 
+from lib.commands import SSHCommandFailed
 from lib.host import Host
 
 from typing import Final
@@ -58,7 +59,11 @@ class FistPoint:
 
     def disable(self):
         logging.info(f"Disabling fistpoint {self.fistpointName}")
-        self.host.ssh(["rm", self._get_path(self.fistpointName)])
+        try:
+            self.host.ssh(["rm", self._get_path(self.fistpointName)])
+        except SSHCommandFailed as e:
+            logging.info(f"Failed trying to disable fistpoint {self._get_path(self.fistpointName)} with error {e}")
+            raise
 
     def __enter__(self):
         self.enable()


lib/fistpoint.py
```diff
@@ -0,0 +1,48 @@
+import logging
+from typing import Final
+
+from lib.host import Host
+
+FISTPOINT_DIR: Final = "/tmp"
+LVHDRT_EXIT_FIST: Final = "fist_LVHDRT_exit"
+
+class FistPoint:
+    fistpointName: str
```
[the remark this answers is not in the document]
Nambrok: Done :)

lib/vdi.py
```diff
@@ -47,6 +47,13 @@ def clone(self):
     def readonly(self) -> bool:
         return strtobool(self.param_get("read-only"))
 
+    def size(self) -> int:
+        return int(self.param_get("virtual-size"))
```
Wescoeur, line 51: I prefer that we be more explicit about this kind of method, `virtual_size` seems better to me. Just to differentiate it from `physical_size` which corresponds to the size used on physical storage and `virtual_size` which corresponds to the choice of the maximum usable size by the user.

lib/vdi.py
```diff
@@ -47,6 +47,13 @@ def clone(self):
     def readonly(self) -> bool:
```
Wescoeur, line 47:
```suggestion
    def is_readonly(self) -> bool:
```

IMHO This seems more logical when I read it: we're asking about the value of a parameter. I would tend to say that we should have `get_name` rather than `name` on this class too, but that's not a change added by this PR.

Wescoeur (changes requested):

PR #342, opened 2025-07-25
Import configuration enhancements from python-project-template

glehmann:


.github/workflows/jobs-check.yml
```diff
@@ -2,18 +2,22 @@ name: Check jobs consistency
...
 jobs:
   jobs-check:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
+      with:
+        persist-credentials: false
     - name: Install uv
-      uses: astral-sh/setup-uv@v6
+      uses: astral-sh/setup-uv@bd01e18f51369d5a26f1651c3cb451d3417e3bba  # v6.3.1
```
psafont, line 15: Are all these different workflow files needed?

I'd rather a single file that reuses a single step that installs uv and its cache, and have different jobs that run. Having this many files means it will be very easy to miss updating some of the versions of the actions

glehmann: Those files were there before me, so I can't say why this way to organize the jobs was chosen.

I'm not much bothered by the update: I use `ambr actions/checkout@v2  actions/checkout@v4` to update something and be sure to not miss anything. But I sure think that's quite a lot of duplication.

We could also define a custom action that checks out the repository and install uv.
Something like (not tested):
```yaml
# .github/actions/setup/action.yml
name: setup
description: 'Checkout the repository and install uv'
runs:
  using: 'composite'
  steps:
      - uses: actions/checkout@v4
        with:
          persist-credentials: false
      - uses: astral-sh/setup-uv@bd01e18f51369d5a26f1651c3cb451d3417e3bba  # v6.3.1
        with:
          version: "0.7.x"
```

and use it in the jobs:

```yaml
jobs:
  flake8:
    runs-on: ubuntu-latest
    steps:
      - uses: ./.github/actions/setup/
      - run: uv sync --frozen
      - run: uv run flake8
```

ydirson (approved):
lgtm, but I it would be great to see how this SARIF data gets presented, the URL in the bot's comment just does not show anything for me

PR #345, opened 2025-09-05
Add toolstack restart after teardown on tests enabling / disabling SRs

Millefeuille42:
This prevents the risk of discrepancy in `sm-list` between multiple hosts 
which can cause a "incompatible set of sm features" error.

### DIFF
--- a/pkgfixtures.py
+++ b/pkgfixtures.py
@@ -38,11 +38,36 @@ def formatted_and_mounted_ext4_disk(host: Host, unused_512B_disks: dict[Host, li
     yield mountpoint
     teardown_formatted_and_mounted_disk(host, mountpoint)
 
-@pytest.fixture(scope='package')
-def host_with_saved_yum_state(host: Host) -> Generator[Host]:
+def _host_with_saved_yum_state(host: Host, restart_toolstack: bool) -> Generator[Host]:
+    """
+    Saves the yum state and restores the saved state on teardown.
+
+    It also optionally restarts the toolstack.
+    Fixtures using this function should not be used concurrently in the same test run.
+    """
     host.yum_save_state()
     yield host
     host.yum_restore_saved_state()
+    if restart_toolstack:
+        host.restart_toolstack(verify=True)
+
+@pytest.fixture(scope='package')
+def host_with_saved_yum_state(host: Host) -> Generator[Host]:
+    """
+    Saves the yum state and then restore it on teardown.
+
+    Should not be used concurrently with another "host_with_saved_yum_state" fixture
+    """
+    yield from _host_with_saved_yum_state(host, False)
+
+@pytest.fixture(scope='package')
+def host_with_saved_yum_state_toolstack_restart(host: Host) -> Generator[Host]:
+    """
+    Saves the yum state then restore it and restarts the toolstack on teardown.
+
+    Should not be used concurrently with another "host_with_saved_yum_state" fixture
+    """
+    yield from _host_with_saved_yum_state(host, True)
 
 @pytest.fixture(scope='package')
 def pool_with_saved_yum_state(host: Host) -> Generator[Pool]:
--- a/tests/storage/fsp/conftest.py
+++ b/tests/storage/fsp/conftest.py
@@ -1,7 +1,7 @@
 import pytest
 
 # Explicitly import package-scoped fixtures (see explanation in pkgfixtures.py)
-from pkgfixtures import host_with_saved_yum_state
+from pkgfixtures import host_with_saved_yum_state_toolstack_restart
 
 FSP_REPO_NAME = 'runx'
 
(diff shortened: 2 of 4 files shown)


pkgfixtures.py
stormi, line 1: I find this context manager approach interesting... But on second thought maybe it's too smart and not the best solution. You are solving a fixture problem by bypassing part of the fixture behaviour.

How about just creating a fixture that does what you need?

```
def host_with_saved_yum_state_toolstack_restart(host):
    host.yum_save_state()
    yield host
    host.restart_toolstack(verify=True)
```

Millefeuille42: I made a context manager to avoid adding a fixture that does almost the same thing to avoid confusion. However, the fixture does not carry the same risk of misusage as the context manager.

I can change it to a fixture if it is preferred.

tests/storage/fsp/conftest.py
```diff
@@ -1,7 +1,7 @@
 import pytest
 
 # Explicitly import package-scoped fixtures (see explanation in pkgfixtures.py)
-from pkgfixtures import host_with_saved_yum_state
+from pkgfixtures import __host_with_saved_yum_state_cm
```
ydirson, line 4: `pkgfixtures` would not be the place for a cm.  If it is meant to be shared, then it should go into `lib/`, and not start with underscores.

tests/storage/fsp/conftest.py
```diff
@@ -10,12 +10,13 @@
...
-    host.add_xcpng_repo(FSP_REPO_NAME, 'vates')
-    yield host
+def host_with_runx_repo(host):
+    with __host_with_saved_yum_state_cm as host:
+        host.add_xcpng_repo(FSP_REPO_NAME, 'vates')
+        yield host
     # teardown
-    host.remove_xcpng_repo(FSP_REPO_NAME)
+        host.remove_xcpng_repo(FSP_REPO_NAME)
+    host.restart_toolstack(verify=True)
```
ydirson, line 19: I'm not sure why `host_with_runx_repo` depends on `host_with_saved_yum_state' , as itself does not make a change to the yum state.  That causes this toolstack restart to be strangely placed.

ydirson (changes requested):
Those commends basically suggest as Sam says that the separate-fixture could be better

pkgfixtures.py
Millefeuille42: I replaced the cm with a fixture as you suggested

ydirson (dismissed):

pkgfixtures.py
```diff
@@ -38,11 +38,22 @@ def formatted_and_mounted_ext4_disk(host: Host, unused_512B_disks: dict[Host, li
...
-    yield host
-    host.yum_restore_saved_state()
+def host_with_saved_yum_state_factory(host):
+    def make(restart_toolstack):
+        host.yum_save_state()
+        yield host
+        host.yum_restore_saved_state()
+        if restart_toolstack:
+            host.restart_toolstack(verify=True)
+    return make
```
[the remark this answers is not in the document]
Millefeuille42: Further experimentation lead to the conclusion that we can keep the factory as a fixture and that managing the state inside of the factory won't work properly. So I'd suggest to leave it as @Wescoeur suggested.

pkgfixtures.py
```diff
@@ -38,11 +38,23 @@ def formatted_and_mounted_ext4_disk(host: Host, unused_512B_disks: dict[Host, li
...
+
+@pytest.fixture(scope="package")
+def host_with_saved_yum_state(host_with_saved_yum_state_factory):
+    yield from host_with_saved_yum_state_factory(restart_toolstack=False)
+
+@pytest.fixture(scope="package")
+def host_with_saved_yum_state_toolstack_restart(host_with_saved_yum_state_factory):
+    yield from host_with_saved_yum_state_factory(restart_toolstack=True)
```
stormi, line 57: The documentation of the fixtures needs to describe the fact that we must not use both in the same test run.

Millefeuille42: Added docstrings in [6506003](https://github.com/xcp-ng/xcp-ng-tests/pull/345/commits/650600366d8496f070f395a26bdec4f906328ae8)

Millefeuille42:
Following discussions with @Wescoeur and @Nambrok, I changed the fixture factory to a function called by the two `host_with_saved_yum_state*` fixtures to simplify to reduce complexity compared to a fixture factory.

pkgfixtures.py
```diff
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import pytest
-from typing import TYPE_CHECKING, Generator
+from typing import TYPE_CHECKING, Generator, Callable
```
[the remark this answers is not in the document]
Nambrok: ruff appears to not run on `pkgfixtures.py` but it's also unsorted.

Nambrok (approved):

PR #353, opened 2025-09-26
Add a xen-bugtool smoke test

last-genius:


tests/packages/bugtool/test_bugtool.py
```diff
@@ -0,0 +1,46 @@
...
+# contains some of the expected logs and files.
+#
+# Requirements:
+# - an XCP-ng host
+
+class TestsBugtool:
+    def verify_contains(host, archive, files):
+        listing = host.ssh(['tar', '-jtf', archive])
+        listing = [x.split('/', maxsplit=1)[1] for x in listing.splitlines()]
+        all(file in listing for file in files)
```
stormi, line 15: This doesn't raise, `all` only returns False if not all the conditions are met. I think the tests won't fail if a file is missing.

tests/packages/bugtool/test_bugtool.py
```diff
@@ -0,0 +1,46 @@
...
+import subprocess
+
+# This smoke test runs xen-bugtool and verifies that the archive it generates
+# contains some of the expected logs and files.
+#
+# Requirements:
+# - an XCP-ng host
+
+class TestsBugtool:
+    def verify_contains(host, archive, files):
```
glehmann, line 12: It should be annotated with `@staticmethod` or have `self` as the first parameter.
Moving it out of the class entirely would work too.

glehmann, line 12: It would be nice to add the types on the parameters

```suggestion
    @staticmethod
    def verify_contains(host: 'Host', archive: str, files: list[str]):
```

tests/packages/bugtool/test_bugtool.py
```diff
@@ -0,0 +1,46 @@
...
+                                             "var/log/xensource.log",
+                                             "var/log/SMlog",
+                                             "xapi-db.xml",
+                                             "acpidump.out",
+                                             "etc/fstab",
+                                             "etc/xapi.conf",
+                                             "etc/xensource/pool.conf",
+                                         ])
+        finally:
+            host.ssh(['rm', '-f', '/var/opt/xen/bug-report/*'])
```
glehmann, line 46: maybe just remove `filename`?

glehmann (approved):

PR #355, opened 2025-09-30
lib/vm: Add check for power action features

dinhngtu:
In theory, is_management_agent_up needs to check for all management agent features required by the tests.
This includes the poweroff and reboot features.


stormi (approved):

dinhngtu:
Revised the check to avoid breaking old images per discussions.

PR #356, opened 2025-10-02
Add coalesce integrity tests on most SR types

glehmann:
Not sure if others should be covered as well.
I've removed the coalesce tests based on `dom0`, because they seemed redundant, but we could also keep them. Just let me know.

### DIFF
--- a/lib/vm.py
+++ b/lib/vm.py
@@ -322,7 +322,7 @@ class VM(BaseVM):
 
     def disconnect_vdi(self, vdi: VDI):
         logging.info(f"<< Unplugging VDI {vdi.uuid} from VM {self.uuid}")
-        assert vdi in self.vdis, "VDI {vdi.uuid} not in VM {self.uuid} VDI list"
+        assert vdi in self.vdis, f"VDI {vdi.uuid} not in VM {self.uuid} VDI list"
         vbd_uuid = self.host.xe("vbd-list", {
             "vdi-uuid": vdi.uuid,
             "vm-uuid": self.uuid


lib/sr.py
```diff
@@ -166,7 +166,7 @@ def get_type(self) -> str:
             self._type = self.pool.master.xe("sr-param-get", {"uuid": self.uuid, "param-name": "type"})
         return self._type
 
-    def create_vdi(self, name_label: str, virtual_size: int = 64, image_format: Optional[str] = None) -> VDI:
+    def create_vdi(self, name_label: str, virtual_size: int | str = 64, image_format: Optional[str] = None) -> VDI:
```
Nambrok, line 169: I would prefer to choose between int or string. String make sense with how you want to use it `1GiB`, so we could replace the default value to be a string and check caller to change them to use a string.

tests/storage/storage.py
```diff
@@ -68,7 +80,25 @@ def get_xapi_session():
...
+
+def wait_for_vdi_coalesce(vdi: VDI):
+    # It is necessary to wait a long time because the GC can be paused for more than 5 minutes.
+    # And it is also necessary to allow a sufficiently long merge time which depends on the amount of data.
+    wait_for_not(lambda: vdi.get_parent(), msg="Waiting for coalesce", timeout_secs=7 * 60)
+    logging.info("Coalesce done")
+
+
+def install_randstream(vm: 'VM'):
+    vm.ssh("wget -nv https://github.com/xcp-ng/randstream/releases/download/0.3.1/randstream-0.3.1-x86_64-unknown-linux-musl.tar.gz -O - | tar -xzC /usr/bin/ ./randstream")  # noqa: E501
```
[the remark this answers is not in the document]
glehmann: We could add it to the Alpine image, but it seems simpler this way

tests/storage/ext/test_ext_sr.py
```diff
@@ -7,13 +7,15 @@
 from lib.commands import SSHCommandFailed
 from lib.common import vm_image, wait_for
 from lib.fistpoint import FistPoint
-from lib.vdi import VDI
 from tests.storage import try_to_create_sr_with_missing_device, vdi_is_open
+from tests.storage.storage import install_randstream, operation_on_vdi, wait_for_vdi_coalesce
```
Nambrok, line 11: Could you add them in the `tests/storage/__init__.py` import to avoid the double storage like it's done for `vdi_is_open` above.

tests/storage/ext/test_ext_sr.py
```diff
@@ -66,6 +68,32 @@ def test_snapshot(self, vm_on_ext_sr):
...
+    @pytest.mark.parametrize("vdi_op", ["snapshot", "clone"])
+    def test_coalesce(self, unix_vm: 'VM', vdi_on_ext_sr: 'VDI', vdi_op):
+        vm = unix_vm
+        vdi = vdi_on_ext_sr
+        vm.connect_vdi(vdi, 'xvdb')
+        new_vdi = None
+        try:
+            vm.start()
+            vm.wait_for_vm_running_and_ssh_up()
+            install_randstream(vm)
```
Nambrok, line 81: It would be best to do it in a fixture

tests/storage/ext/test_ext_sr.py
```diff
@@ -66,6 +68,32 @@ def test_snapshot(self, vm_on_ext_sr):
...
+            vm.ssh("randstream generate -v /dev/xvdb")
+            vm.ssh("randstream validate -v --expected-checksum 65280014 /dev/xvdb")
```
Nambrok, line 83: Where are you getting the expected checksum that is hardcoded?
There is a seed on the second generate call a few lines later but not here.

tests/storage/ext/test_ext_sr.py
```diff
@@ -66,6 +68,32 @@ def test_snapshot(self, vm_on_ext_sr):
...
+            new_vdi = operation_on_vdi(vm.host, vdi.uuid, vdi_op)
+            vm.ssh("randstream generate -v --seed 1 --size 128Mi /dev/xvdb")
+            vm.ssh("randstream validate -v --expected-checksum ad2ca9af /dev/xvdb")
+            vm.host.xe("vdi-destroy", {"uuid": new_vdi.uuid})
```
Nambrok, line 87: Can't you just do `new_vdi.destroy()` like you do in the `finally`?

tests/storage/lvm/test_lvm_sr.py
```diff
@@ -7,13 +7,15 @@
 from lib.commands import SSHCommandFailed
 from lib.common import vm_image, wait_for
 from lib.fistpoint import FistPoint
-from lib.vdi import VDI
 from tests.storage import try_to_create_sr_with_missing_device, vdi_is_open
+from tests.storage.storage import install_randstream, operation_on_vdi, wait_for_vdi_coalesce
```
Nambrok, line 11: Same here, adding the import in `__init__.py`

tests/storage/ext/test_ext_sr.py
Nambrok: and attaching a VDI to a running VM if it has the PV drivers is ok

tests/storage/ext/test_ext_sr.py
glehmann: done

tests/storage/ext/conftest.py
```diff
@@ -20,8 +20,8 @@ def ext_sr(host: Host, unused_512B_disks: dict[Host, list[Host.BlockDeviceInfo]]
     sr.destroy()
 
 @pytest.fixture(scope='module')
-def vdi_on_ext_sr(ext_sr):
-    vdi = ext_sr.create_vdi('EXT-local-VDI-test')
+def vdi_on_ext_sr(ext_sr: 'SR'):
+    vdi = ext_sr.create_vdi('EXT-local-VDI-test', virtual_size='1GiB')
```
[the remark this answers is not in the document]
glehmann: It seems ok, I'll `1GiB` the default

lib/sr.py
```diff
@@ -166,7 +166,7 @@ def get_type(self) -> str:
             self._type = self.pool.master.xe("sr-param-get", {"uuid": self.uuid, "param-name": "type"})
         return self._type
 
-    def create_vdi(self, name_label: str, virtual_size: int = 64, image_format: Optional[str] = None) -> VDI:
+    def create_vdi(self, name_label: str, virtual_size: int | str = '1GiB', image_format: Optional[str] = None) -> VDI:
```
Wescoeur, line 169: I'm not a fan of type unions except in special cases. Here, I think we should define an alias because I'm afraid we'll have to specify the union in several places in the tests in the future like resize tests.

For example:
```python
Size = int | str

...

def create_vdi(self, name_label: str, virtual_size: Size = '1GiB', image_format: Optional[str] = None) -> VDI:
```

Otherwise it's fine to use: 2**30 directly.

lib/sr.py
```diff
@@ -166,7 +167,7 @@ def get_type(self) -> str:
             self._type = self.pool.master.xe("sr-param-get", {"uuid": self.uuid, "param-name": "type"})
         return self._type
 
-    def create_vdi(self, name_label: str, virtual_size: int = 64, image_format: Optional[str] = None) -> VDI:
+    def create_vdi(self, name_label: str, virtual_size: int = 1 * GiB, image_format: Optional[str] = None) -> VDI:
```
ydirson: That feels strange at first look, but we were indeed requesting a default VDI size of *64 bytes*, wow.  Maybe mention this in commit message to avoid too many readers taking time to double-check that new value would really do what it's meant to.

tests/storage/__init__.py
```diff
@@ -1,6 +1,9 @@
 from .storage import (
     cold_migration_then_come_back,
+    install_randstream,
     live_storage_migration_then_come_back,
+    operation_on_vdi,
     try_to_create_sr_with_missing_device,
     vdi_is_open,
+    wait_for_vdi_coalesce,
 )
```
ydirson: This remark might feel a bit off-topic, but having to re-export all those funcs seems awkward.
This `tests.storage` module seems inconsistent with the usual practice to put shared code inside `lib`, we might want to move that module instead of having to maintain this `__init__.py`?

tests/storage/storage.py
```diff
@@ -72,3 +82,21 @@ def get_xapi_session():
...
+def wait_for_vdi_coalesce(vdi: VDI):
+    # It is necessary to wait a long time because the GC can be paused for more than 5 minutes.
+    # And it is also necessary to allow a sufficiently long merge time which depends on the amount of data.
+    wait_for_not(lambda: vdi.get_parent(), msg="Waiting for coalesce", timeout_secs=7 * 60)
+    logging.info("Coalesce done")
+
+
+def install_randstream(vm: 'VM'):
+    logging.debug("Installing randstream")
+    vm.ssh("wget -nv https://github.com/xcp-ng/randstream/releases/download/0.3.1/randstream-0.3.1-x86_64-unknown-linux-musl.tar.gz -O - | tar -xzC /usr/bin/ ./randstream")  # noqa: E501
```
ydirson: Doesn't this look like a good candidate for being a fixture?

tests/storage/conftest.py
```diff
@@ -1,7 +1,25 @@
+import pytest
+
+from tests.storage import install_randstream
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from lib.vdi import VDI
+    from lib.vm import VM
```
ydirson: Using `TYPE_CHECKING` here seems to go in the opposite direction from the "add some type annotations" commit (on which I suggested to move the removal of those into a separate commit)

tests/storage/nfs/test_nfs_sr.py
```diff
@@ -55,6 +59,29 @@ def test_snapshot(self, dispatch_nfs):
         finally:
             vm.shutdown(verify=True)
 
+    @pytest.mark.small_vm
+    @pytest.mark.parametrize('dispatch_nfs', ['vdi_on_nfs_sr', 'vdi_on_nfs4_sr'], indirect=True)
+    @pytest.mark.parametrize("vdi_op", ["snapshot", "clone"])
+    def test_coalesce(self, storage_test_vm: VM, dispatch_nfs: VDI, vdi_op):
+        vm = storage_test_vm
+        vdi = dispatch_nfs
+        vm.connect_vdi(vdi, 'xvdb')
```
ydirson: Why a hardcoded `xvdb`, when we have `*disks` fixtures?

lib/vm.py
```diff
@@ -309,11 +309,12 @@ def connect_vdi(self, vdi: VDI, device: str = "autodetect") -> str:
...
-            self.host.xe("vbd-plug", {"uuid": vbd_uuid})
-        except commands.SSHCommandFailed:
-            self.host.xe("vbd-destroy", {"uuid": vbd_uuid})
-            raise
+        if self.is_running():
+            try:
+                self.host.xe("vbd-plug", {"uuid": vbd_uuid})
+            except commands.SSHCommandFailed:
+                self.host.xe("vbd-destroy", {"uuid": vbd_uuid})
+                raise
```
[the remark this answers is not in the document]
glehmann: The PDB is automatically plugged when booting the VDI, so this allows connecting a VDI to a stopped VM.
The behavior remains the same as before for running VMs.

I'm not even sure to understand why we would want to deal with an extra `plug_vbd` parameter here.

We might not need it anymore, I'll check.

tests/storage/nfs/test_nfs_sr.py
ydirson: yes, those ones (and the 4k variant would likely allow to test coalesce on `largeblocks` too?)

lib/vm.py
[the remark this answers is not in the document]
glehmann: We no longer need this change, so I will remove it.
I still believe that this has nothing to do with DWIM, and I would appreciate it if you would justify this kind of statement when you make it.

lib/vdi.py
```diff
@@ -89,3 +93,9 @@ def param_clear(self, param_name):
     def param_remove(self, param_name, key, accept_unknown_key=False):
         _param_remove(self.sr.pool.master, self.xe_prefix, self.uuid,
                       param_name, key, accept_unknown_key)
+
+    def wait_for_coalesce(self):
+        # It is necessary to wait a long time because the GC can be paused for more than 5 minutes.
+        # And it is also necessary to allow a sufficiently long merge time which depends on the amount of data.
+        wait_for_not(lambda: self.get_parent(), msg="Waiting for coalesce", timeout_secs=7 * 60)
```
Nambrok: It should be noted that it will only work if we are on depth=2 (i.e. only one snapshot and we are the leaf).
If we have multiple snapshot, the parent will be the grandparent instead, resulting in a timeout here but the coalesce did happen.

tests/storage/__init__.py
glehmann: Yes, that's unusual. I'll move it in another PR.

lib/vdi.py
```diff
@@ -89,3 +93,9 @@ def param_clear(self, param_name):
     def param_remove(self, param_name, key, accept_unknown_key=False):
         _param_remove(self.sr.pool.master, self.xe_prefix, self.uuid,
                       param_name, key, accept_unknown_key)
+
+    def wait_for_coalesce(self):
+        # It is necessary to wait a long time because the GC can be paused for more than 5 minutes.
```
rzr: does it depends on size ?  can the value be computed ? btw shannon would double the value and set to 10min

PR #357, opened 2025-10-07
Don't install sr tools on failed teardown

klmp200:


klmp200:
> A PR title is important. It's our first contact with the change. Here, it looks like the title currently says the contrary of what the PR actually does :)

Indeed, it does, I made a typo writing it ^^'
It was even contradicting the commit messages which is accurate with it's content :)

That's fixed

I'll add a proper description once I'm done with the content of the PR, it's not ready yet :)

PR #360, opened 2025-10-24
Add tests for corruption recovery on LINSTOR SR

Millefeuille42:
Tests DRBD's corruption recovery using drbdadm verify and invalidate-remote commands as well as basic VM operations (startup, snapshot, shutdown)

### DIFF
--- a/tests/storage/linstor/conftest.py
+++ b/tests/storage/linstor/conftest.py
@@ -3,12 +3,15 @@ from __future__ import annotations
 import pytest
 
 import functools
+import json
 import logging
 import os
+from contextlib import contextmanager
 from dataclasses import dataclass
 
 import lib.commands as commands
 from lib import config
+from lib.common import safe_split
 
 try:
     from data import LINSTOR_REDUNDANCY  # type: ignore
@@ -186,9 +189,25 @@ def vdi_on_linstor_sr(linstor_sr: SR) -> Generator[VDI, None, None]:
     yield vdi
     vdi.destroy()
 
+@contextmanager
+def _vm_on_linstor_sr(host: Host, linstor_sr: SR, vm_ref: str) -> Generator[VM]:
+    """
+    Context manager to provide the fixture lifecycle on a VM on a Linstor SR
+    with different scopes without repeating the code.
+    """
+    vm = host.import_vm(vm_ref, sr_uuid=linstor_sr.uuid)
+    try:
+        yield vm
+    finally:
+        logging.info("<< Destroy VM")
+        vm.destroy(verify=True)
+
 @pytest.fixture(scope='module')
 def vm_on_linstor_sr(host: Host, linstor_sr: SR, vm_ref: str) -> Generator[VM, None, None]:
-    vm = host.import_vm(vm_ref, sr_uuid=linstor_sr.uuid)
-    yield vm
-    logging.info("<< Destroy VM")
-    vm.destroy(verify=True)
+    with _vm_on_linstor_sr(host, linstor_sr, vm_ref) as vm:
+        yield vm
+
+@pytest.fixture(scope='function')
+def vm_on_linstor_sr_function(host: Host, linstor_sr: SR, vm_ref: str) -> Generator[VM]:
+    with _vm_on_linstor_sr(host, linstor_sr, vm_ref) as vm:
+        yield vm
--- a/tests/storage/linstor/test_linstor_sr.py
+++ b/tests/storage/linstor/test_linstor_sr.py
@@ -1,10 +1,12 @@
 import pytest
 
+import json
 import logging
+import shlex
 import time
 
 from lib.commands import SSHCommandFailed
-from lib.common import vm_image, wait_for
+from lib.common import safe_split, vm_image, wait_for
 from lib.host import Host
 from lib.pool import Pool
 from lib.sr import SR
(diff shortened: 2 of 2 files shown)


klmp200 (changes requested):
Otherwise, lgtm

tests/storage/linstor/test_linstor_sr.py
```diff
@@ -52,6 +53,29 @@ def test_create_and_destroy_sr(self, pool_with_linstor, provisioning_type, stora
...
+            conn.get("name", ""),
+            peer.get("out-of-sync", 0),
+        )
+        for res in get_drbd_status(host, resource)
+        for conn in res.get("connections", [])
+        for peer in conn.get("peer_devices", [])
+        if peer.get("out-of-sync", 0) > 0
+    ]
+
+def wait_sync(host, resource):
```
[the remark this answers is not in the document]
Millefeuille42: done in [5af35f6](https://github.com/xcp-ng/xcp-ng-tests/pull/360/commits/5af35f6bfc7f3904d2d5756bb568c728284ca186)

klmp200 (approved):

tests/storage/linstor/test_linstor_sr.py
```diff
@@ -52,6 +57,29 @@ def test_create_and_destroy_sr(self, pool_with_linstor, provisioning_type, stora
...
+            peer.get("out-of-sync", 0),
+        )
+        for res in get_drbd_status(host, resource)
+        for conn in res.get("connections", [])
+        for peer in conn.get("peer_devices", [])
+        if peer.get("out-of-sync", 0) > 0
+    ]
+
+def wait_drbd_sync(host: Host, resource: str):
+    logging.info(f"[{host}] Waiting for DRBD sync on resource `{resource}`...")
```
rzr: This works but pylint dont like fstrings: W1203: Use lazy % formatting in logging functions (logging-fstring-interpolation)

rzr (approved):
few remarks but not my expertise (yet)

glehmann (approved):

tests/storage/linstor/conftest.py
```diff
@@ -136,3 +139,56 @@ def vm_on_linstor_sr(host, linstor_sr, vm_ref):
     yield vm
     logging.info("<< Destroy VM")
     vm.destroy(verify=True)
+
+@pytest.fixture(scope='function')
+def host_and_corrupted_vdi_on_linstor_sr(host, linstor_sr, vm_ref):
+    vm: VM = host.import_vm(vm_ref, sr_uuid=linstor_sr.uuid)
```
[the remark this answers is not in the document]
guzu: Maybe worth a comment ?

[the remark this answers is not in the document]
stormi: This will break when vm_ref is a UUID referring to an existing VM. This also totally bypasses local cache for imported VMs.

If it's best to import a VM and you don't care which VM, that there's no advantage in , then you should import one of the mini VMs without depending on vm_ref, as we already do in some storage tests:
```
vm = host.import_vm(vm_image('mini-linux-x86_64-bios'), sr_uuid=sr.uuid)
```

On the contrary, if you want the test to run with any VM we can test (as part of a "multi" job), then maybe you should still rely on the imported_vm fixture, but then copy it to the SR where you need it. 

And if what you want is actually an empty VM with just a disk that you can break, then maybe it's time we implemented creating a VM from scratch, without relying on an XVA.

Worth a discussion.

stormi (changes requested):

tests/storage/linstor/conftest.py
```diff
@@ -171,3 +175,56 @@ def vm_on_linstor_sr(host: Host, linstor_sr: SR, vm_ref: str):
...
     logging.info("<< Destroy VM")
     vm.destroy(verify=True)
+
+@pytest.fixture(scope='function')
+def host_and_corrupted_vdi_on_linstor_sr(host: Host, linstor_sr: SR, vm_ref: str):
+    vm: VM = host.import_vm(vm_ref, sr_uuid=linstor_sr.uuid)
+    pool: Pool = host.pool
+    master: Host = pool.master
+
+    def get_vdi_volume_name_from_linstor() -> str:
```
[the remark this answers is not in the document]
Millefeuille42: done in [2580566](https://github.com/xcp-ng/xcp-ng-tests/pull/360/commits/2580566672476d654c5a76b91564a88685fdb6d9)

tests/storage/linstor/test_linstor_sr.py
```diff
@@ -54,6 +108,87 @@ def test_snapshot(self, vm_on_linstor_sr: VM) -> None:
         finally:
             vm.shutdown(verify=True)
 
+    @pytest.fixture(scope='function')
+    def host_and_corrupted_vdi_on_linstor_sr(self, host: Host, linstor_sr: SR, vm_on_linstor_sr_function: VM):
```
stormi: Why does the name only mention two of the three yielded objects?

tests/storage/linstor/test_linstor_sr.py
```diff
@@ -54,6 +108,87 @@ def test_snapshot(self, vm_on_linstor_sr: VM) -> None:
...
+                for peer in conn.get("peer_devices", [])
+                if peer.get("peer-disk-state", "") == "UpToDate"
+            )
+            logging.info("Elected `%s` as peer for verification and repair", other_host)
+        except StopIteration:
+            pytest.fail("Could not find an UpToDate peer host")
+
+        corrupted = None
+        max_attempts = 3
+        # Attempting several times since testing revealed `drbdadm verify` can be flaky
```
stormi, line 156: This is likely good enough, but if despite the max_attempts the flakiness reaches CI, even just once, we'll need to investigate the source of the flakiness to robustify the test.

tests/storage/linstor/test_linstor_sr.py
```diff
@@ -54,6 +108,87 @@ def test_snapshot(self, vm_on_linstor_sr: VM) -> None:
...
+            if not corrupted_resources:
+                logging.warning("No corrupted resources found on attempt #%d", attempt)
+                continue
+            for res_name, peer_name, out_of_sync in corrupted_resources:
+                if res_name == resource_name and peer_name == hostname:
+                    corrupted = (res_name, peer_name, out_of_sync)
+            if corrupted:
+                break
+        if not corrupted:
+            pytest.fail(f"Failed to identify corrupted resource after {max_attempts} attempts")
```
stormi, line 173: If we fail here, what happens to the VDI that we corrupted? The corruption would "leak" further than the function scope of the fixture?

tests/storage/linstor/test_linstor_sr.py
Millefeuille42: The flakiness is on DRBD's end. We could discuss it with Linbit to identify the issue.

tests/storage/linstor/test_linstor_sr.py
Millefeuille42: The teardown of `vm_on_linstor_sr_function` should happen, which calls `vm.destroy`, which calls `self.destroy_vdi(vdi_uuid)` on each of its VDIs. So it's not supposed to leak outside of its scope. Since it's a function scope I'd say it's acceptable here.

tests/storage/linstor/test_linstor_sr.py
Millefeuille42: I probably added it afterwards and forgot to rename the fixture. I'll adjust the name.

Millefeuille42: fixed in [57d6921](https://github.com/xcp-ng/xcp-ng-tests/pull/360/commits/57d69218d7326a73b4ebdc34454f826e805cd813)

tests/storage/linstor/test_linstor_sr.py
```diff
@@ -54,6 +108,87 @@ def test_snapshot(self, vm_on_linstor_sr: VM) -> None:
...
+        logging.info("[%s]: corrupting `%s`", host, lv_path)
+        vdi_host.ssh(shlex.join([
+            "dd",
+            "if=/dev/urandom",
+            f"of={lv_path}",
+            "bs=4096",
+            # Lower values seem to go undetected sometimes
+            "count=10000"  # ~40MB
+        ]))
+        yield vm, vdi_host, volume_name
```
stormi: I think could say that there's a tacit rule that we'd yield in the same order as we list objects in a fixture name. host, vm then volume_name,

tests/storage/linstor/test_linstor_sr.py
stormi: I agree. Maybe there would be a way to document it somehow but I have not precise suggestion. We just need to be careful if we start changing scopes, if we rely on implicit teardown by parent fixture.

stormi (approved):
One minor comment posted just before, but not a blocker.

tests/storage/linstor/test_linstor_sr.py
Millefeuille42: fixed in [7b32adb](https://github.com/xcp-ng/xcp-ng-tests/pull/360/commits/7b32adb323eaebd11d884a0bf5a66c99de134d0e).

klmp200 (approved):

PR #362, opened 2025-11-10
Collect logs in case of test failure

guzu:
For tests that fails rarely or only on CI we need a way to investigate the failure.
This PR is perform a log collection (only using xen-bugtool for the moment)

xen-bugtool reports can be very big (~100MB or more) so we might change that
later, or filter report.
Timestamps for each tests are recorded and the start and stop of tests, so that corresponding
logs can be extracted for analysis (see extract_logs.py). Timestamps are recorded in XCP-ng to avoid difference in
clocks.

No change is required to enable log collection. But console capture is only enabled
on demand, using @pytest.mark.capture_console.

Log collection do not happen at each test, but at the end of the test session.
For Windows (mainly), a screenshot can be done for each failing test; this uses the VNC console
through XO-lite.


conftest.py
```diff
@@ -760,3 +786,257 @@ def cifs_iso_sr(host, cifs_iso_device_config):
...
+                # Use the session-wide log directory
+                for vm in test_vms:
+                    try:
+                        screenshot_path = capture_vm_console(vm, session_log_dir)
+                        logging.debug(f"Console captured: {screenshot_path}")
+                    except Exception as e:
+                        logging.error(f"Failed to capture console for VM {vm.uuid}: {e}")
+
+@pytest.fixture(scope='session', autouse=True)
+def collect_logs_on_session_failure(request, hosts, session_log_dir):
```
dinhngtu: IMO this function is quite chatty logging-wise.

lib/host.py
```diff
@@ -84,6 +84,8 @@ def __init__(self, pool: Pool, hostname_or_ip):
         self.xcp_version_short = f"{self.xcp_version.major}.{self.xcp_version.minor}"
         self._dom0: Optional[VM] = None
 
+        self._check_bug_reports()
```
dinhngtu: A Host object could be constructed at any time. So this'd be better placed near e.g. `track_test_host()` to avoid issuing duplicate warnings.

dinhngtu (changes requested):
I've reviewed up to "lib: use option to put ssh in background". Could you squash the console capture commits, as it's a bit difficult to review otherwise?

guzu:
> > Sure, I could do that. About the script for console capture and VNC client part, it was "vibe coded". I didn't find a proper library to handle just the VNC decoding.
> 
> I found https://github.com/sibson/vncdotool which looks quite feature-rich.

I saw it, but it seems to handle the connection itself. In our case the VNC connection is handled by XAPI via websocket.

conftest.py
guzu: You're right. I tested in debug and didn't noticed that. I removed some put some others in debug.

guzu:
I change VM console capture and removed the VNC client implementation. Now the script mainly act as a proxy and uses a library for VNC capture.

dinhngtu:
"lib: check presence of xen-bugtool reports" has become empty. Could this commit be dropped, or did something get lost during the rebase?

guzu:
> "lib: check presence of xen-bugtool reports" has become empty. Could this commit be dropped, or did something get lost during the rebase?

Yes I remember seeing that. It seems that I forgot to remove the commit, but it wasn't lost it was merged into another. And it looks like it was merged into the wrong one, it should be in 10f4f1f and not 61d2e2f.

stormi:
Similarly to what I said regarding the follow-up commit, such big change needs to come with an explanation for the reviewers/maintainers regarding what the goal is, how it is meant to be used and when, how it is supposed to integrate in CI jobs, why this is the appropriate solution, etc. 

Here I'm not sure that @xcp-ng/os-platform-release as maintainers have been provided with these answers.

guzu:
> Similarly to what I said regarding the follow-up commit, such big change needs to come with an explanation for the reviewers/maintainers regarding what the goal is, how it is meant to be used and when, how it is supposed to integrate in CI jobs, why this is the appropriate solution, etc.

Description was updated

conftest.py
```diff
@@ -816,7 +816,6 @@ def bugreport_timestamp(request, host):
     logging.debug(f"Record timestamp (end): {class_name}")
     host.ssh(f"echo end   {class_name} $(date '+%s') >> {HOST_TIMESTAMPS_FILE}")
 
-
```
glehmann: This commit is mostly empty and doesn't match the description

scripts/capture-console.py
```diff
@@ -0,0 +1,300 @@
...
+            return result['result']
+        raise Exception(f"XAPI authentication failed: {result.get('error', result)}")
+
+
+async def main():
+    # Create argument parser with custom usage message
+    parser = argparse.ArgumentParser(
+        description='VNC WebSocket Console Capture',
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+        usage="""
```
glehmann: Isn't argparse able to generate that by itself?

scripts/capture-console.py
```diff
@@ -0,0 +1,300 @@
...
+            print(f"Authentication failed: {e}")
+            sys.exit(1)
+
+        # Construct WebSocket URL
+        ws_url = f"wss://{args.host}/console?uuid={args.vm_uuid}&session_id={session_id}"
+        cookie = None
+
+    else:
+        # Neither mode specified - show nice usage
+        print("VNC WebSocket Console Capture")
```
glehmann: This should also probably be generated by argparse

glehmann (commented):
Could you add the types in the signature of the new methods in conftest.py?

scripts/capture-console.py
guzu: Done

conftest.py
```diff
@@ -805,6 +812,49 @@ def session_log_dir():
     # postpone directory creation only if some test failed
     return log_dir
 
+# Helper function for immediate console capture
+def capture_vm_console(vm: VM, log_dir: str) -> str:
+    import sys
```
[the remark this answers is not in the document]
guzu: Done

conftest.py
```diff
@@ -45,6 +46,13 @@
     CACHE_IMPORTED_VM = False
 assert CACHE_IMPORTED_VM in [True, False]
 
+# Session-level tracking for failed tests and their associated hosts
+FAILED_TESTS_HOSTS = {}  # {test_nodeid: set(Host)}
```
ydirson: This comment essentially looks like a good candidate for being a type hint?

conftest.py
```diff
@@ -45,6 +46,13 @@
...
+SESSION_HAS_FAILURES = False
```
ydirson: Do we really need this?  It looks like it just reflects that `FAILED_TESTS_HOSTS` has contents or not?

conftest.py
```diff
@@ -760,3 +780,171 @@ def cifs_iso_sr(host, cifs_iso_device_config):
...
+    else:
+        log_dir = f"pytest-logs/report_{timestamp}"
+
+    # postpone directory creation only if some test failed
+    return log_dir
+
+
+# Record test timestamps to easily extract logs
+@pytest.fixture(scope='class', autouse=True)
+def bugreport_timestamp(request, host):
```
ydirson: "Record test start and end timestamps"?  Could make sense to be a docstring?

The "to easily extract logs" could have more details, as is it is no obvious how those timestamps help that.

ydirson (changes requested):
The review below is only for the host log collection feature.
The VM console capture and others should IMO live a separate PRs, there seems to be no reason one would block the other.
Same for the "smaller general improvements" like bumping pyright and ssh fix, they could have their own PR which could be merged faster.

PR #388, opened 2026-01-23
Add VM cache scanning script

dinhngtu:
Stale cached VMs will fill up the SR and cause test failures when importing new images.

This script helps detect those. It's non-destructive; deleting the VMs is left to the user.


dinhngtu:
Updated to call remote XAPI

gduperrey (approved):

PR #390, opened 2026-01-27
Bond tests

semarie:
- add some tests for Bond
- add new abstractions (`Network` and ` Bond`) with fixtures

### DIFF
--- a/jobs.py
+++ b/jobs.py
@@ -75,6 +75,20 @@ JOBS: dict[str, JobData] = {
         "paths": ["tests/misc", "tests/migration"],
         "markers": "multi_vms and not flaky and not reboot",
     },
+    "network-advanced": {
+        "description": "a group of network tests with complex prerequisites",
+        "requirements": [
+            "A pool with at least 1 host.",
+            "At least 2 free NICs on every host.",
+            "A small VM that can be imported on the SRs.",
+        ],
+        "nb_pools": 1,
+        "params": {
+            "--vm": "single/small_vm",
+        },
+        "paths": ["tests/network"],
+        "markers": "complex_prerequisites",
+    },
     "packages": {
         "description": "tests that packages can be installed correctly",
         "requirements": [
--- a/lib/network.py
+++ b/lib/network.py
@@ -1,18 +1,20 @@
 from __future__ import annotations
 
-from lib.common import _param_add, _param_clear, _param_get, _param_remove, _param_set, strtobool
+import logging
+
+from lib.common import _param_add, _param_clear, _param_get, _param_remove, _param_set, safe_split
 
 from typing import TYPE_CHECKING, Literal, overload
 
 if TYPE_CHECKING:
     from lib.host import Host
 
-class PIF:
-    xe_prefix = "pif"
+class Network:
+    xe_prefix = "network"
 
-    def __init__(self, uuid: str, host: Host):
-        self.uuid = uuid
+    def __init__(self, host: Host, uuid: str):
         self.host = host
+        self.uuid = uuid
 
     @overload
     def param_get(self, param_name: str, key: str | None = ...,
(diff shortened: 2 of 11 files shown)


tests/network/test_bond.py
```diff
@@ -0,0 +1,148 @@
...
+    return None
+
+@pytest.mark.complex_prerequisites
+@pytest.mark.small_vm
+@pytest.mark.parametrize("mode", ["lacp", "active-backup", "balance-slb"])
+def test_bond(host: Host, empty_network: Network, imported_vm: VM, mode):
+    network = empty_network
+    assert len(network.PIF_uuids()) == 0
+
+    assert host.ssh_with_result("which tcpdump").returncode == 0, \
```
glehmann: This is installed by default on xcp-ng, so you can remove the assert IMO

lib/network.py
```diff
@@ -0,0 +1,51 @@
...
+        _param_remove(self.host, Network.xe_prefix, self.uuid, param_name, key, accept_unknown_key)
+
+    def destroy(self):
+        logging.info(f"Destroying network '{self.param_get('name-label')}': {self.uuid}")
+        self.host.xe('network-destroy', {'uuid': self.uuid})
+
+    def PIF_uuids(self) -> list[str]:
+        return safe_split(self.param_get('PIF-uuids'), '; ')
+
+    def VIF_uuids(self) -> list[str]:
```
glehmann:
```suggestion
    def vif_uuids(self) -> list[str]:
```

tests/network/test_bond.py
```diff
@@ -0,0 +1,130 @@
...
+            assert vm1.ssh_with_result(["ping", "-c3", "-w5", "169.254.2.1"]).returncode == 0
+            assert vm2.ssh_with_result(["ping", "-c3", "-w5", "169.254.1.1"]).returncode == 0
+
+        finally:
+            # VIFs are destroyed by VM.destroy()
+            vm2.destroy()
+            vm1.destroy()
+
+def _wait_for_packet(host: Host | VM, interface: str, sfilter: str, timeout: int = 30) -> None:
+    ret = host.ssh_with_result(f"timeout {timeout} tcpdump -i {interface} -n -c1 '{sfilter}'")
```
glehmann: I think you can just use `host.ssh("…")`. `ssh` already checks the return code and generates an exception (containing the output and the error code) that will make the test fail.
I don't think you need `ssh_with_result()` in this PR

glehmann (commented):
A few minor comments.
I let someone more knowledgeable in network comment on what the test is actually doing :)

glehmann (approved):
(forgot to click on approve when submitting my review)

bleader (approved):

tests/network/test_bond.py
```diff
@@ -0,0 +1,132 @@
...
+
+from lib.bond import Bond
+from lib.common import PackageManagerEnum
+from lib.host import Host
+from lib.network import Network
+from lib.vm import VM
+
+# Requirements:
+# From --hosts parameter:
+# - host(A1): first XCP-ng host, with at least 3 NICs (1 management, 2 unused), with tcpdump
```
stormi: You're mentioning a first host. Does it mean there must be a second one? Else, I think we usuallyjust say "an XCP-ng host".

stormi (changes requested):
Sorry for the late review. I had a quick look, ready to approve as there were already two approvals, but there are many things that I see as important enough to ask for a new round. We can try to find a time to discuss them vocally, so that I can explain whatever might be unclear in my comments and how I think the tests should look like in the end (a lot shorter, with more fixtures).

tests/network/conftest.py
```diff
@@ -1,7 +1,47 @@
...
+from lib.network import Network
+
+from typing import Generator
+
 @pytest.fixture(scope='package')
-def host_no_sdn_controller(host):
+def host_no_sdn_controller(host: Host):
     """ An XCP-ng with no SDN controller. """
     if host.xe('sdn-controller-list', minimal=True):
         pytest.skip("This test requires an XCP-ng with no SDN controller")
```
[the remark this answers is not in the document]
semarie: I kept it as it was. Should I add a separated commit to switch it to fail ?

jobs.py
```diff
@@ -468,6 +468,8 @@
 BROKEN_TESTS = [
     # not really broken but has complex prerequisites (3 NICs on 3 different networks)
     "tests/migration/test_host_evacuate.py::TestHostEvacuateWithNetwork",
+    # not really broken but has complex prerequisites (3 NICs)
+    "tests/network/test_bond.py::test_bond",
```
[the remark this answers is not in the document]
semarie: ok. going back to complex_prerequisites

tests/network/conftest.py
```diff
@@ -1,7 +1,104 @@
...
+
+from typing import Generator
+
 @pytest.fixture(scope='package')
-def host_no_sdn_controller(host):
+def host_no_sdn_controller(host: Host):
     """ An XCP-ng with no SDN controller. """
     if host.xe('sdn-controller-list', minimal=True):
-        pytest.skip("This test requires an XCP-ng with no SDN controller")
+        pytest.fail("This test requires an XCP-ng with no SDN controller")
```
stormi, line 19: You left pytest.skip in the third commit and then fixed it in the fifth commit by removing the extraneous line. Please fix it in the third commit directly.

stormi (changes requested):
Just one remaining comment.

Regarding the rest of the test, the next step before merging would be seeing with the @xcp-ng/os-platform-release team how to add this test to CI (using nested XCP-ng hosts with 3 nics? - which would only work in BIOS mode unless we'd make it so UEFI VMs can have more than 2 emulated NICs, right?).

tests/network/conftest.py
semarie: my bad. fixed.

semarie:
> Regarding the rest of the test, the next step before merging would be seeing with the @xcp-ng/os-platform-release team how to add this test to CI (using nested XCP-ng hosts with 3 nics? - which would only work in BIOS mode unless we'd make it so UEFI VMs can have more than 2 emulated NICs, right?).

yes, if using XCP-ng nested, it should be in BIOS mode for now, as with UEFI the XCP-ng couldn't have more than 2 emulated NICs.

stormi (approved):

stormi:
@semarie are there changes to review in the latest push?

semarie:
As Gaël provided to me an xcp-ng host to look at integrating in CI, I fixed some wrong assumptions (like assuming "eth1" would in the free nics list). I pushed these fixes, and now I am looking on CI integration.

semarie:
> @semarie are there changes to review in the latest push?

yes they should be reviewed.
but as I will add some others changes of CI integration (after discussion), it could also wait a bit.

semarie:
updated PR after type hints merge.
I checked each commit for correctness (running `make`), and next checked that tests was still good

stormi (approved):
I haven't re-reviewed, but there are two approvals already.

PR #393, opened 2026-02-09
Add Intel Temperature sensors tests

TSnake41:
Also add cpu policy code to query CPU policies if tests needs to (DTS wants it).

Needs ~~https://github.com/xcp-ng-rpms/xen/pull/78~~

### DIFF
--- a/jobs.py
+++ b/jobs.py
@@ -475,6 +475,17 @@ JOBS: dict[str, JobData] = {
         "nb_pools": 1,
         "params": {},
         "paths": ["tests/misc/test_cpu_policy.py"]
+    },
+    "dts": {
+        "description": "Test verifying that Intel Digital Temperature Sensors works",
+        "requirements": [
+            "1 XCP-ng host >= 8.3",
+            "Intel CPU",
+            "DTS capability available"
+        ],
+        "nb_pools": 1,
+        "params": {},
+        "paths": ["tests/misc/test_dts.py"],
     }
 }
 
--- a/jobs.py
+++ b/jobs.py
@@ -469,6 +469,12 @@ JOBS: dict[str, JobData] = {
             "--vm[]": "multi/limits",
         },
         "paths": ["tests/limits"],
+    },
+    "cpu-policy": {
+        "description": "Tests displaying CPU policies",
+        "nb_pools": 1,
+        "params": {},
+        "paths": ["tests/misc/test_cpu_policy.py"]
     }
 }
 
--- a/tests/misc/test_dts.py
+++ b/tests/misc/test_dts.py
@@ -1,0 +1,22 @@
+import pytest
+
+import logging
+
+from lib.cpu_policy import NO_SUBLEAF, HostCpuPolicy
+
+# Requirements:
+# - one XCP-ng host (--host) >= 8.3
+# - Intel CPU with DTS support (i.e no nested virtualization)
+
+@pytest.mark.usefixtures("host_at_least_8_3")
+class TestDts:
+    def test_dts(self, host):
+        cpu_policy = HostCpuPolicy(host)
+
+        power_leaf = cpu_policy.policies["Host"].cpuid.get((0x6, NO_SUBLEAF))
+        
+        if power_leaf is None or power_leaf[0] & 1 == 0:
+            pytest.skip("This test require a host with DTS support")
+            return
+
+        host.ssh(["xenpm get-intel-temp"])
--- a/tests/xen/test_dts.py
+++ b/tests/xen/test_dts.py
@@ -1,6 +1,7 @@
 import pytest
 
 import logging
+import re
 
 from lib.cpu_policy import NO_SUBLEAF, HostCpuPolicy
 
@@ -31,4 +32,18 @@ class TestDts:
         if power_leaf is None or power_leaf.eax & 1 == 0:
             pytest.skip("This test require a host with DTS support")
 
-        host.ssh("xenpm get-core-temp")
+        output = host.ssh("xenpm get-core-temp")
+
+        # Test if all CPU temperatures are acceptable
+        TEMPERATURE_THRESHOLD_MIN = 5
+        TEMPERATURE_THRESHOLD_MAX = 120
+        has_any_temp = False
+
+        for (entry, temp) in re.findall(r"(\D+\d+): (\d+)°C", output):
+            temp = int(temp)
+            has_any_temp = True
+
+            assert TEMPERATURE_THRESHOLD_MIN < temp < TEMPERATURE_THRESHOLD_MAX, \
+                f"{entry} temperature must be within safe operational limits"
+
+        assert has_any_temp, "The tool should at least output 1 temperature"
--- a/tests/xen/test_dts.py
+++ b/tests/xen/test_dts.py
@@ -1,0 +1,34 @@
+import pytest
+
+import logging
+
+from lib.cpu_policy import NO_SUBLEAF, HostCpuPolicy
+
+# Requirements:
+# - one XCP-ng host (--host) >= 8.3
+# - Intel CPU with DTS support (i.e no nested virtualization)
+
+@pytest.mark.usefixtures("host_at_least_8_3")
+class TestDts:
+    def test_dts(self, host):
+        """
+        We can't meaningfully check if DTS/PTS hardware feature works neither
+        if Xen returns correct values. The goal here is to make sure that DTS
+        bit exposed implies that `xenpm get-core-temp` doesn't fail.
+
+        The use of Host policy requires [1] backport, otherwise, Host policy
+        doesn't match the hardware CPUID. IOW, the test assumes that this backport
+        exist, or at least, if it's missing, it's only because the DTS feature hasn't
+        been implemented (yet) in this XCP-ng build.
+
+        [1] f6894fdf ("x86/cpu-policy: move invocation of recalculate_misc()")
+        https://gitlab.com/xen-project/xen/-/commit/f6894fdfa83359d0686a41aee5c2a8ba7b2878b4
+        """
+        cpu_policy = HostCpuPolicy(host)
+
+        power_leaf = cpu_policy.policies["Host"].cpuid.get((0x6, NO_SUBLEAF))
+
+        if power_leaf is None or power_leaf.eax & 1 == 0:
+            pytest.skip("This test require a host with DTS support")
+
+        host.ssh("xenpm get-core-temp")
--- a/lib/cpu_policy.py
+++ b/lib/cpu_policy.py
@@ -1,0 +1,83 @@
+from lib.host import Host
+
+NO_SUBLEAF = 0xffffffff
+
+class CpuPolicy:
+    """A specific CPU policy (PV, HVM, ...)"""
+    # (leaf, subleaf) -> (eax, ebx, ecx, edx)
+    cpuid: dict[(int, int), (int, int, int, int)]
+    # msr -> val
+    msr: dict[int, int]
+
+    def __init__(self, cpuid: dict[(int, int), int], msr: dict[int, int]):
+        self.cpuid = cpuid
+        self.msr = msr
+
+class HostCpuPolicy:
+    """All CPU policies of a host (CPUID, specific MSRs)"""
+    policies: dict[str, CpuPolicy]
+
+    def __init__(self, host: Host):
+        self.policies = dict()
+        text = host.ssh("xen-cpuid --policy")
+        current_policy: CpuPolicy = None
+        mode = None  # None | "cpuid" | "msr"
+
+        lines = text.splitlines()
+
+        for line in lines:
+            line = line.rstrip()
+
+            # ---- Policy header ----
+            if "policy:" in line:
+                # Example: "Raw policy: 32 leaves, 2 MSRs"
+                name, rest = line.split("policy:", 1)
+                name = name.strip()
+
+                self.policies[name] = CpuPolicy(dict(), dict())
+                current_policy = self.policies[name]
+                mode = None
+                continue
+
+            if current_policy is None:
+                continue
+
+            # ---- Section switches ----
+            if line.strip() == "CPUID:":
+                mode = "cpuid"
+                continue
+
+            if line.strip() == "MSRs:":
+                mode = "msr"
+                continue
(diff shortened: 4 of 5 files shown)


dinhngtu (changes requested):

tests/misc/test_dts.py
```diff
@@ -0,0 +1,26 @@
...
+# Requirements:
+# - one XCP-ng host (--host) >= 8.3
+# - Intel CPU with DTS support (i.e no nested virtualization)
+
+@pytest.mark.usefixtures("host_at_least_8_3")
+class TestDts:
+    def test_dts(self, host):
+        policy = get_cpu_policy(host)
+
+        power_leaf = policy["Host"]["cpuid"][(0x6, NO_SUBLEAF)]
```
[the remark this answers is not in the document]
TSnake41: What do you mean by using `.get()` ?

tests/misc/test_dts.py
```diff
@@ -0,0 +1,35 @@
...
+        """
+        cpu_policy = HostCpuPolicy(host)
+
+        power_leaf = cpu_policy.policies["Host"].cpuid.get((0x6, NO_SUBLEAF))
+
+        if power_leaf is None or power_leaf.eax & 1 == 0:
+            pytest.skip("This test require a host with DTS support")
+            return
+
+        host.ssh(["xenpm get-intel-temp"])
```
dinhngtu: Hasn't the name of this command change since?

tests/misc/test_dts.py
```diff
@@ -0,0 +1,35 @@
...
+        [1] f6894fdf ("x86/cpu-policy: move invocation of recalculate_misc()")
+        https://gitlab.com/xen-project/xen/-/commit/f6894fdfa83359d0686a41aee5c2a8ba7b2878b4
+        """
+        cpu_policy = HostCpuPolicy(host)
+
+        power_leaf = cpu_policy.policies["Host"].cpuid.get((0x6, NO_SUBLEAF))
+
+        if power_leaf is None or power_leaf.eax & 1 == 0:
+            pytest.skip("This test require a host with DTS support")
+            return
```
dinhngtu: Technically you don't need `return` here since `pytest.skip` raises an error, but it doesn't make much of a difference for me personally.

tests/misc/test_dts.py
TSnake41: yes it did, just fixed it

jobs.py
```diff
@@ -575,6 +575,17 @@ class JobData(TypedDict):
             "--vm[]": "multi/limits",
         },
         "paths": ["tests/limits"],
+    },
+    "dts": {
+        "description": "Test verifying that Intel Digital Temperature Sensors works",
```
glehmann: Can't we have the test added to an already existing job?
The test already skips if the hardware is not supported.

TSnake41: Just moved the test in xen job with https://github.com/xcp-ng/xcp-ng-tests/compare/57c6a9c53d3aac73ca86da4578c3001e8d0db47e..3fe393cecc99023733ba1a6b24c55701c5621ba3.

fallen (approved):
lgtm

tests/xen/test_dts.py
```diff
@@ -0,0 +1,34 @@
...
+        https://gitlab.com/xen-project/xen/-/commit/f6894fdfa83359d0686a41aee5c2a8ba7b2878b4
+        """
+        cpu_policy = HostCpuPolicy(host)
+
+        power_leaf = cpu_policy.policies["Host"].cpuid.get((0x6, NO_SUBLEAF))
+
+        if power_leaf is None or power_leaf.eax & 1 == 0:
+            pytest.skip("This test require a host with DTS support")
+
+        host.ssh(["xenpm get-core-temp"])
```
stormi: Is there a way to differenciate meaningful output from output that would mean that it basically couldn't get any relevant information?

TSnake41: That's a (partially) insolvable problem, due to the hardware aspect of the feature.
If it can't query anything, it will fail; although, the feature can "fail" in numerous ways without being "trivially" detectable.

For instance, if the related Xen hypercall returns all zeros (e.g due to a regression), the final reading will be all 0°C, which is obviously incorrect, but a legal value that you can have on real hardware (in unusual conditions). Things get trickier if the value is corrupted, and you still get incoherent yet still legal values in the end, that could bypass all sanity checks.

The most obvious regression is losing the feature, which currently leads to skipped test, it can be worked around by checking for "Raw policy" instead (so it bypasses the need for https://github.com/xen-project/xen/commit/f6894fdfa83359d0686a41aee5c2a8ba7b2878b4), but that makes this test fail on all XCP-ng versions that don't have the feature (maybe we care about that instead ?).

To be fair, the best way of testing this feature is manually and checking if things are consistent and realistic in the context of the test.

stormi: Differenciating between hardware support and Xen support would be a good first step towards robustifying the test. We can assume XCP-ng 8.3 supports the feature, as our tests apply to the moving target that is XCP-ng X.Y with all updates applied, and we don't offer a granularity higher than X.Y.

However if we want, we can. We can detect if Xen is at least in some package version, for example, and skip when the release is too ancient. I'm not saying we should (or shouldn't), but I'm saying we can.

As for detecting whether the command gave a sensible result, I'm pretty sure we can have simple heuristics to at least detect obvious problems. We can't guarantee the result consistency (we won't go put a thermometer on the CPU), but we can detect things that are clearly failure cases. But maybe checking for the return code as currently implicitly done is enough of a smoke test. You can tell us.

This is the kind of things that brings a test closer to being able to detect regressions, which is the purpose of this test.

dinhngtu: My take is, this test checks for the support of `xenpm get-core-temp` using what's basically implementation detail of how get-core-temp works (since the command name was generalized upstream to cover anything with a temp sensor in theory). As such, it's still not quite clear to me what are the conditions where this test is supposed to pass or fail.

tests/xen/test_dts.py
stormi, line 1: Moving discussion to a thread.

>> IMO, we can test that either the command succeeded, or the command failed specifically because get-core-temp was not supported on the host; if the command failed for any other reason, it should fail the test.

> I assume coupling between x86/cpu-policy: move invocation of recalculate_misc() and support for get-core-temp as they come in Xen in the same release (part of 4.22). And in our case are backported in the same PR.

If you're saying that we consider that the tested Xen has support for it, yes, that's a valid requirement. So it's OK to fail the test for any failure of the command as currently done.

> We can't perform hypercalls in pytest due to the test logic not running on the end host

We can run anything on the end host, via SSH. So there's no "we can't" here. Maybe there's a "it would be complex", but I'm not even sure it is. It depends what can be done from the host itself.

>> We can also argue that failing the test on certain values is sensible. Sure, 0oC is possible in some unusual conditions (LN2 etc), but do our test labs run the CPU at 0oC?

> Well, that would be biased to a specific failure case.

Objection rejected. As you stated yourself the purpose is to smoke-test the feature. Thus don't dismiss checks that would only cover part of the failure cases on the pretext that they are not exhaustive.

I myself already asked earlier in a thread that we'd consider checking the actual output, which you ignored and now dismiss. It's time to comply with what the maintainers ask, now, not find pretextes to dimiss their requests.

stormi (changes requested):
Marking as changes requested, so that the status is clear.

tests/xen/test_dts.py
```diff
@@ -0,0 +1,45 @@
...
+
+        # Test if all CPU temperatures are acceptable
+        TEMPERATURE_THRESHOLD_MIN = 5
+        TEMPERATURE_THRESHOLD_MAX = 120
+
+        for (entry, temp) in re.findall(r"(\D+\d+): (\d+)°C", output):
+            temp = int(temp)
+
+            if temp > TEMPERATURE_THRESHOLD_MAX or temp < TEMPERATURE_THRESHOLD_MIN:
+                pytest.fail(f"'{entry}' temperature (={temp}) is outside threshold !")
```
fallen: This can also be written like this:
`assert TEMPERATURE_THRESHOLD_MIN < temp < TEMPERATURE_THRESHOLD_MAX, "temperature is outside threshold"`

Not sure which is better.
But this is just a question of style anyway, so it's definitely not a blocking comment.

TSnake41: I think it can be useful to keep what entry got problematic, but I don't know if assert allows inline formatted strings.

dinhngtu: `assert` is almost certainly what we want here as it indicates that this is an expected result of the test. From my quick testing, assert with a formatted string seems to work, but I don't think it's absolutely needed.

TSnake41: hmm, tried it; but now linter complains with a too large line, and I can't figure out how to cut it

TSnake41: I found a hack that is documented in the PEP 8 styleguide.

tests/xen/test_dts.py
```diff
@@ -0,0 +1,45 @@
...
+
+        # Test if all CPU temperatures are acceptable
+        TEMPERATURE_THRESHOLD_MIN = 5
+        TEMPERATURE_THRESHOLD_MAX = 120
+
+        for (entry, temp) in re.findall(r"(\D+\d+): (\d+)°C", output):
+            temp = int(temp)
+
+            assert temp < TEMPERATURE_THRESHOLD_MAX and temp > TEMPERATURE_THRESHOLD_MIN, \
+                f"'{entry}' temperature (={temp}) is outside threshold !"
```
glehmann: nitpick:

```suggestion
            assert TEMPERATURE_THRESHOLD_MIN < temp < TEMPERATURE_THRESHOLD_MAX, f"{entry} temp is out of range"
```
Python supports multiple comparisons at once, and pytest shows the compared values, so you don't need to show them in the message:

```
>       assert TEMPERATURE_THRESHOLD_MIN < temp < TEMPERATURE_THRESHOLD_MAX, f"{entry} temp is out of range"
E       AssertionError: foo temp is out of range
E       assert 5 < 0
```

glehmann (approved):
Just a minor comment, feel free to ignore it if you prefer

tests/xen/test_dts.py
```diff
@@ -0,0 +1,44 @@
...
+
+@pytest.mark.usefixtures("host_at_least_8_3")
+class TestDts:
+    def test_dts(self, host):
+        """
+        We can't meaningfully check if DTS/PTS hardware feature works neither
+        if Xen returns correct values. The goal here is to make sure that DTS
+        bit exposed implies that `xenpm get-core-temp` doesn't fail.
+
+        The use of Host policy requires [1] backport, otherwise, Host policy
```
dinhngtu, line 20:
```suggestion
        The use of Host policy requires [1] backport, otherwise, Host policy
```

tests/xen/test_dts.py
```diff
@@ -0,0 +1,44 @@
...
+        output = host.ssh("xenpm get-core-temp")
+
+        # Test if all CPU temperatures are acceptable
+        TEMPERATURE_THRESHOLD_MIN = 5
+        TEMPERATURE_THRESHOLD_MAX = 120
+
+        for (entry, temp) in re.findall(r"(\D+\d+): (\d+)°C", output):
+            temp = int(temp)
+
+            assert TEMPERATURE_THRESHOLD_MIN < temp < TEMPERATURE_THRESHOLD_MAX, f"{entry} temp is out of range"
```
stormi: The message from an assert must describe what is expected, not what is wrong. Or can be omitted altogether if the assert condition itself is already clear enough (IMO it's the case here).

Other comments made internally, regarding detecting the absence of temperature entry for devices for which we'd expect results.

TSnake41: this is contradictory to @glehmann comment https://github.com/xcp-ng/xcp-ng-tests/pull/393#discussion_r3537635080

TSnake41: > Other comments made internally, regarding detecting the absence of temperature entry for devices for which we'd expect results.

absence of data is already a error condition in xenpm binary

dinhngtu: > this is contradictory to @glehmann comment [#393 (comment)](https://github.com/xcp-ng/xcp-ng-tests/pull/393#discussion_r3537635080)

I don't think it's contradictory, can't both comments be resolved by simply dropping the assert error message?

> > Other comments made internally, regarding detecting the absence of temperature entry for devices for which we'd expect results.
> 
> absence of data is already a error condition in xenpm binary

The test should not need to be aware of whether absence of data is an error condition in xenpm.

dinhngtu (approved):

tests/xen/test_dts.py
glehmann: The `entry` is actually useful in the assertion message.

I suppose it can be written as `f"{entry} temperature must be within safe operational limits"`

stormi: Let's do as Gaëtan suggests in his latest comment.

> this is contradictory to @glehmann comment #393 (comment)

I guess Gaëtan just missed the fact that your message was an error message rather than an assert message. So he offered a different message, but still following your initial logic.

My comment was legit.

stormi (changes requested):
Only two changes left to be done to the PR:
- assert message logic inversion
- despite we expect xenpm to return nonzero when there's no CPU with any temperature data, let's verify that the output does contain at least one entry.

fallen (approved):
lgtm

tests/xen/test_dts.py
```diff
@@ -0,0 +1,49 @@
...
+        has_any_temp = False
+
+        for (entry, temp) in re.findall(r"(\D+\d+): (\d+)°C", output):
+            temp = int(temp)
+            has_any_temp = True
+
+            assert TEMPERATURE_THRESHOLD_MIN < temp < TEMPERATURE_THRESHOLD_MAX, \
+                f"{entry} temperature must be within safe operational limits"
+
+        assert has_any_temp, "No temperature parsed"
```
fallen: The assert "text" should be inversed => "the tool should at least output 1 temperature"

fallen (changes requested):

fallen (approved):
lgtm

stormi (approved):
All good now, thanks!

PR #394, opened 2026-02-11
Add traffic rules tests

semarie:
- [x] proper integration in tests
- [x] simple rules
  - [x] VIF rule: add/delete with simple vm.start/destroy cycle
  - [x] Network rule: add/delete with simple vm.start/destroy cycle
    - **FAIL** without cookie usage:
      - [xo-server-sdn-controller part](https://github.com/vatesfr/xen-orchestra/compare/srt/xo-server-sdn-controller/cookie)
      - [xapi plugin part](https://github.com/xcp-ng/xcp-ng-xapi-plugins/pull/58)
- [x] VM migration (VIF / Network rules **FAIL** without some patches)
- [x] tests with VLAN
  - [x] VIF rule
  - [x] Network rule (**FAIL** without cookie PR)
- [x] tests with Tunnel (aka 'Private Network' in XO)


semarie:
To properly integrate this PR, the following PR should be merged first:
- [ ] https://github.com/xcp-ng/xcp-ng-xapi-plugins/pull/58
- [ ] https://github.com/vatesfr/xen-orchestra/pull/9518

bleader:
you're missing a newline in the last commit message, making the title being the whole commit log

lib/pif.py
```diff
@@ -26,3 +33,49 @@ def param_add(self, param_name, value, key=None):
...
+        return mode
+
+    def device(self) -> str:
+        device = self.param_get("device")
+        assert device
+        return device
+
+    def vlan(self) -> int | None:
+        vlan = self.param_get('VLAN')
+        if vlan is None:
```
bleader: I believe xapi can return -1 when there is no vlan, shouldn't this be something like `if vlan is None or vlan == -1:`? Or is than handled elsewhere I missed with the VLAN class maybe?

tests/network/conftest.py
```diff
@@ -1,7 +1,200 @@
...
+        network_uuid = tunnel.access_PIF().network_uuid()
+        network = Network(host, network_uuid)
+
+        yield tunnel
+    finally:
+        if tunnel is not None:
+            tunnel.destroy()
+
+        if network is not None:
+            # sdnController.createPrivateNetwork might have create several Tunnel (one per host)
```
bleader: created

tests/network/conftest.py
```diff
@@ -1,7 +1,200 @@
...
+        tunnel = Tunnel(host, tunnel_uuid)
+
+        # get Network from Tunnel
+        network_uuid = tunnel.access_PIF().network_uuid()
+        network = Network(host, network_uuid)
+
+        yield tunnel
+    finally:
+        if tunnel is not None:
+            tunnel.destroy()
```
bleader: wouldn't that one also be caught by the loop below?

jobs.py
```diff
@@ -467,6 +467,10 @@
 BROKEN_TESTS = [
     # not really broken but has complex prerequisites (3 NICs on 3 different networks)
     "tests/migration/test_host_evacuate.py::TestHostEvacuateWithNetwork",
+    # not really broken but has complex prerequisites (3 NICs)
+    "tests/network/test_bond.py::TestBond",
+    # not really broken but has complex prerequisites (1 free NICs)
```
bleader: 1 free NIC

tests/network/test_traffic_rules.py
```diff
@@ -0,0 +1,772 @@
...
+    def test_networkRule(self, connected_hosts_with_xo: list[Host], hostA2: Host,
+                         local_sr_on_hostA1, local_sr_on_hostA2, running_vm: VM):
+        hostA1 = connected_hosts_with_xo[0]
+
+        vm = running_vm
+        networkId = hostA1.management_network()
+        hostA1Br = Network(hostA1, networkId).bridge()
+        hostA2Br = Network(hostA2, networkId).bridge()
+
+        # the test is asymetric, start with the VM on know host
```
bleader: assymetric

tests/network/conftest.py
semarie: yes, it should be destroyed in the next loop. I also cleared the try/finally as we are in fixture and pytest is already managing exceptions in tests.

lib/pif.py
semarie: thanks to caught this. `None` case is for `"VLAN"` property is absent (which is not possible).
I changed it to `assert vlan`, and next convert `-1` to `None`.

semarie:
I also enhanced a bit the `tunnel` fixture to avoid destroying **all** tunnels attached to managment interface. now, it only destroy the ones created (but a race is still possible)

semarie:
Last push:
- resolving conflicts (`lib/vif.py`)
- rebasing

Marking ready for review as all components are published with 8.3-20260818

lib/tunnel.py
```diff
@@ -0,0 +1,57 @@
...
+        _param_add(self.host, self.xe_prefix, self.uuid, param_name, value, key)
+
+    def param_clear(self, param_name: str) -> None:
+        _param_clear(self.host, self.xe_prefix, self.uuid, param_name)
+
+    def destroy(self):
+        logging.info(f"Destroying Tunnel: {self.uuid}")
+        self.host.xe('tunnel-destroy', {'uuid': self.uuid})
+
+    def access_PIF(self) -> PIF:
```
glehmann:
```suggestion
    def access_pif(self) -> PIF:
```

lib/tunnel.py
```diff
@@ -0,0 +1,57 @@
...
+        uuid = self.param_get("access-PIF")
+        assert uuid
+        return PIF(uuid, self.host)
+
+    def transport_PIF(self) -> PIF:
```
glehmann:
```suggestion
    def transport_pif(self) -> PIF:
```

lib/vlan.py
```diff
@@ -0,0 +1,62 @@
...
+    def destroy(self):
+        logging.info(f"Destroying VLAN: {self.uuid}")
+        self.host.xe('vlan-destroy', {'uuid': self.uuid})
+
+    def tag(self) -> int:
+        tag = self.param_get('tag')
+        assert tag
+        return int(tag)
+
+    def tagged_PIF(self) -> PIF:
```
glehmann:
```suggestion
    def tagged_pif(self) -> PIF:
```

lib/vlan.py
```diff
@@ -0,0 +1,62 @@
...
+        uuid = self.param_get("tagged-PIF")
+        assert uuid
+        return PIF(uuid, self.host)
+
+    def untagged_PIF(self) -> PIF:
```
glehmann:
```suggestion
    def untagged_pif(self) -> PIF:
```

tests/network/conftest.py
```diff
@@ -18,6 +24,101 @@ def host_no_sdn_controller(host: Host) -> None:
     if host.xe('sdn-controller-list', minimal=True):
         pytest.fail("This test requires an XCP-ng with no SDN controller")
 
+RE_digits = re.compile(r'(\d+)')
```
glehmann:
```suggestion
RE_DIGITS = re.compile(r'(\d+)')
```

tests/network/conftest.py
```diff
@@ -18,6 +24,101 @@ def host_no_sdn_controller(host: Host) -> None:
...
+
+def compare_versions(a: str, b: str) -> int:
```
glehmann: Wouldn't `packaging.version.Version` work?

tests/network/conftest.py
```diff
@@ -110,3 +200,107 @@ def bond_balanceslb(host: Host, empty_network: Network):
...
+@pytest.fixture
+def tunnel(
+    hosts_with_xo: list[Host],
+    tunnel_protocol: str, tunnel_encryption: bool,
+) -> Generator[Tunnel, None, None]:
+    host = hosts_with_xo[0]
+
+    # check system requirements
+    if not host.is_package_installed("openvswitch-ipsec"):
+        pytest.fail("'tunnel' fixture requires configuration, see https://docs.xen-orchestra.com/sdn_controller")
```
glehmann: We usually prefer to install the required dependencies in the fixture setup, and remove it in the teardown.

tests/network/test_traffic_rules.py
```diff
@@ -0,0 +1,811 @@
...
+# xo-cli (on the host running the test) is expected to be usable
+# From --hosts parameter:
+# - host(A1): first XCP-ng host (no traffic rules should be already present on the host)
+# From --vm parameter
+# - A VM to import
+
+# Special requirements for some tests:
+# - TestVLAN needs at least 1 free NICs (see HOST_FREE_NICS in data.py)
+# - TestMigrate needs second XCP-ng host in the same pool
+# - TestTunnel will create encrypted tunnel (and only one could be created at a time)
```
glehmann, line 30: So it requires no encrypted tunnel already configured?

tests/network/test_traffic_rules.py
```diff
@@ -0,0 +1,811 @@
...
+
+CACHE_ovs_vsctl_bridge_to_parent: dict[str, str] = {}
```
glehmann: Consider renaming this to `cache_ovs_vsctl_bridge_to_parent` (lowercase snake_case).
Since this dictionary represents mutable state rather than an immutable module-level constant, PEP 8 recommends standard lowercase naming.

Also, don't we risk leaking data between tests with a module-level cache?

glehmann (commented):
Some minor comments: I clearly don't have the network knowledge to evaluate the tests.

tests/network/test_traffic_rules.py
```diff
@@ -0,0 +1,811 @@
...
+
+def ovs_vsctl_bridge_to_parent(host: Host, br: str) -> str:
+    global CACHE_ovs_vsctl_bridge_to_parent
```
glehmann: I think this line is not needed, as we just modify a key in the dict, and not replacing the dict itself.
We usually avoid the `global` usage.

tests/network/test_traffic_rules.py
```diff
@@ -0,0 +1,811 @@
...
+                ),
+                msg="Wait for the rule to not apply for now (nobody connected to the network)",
+                timeout_secs=30,
+                delayed_start_secs=2,
+            )
+
+        finally:
+            # delete networkRule
+            logging.info("sdnController.deleteNetworkRule")
+            xo_cli('sdnController.deleteNetworkRule', {
```
glehmann: Shouldn't it go in a `defer()`?

tests/network/conftest.py
semarie: I will take a look. I didn't know about it

tests/network/test_traffic_rules.py
semarie: no. the test will setup a tunnel itself. it is created by the `tunnel` fixture.

tests/network/test_traffic_rules.py
semarie: it can't go in `defer()` because it is also part of the test and things are tested after running it.
so if I put it in `defer()`, it means we can't test that the delete was properly done.

tests/network/conftest.py
semarie: it doesn't like yum version...

```py
>>> Version("2.1.15-2.xcpng8.3")
  File "<python-input-15>", line 1, in <module>
    Version("2.1.15-2.xcpng8.3")
    ~~~~~~~^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.14/site-packages/packaging/version.py", line 202, in __init__
    raise InvalidVersion(f"Invalid version: {version!r}")
packaging.version.InvalidVersion: Invalid version: '2.1.15-2.xcpng8.3'
```

glehmann: I missed that it's used to compare rpm versions.
Comparing versions in the rpm world is notoriously difficult to do properly, and unfortunately I don't know of any maintained pure-Python module that does it.
I suppose this is good enough for this test, though :)

tests/network/test_traffic_rules.py
semarie: from generic point of view, I agree that module-level cache is not optimal.
I changed it a bit to have a key linked to the host and to the current test. this way, it should still cache data (and avoid too much ssh to the host) and don't be shared between tests.

semarie:
Last push:
- integrate remark from @glehmann

tests/network/test_traffic_rules.py
glehmann: It can still go in a `defer`, but it needs to be guarded, to make sure it's not already deleted:

```python
            xo_cli('sdnController.addNetworkRule', {
                'networkId': networkId,
                'ipRange': '10.0.0.1',
                'direction': 'to',
                'protocol': 'icmp',
                'allow': 'false',
            })
            fn delete_rule():
                xo_cli('sdnController.deleteNetworkRule', {
                    'networkId': networkId,
                    'ipRange': '10.0.0.1',
                    'direction': 'to',
                    'protocol': 'icmp',
                })
                return True
            deleted = False
            defer(lambda: delete_rule() if not deleted else None)


            # do some stuff


            deleted = delete_rule()
            assert …
```

This is more convoluted than the try/finally, but in case of problem, it prevents the rule from being deleted before reaching the debugger.
Just pick the version you prefer :)

glehmann (approved):

tests/network/conftest.py
stormi: Try `packaging.version.LooseVersion`. 

If this doesn't work, then a generic function to compare versions should not live in at network specific conftest.py, but in lib/common instead.

stormi: Else, we could have a look at https://pypi.org/project/rpm-version/ (packaging.version being susceptible to fall into RPM-specific traps). 

After all, we have `uv` to pull deps now. It would make sense to have a reliable RPM version comparison in xcp-ng-tests and stop using workarounds.

semarie: I see no `packaging.version.LooseVersion`. Do you mean https://pypi.org/project/looseversion/ ?

stormi: I guess I was thinking about `distutils.version.LooseVersion` but distutils was deprecated. Anyway I think using a module that faithfully implements RPM's comparison logic is the way to go.

semarie:
Last push:
- rebased on master
- add looseversion dependency, and replace the custom `compare_versions` function by comparing LooseVersion2 objects

semarie:
Last push:
- update `requirements/base.txt` after looseversion addition

tests/network/conftest.py
stormi: I see that you used the looversion PyPi package, but what about https://pypi.org/project/rpm-version/ which I mentioned earlier and would actually apply the right comparison rules?

stormi (changes requested):
I'm trying to review the PR, which deserves reaching a point where we can merge it and have it running in CI, but right now most commits are lacking an explanation about the rationale behind them (see https://docs.xcp-ng.org/project/development-process/commit-message-conventions/), and there are important implementation choices made without any explanation that I can find, such as delaying the moment we start checking for rules to apply. Fixed delays are something we avoid, so they must be justified explicitly, and if there's a way to check differently without a delay, this must be preferred.

tests/network/conftest.py
semarie: Ah sorry, I missed it. Yes it is should be possible to use it too.

semarie:
Last push:
- use `rpm_version` (instead of `looseversion`)

semarie:
last push:
- rebased on master
- needs https://github.com/xcp-ng/xcp-ng-tests/pull/679 (for `xo_cli` fix)
- remove change to introduce delay on `wait_for` and (try to) synchronize with delayed operations
  - `vm.start()` and `vm.shutdown()` aren't seen immediately by XO : ask XO about the VM status before continuing
  - `sdnController` operations aren't immediate: wait for the log file to be updated before continuing (still with hardcoded max timeout, as XO doesn't necessary contact all hosts in the pool)
- update some commits messages

bleader (approved):

semarie:
Last push:
- rebase on master
- be more explicit on commit message, about what I needed and what I did for abstractions

jobs.py
```diff
@@ -79,7 +79,7 @@ class JobData(TypedDict):
     "network-advanced": {
         "description": "a group of network tests with complex prerequisites",
         "requirements": [
-            "A pool with at least 1 host.",
+            "A pool with at least 1 host (if more, with same network configuration).",
             "At least 2 free NICs on every host.",
             "A small VM that can be imported on the SRs.",
```
stormi, line 145: It is important to mention here that xo_cli is necessary. This PR changes a situation where only one job used to require XO, and now we have another job that requires XO (unless XO was already required by this job previously but not mentioned at the time in the requirements).

Let's add it here to the list of requirements, and I'll make sure it's clear to the team managing CI that this job has such as requirement.

I don't expect any issue, but they need to know.

stormi: I sent the message. Only remains the need to update the list of requirements.

semarie:
Last push:
- updated commit messages

tests/network/test_traffic_rules.py
```diff
@@ -0,0 +1,849 @@
...
+from lib.sr import SR
+from lib.tunnel import Tunnel
+from lib.vlan import VLAN
+from lib.vm import VM
+from lib.xo import xo_cli
+
+from typing import Callable
+
+# Requirements:
+# xo-cli (on the host running the test) is expected to be usable
```
stormi, line 21: As discussed together, we don't want to add this requirement to the `main` job, so we'll have to move all traffic rules tests, even the simple ones, to the `network-advanced` job.

We discussed a possible implementation for that: make xo_cli a fixture and create a marker named the same, automatically set when the fixture is used. Then use the marker to exclude or include tests in jobs.py.

stormi:
I voided @bleader's review and @glehmann's review so that they may re-review once we've removed the dependency to xo_cli from the `main` job.

jobs.py
semarie: I added XO mention to the jobs entry (in the last commit I pushed).

tests/network/test_traffic_rules.py
semarie: I used the existing `hosts_with_xo` fixture for automatically add a `hosts_with_xo` mark and exclude it from 'main' jobs.

semarie:
Last push:
- new commit
  - automatically add `hosts_with_xo` mark on tests using `hosts_with_xo` fixture
  - exclude tests wih `hosts_with_xo` mark to run in `main` jobs
  - run tests with `hosts_with_xo` mark in `tests/network` in `network-advanced` jobs
  - add a check in `xo_cli()` function to run only on tests using the `hosts_with_xo` fixture

semarie:
Tip for reviewers: by running `uv run ./jobs.py collect network-advanced` and `uv run ./jobs.py collect main`, you could see the tests that will be run (and check that `hosts_with_xo` marker is properly propagated to tests)

semarie:
Last push:
- fix style issues (line too long on markers used by 'main' tests, and import order)

bleader (approved):
Seems to cover the raised concerns in the various threads, should be good to go.

lib/xo.py
```diff
@@ -6,6 +8,20 @@
 
 from typing import Literal, overload
 
+__allow_xo_cli = False
+def _allow_xo_cli(value: bool) -> bool:
+    """
+    Permit to configure the usage of xo_cli function (returns the previous value).
+    This function shoudln't be called directly.
```
stormi: s/shoudln't/shouldn't/

stormi (approved):
Two minor comments, left for fixing after my approval.

In the commit message: "and run them in 'advanced-network' jobs 
instead of (for networks tests)." looks like an unfinished sentence.

The second comment is inline.

semarie:
Last push:
- fix typo in docstring of `_allow_xo_cli` function
- fix phrasing in commit message

semarie:
Last push:
- fix typo in docstring of `_allow_xo_cli` function (for real this time)

semarie:
Last push:
- rebase to master
- fix conflict in `lib/vif.py`
- use `XeParams` type

glehmann (approved):

PR #396, opened 2026-02-12
Use passlib in place of legacycrypt

glehmann:
legacycrypt requires to install libcrypt.so.1 manually

Keep legacycrypt as a dependency for now, to ease the transition to those who already have a custom data.py.

Move hash_password() to lib.common, to make easier to change the implementation in the future.

This is in draft while I test if it actually works with an install test.

1. `master`
2. **"Use passlib in place of legacycrypt" (this PR) → `master`**
3. #431 → #396
4. #491 → #396

### DIFF
--- a/lib/common.py
+++ b/lib/common.py
@@ -18,6 +18,7 @@ from functools import lru_cache
 from uuid import UUID
 
 import requests
+from passlib.hash import sha512_crypt
 from pydantic import TypeAdapter
 
 from typing import (
@@ -398,3 +399,8 @@ def _param_clear(host: Host, xe_prefix: str, uuid: str, param_name: str) -> None
     """ Common implementation for param_clear. """
     args: dict[str, str | bool | dict[str, str]] = {'uuid': uuid, 'param-name': param_name}
     host.xe(f'{xe_prefix}-param-clear', args)
+
+def hash_password(password: str) -> str:
+    """Hash password for /etc/shadow."""
+    # XCP-ng uses sha512 with 5000 rounds by default
+    return sha512_crypt.using(rounds=5000).hash(password)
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -7,8 +7,10 @@ requires-python = ">=3.11"
 dependencies = [
     "cryptography>=3.3.1",
     "gitpython",
+    # TODO: keep legacycrypt to let the user transition to lib.common.hash_password in their data.py
     "legacycrypt",
```
olivierh-pro, line 11: If you remove `hash_password` in `./data.py-dist`, I think this dependency is likely useless, isn't it ?

olivierh-pro (approved):

vxgmichel (approved):
Maybe document the `rounds=5000` since it's not obvious that it's the default value that the system assumes when it's not explicitly provided in the corresponding `/etc/shadow` line.

LGTM otherwise 👍

lib/common.py
```diff
@@ -386,3 +387,7 @@ def _param_clear(host: Host, xe_prefix: str, uuid: str, param_name: str) -> None
     """ Common implementation for param_clear. """
     args: dict[str, str | bool | dict[str, str]] = {'uuid': uuid, 'param-name': param_name}
     host.xe(f'{xe_prefix}-param-clear', args)
+
+def hash_password(password: str) -> str:
+    """Hash password for /etc/password."""
```
vxgmichel:
```suggestion
    """Hash password for /etc/shadow."""
```

conftest.py
```diff
@@ -22,9 +22,7 @@
     is_uuid,
     parse_size,
     prefix_object_name,
-    setup_formatted_and_mounted_disk,
     shortened_nodeid,
-    teardown_formatted_and_mounted_disk,
```
ydirson, line 28: this does not seem to belong to this commit

ydirson (changes requested):
other than that remark, lgtm

conftest.py
glehmann: fixed!

ydirson (approved):

PR #398, opened 2026-02-13
ssh: only allow string command

glehmann:
List of strings was previously accepted, but was giving the false feeling
that we could avoid the command being parsed by the remote shell.
ssh doesn't support passing separated parameters. It reconstructs a string
with the arguments it gets from the command line separated by a single space.
This string is then passed to the ssh servers that passes it to the shell.

With this change, the string passed to the ssh() function is the exact
string received by the shell on the remote host.

This PR is part of a stack containing 9 PRs:

1. `master`
2. #387
3. **"ssh: only allow string command" (this PR)**
4. #399
5. #400
6. #401
7. #402
8. #403
9. #460
10. #476


lib/vm.py
```diff
@@ -55,10 +55,10 @@ def is_paused(self):
         return self.power_state() == 'paused'
 
     # `on` can be an host name-label or UUID
-    def start(self, on=None):
+    def start(self, on: str | bool | None = None) -> str:
```
semarie: side remark regarding ssh refactor: does `bool` is required here ? I agree that `args` could accept a `bool`, it seems to me that `on=` has no sense with a `bool`

semarie (approved):
globally LGTM (so the approve).
just few remarks that might be addressed but nothing blocking from my point of vue.

PR #431, opened 2026-03-13
Add Docker-based compatibility test kit for external validation

glehmann:
Introduces a containerized test kit that enables external teams to easily validate xcp-ng compatibility on their own hardware without complex setup requirements.
The compat_kit module provides a straightforward, self-contained environment for running compatibility tests, lowering the barrier to entry for third-party validation.

1. `master`
2. #396 → `master`
3. **"Add Docker-based compatibility test kit for external validation" (this PR) → #396**
4. #491 → #396


compat_kit/entrypoint.py
```diff
@@ -0,0 +1,689 @@
...
+    RESET = "\033[0m"
+
+    def format(self, record: logging.LogRecord) -> str:
+        color = self.COLORS.get(record.levelno, "")
+        original_levelname = record.levelname
+        if color:
+            record.levelname = f'{color}{record.levelname}{self.RESET}'
+        result = super().format(record)
+        record.levelname = original_levelname
+        return result
```
olivierh-pro, line 56: Are you sure for this workflow ? The order of assignment looks like you never reach a colored level.

compat_kit/entrypoint.py
```diff
@@ -0,0 +1,689 @@
...
+
+# flake8: noqa: E402
+sys.path.append(f'{os.path.abspath(os.path.dirname(__file__))}/..')
+from lib.commands import local_cmd
+from lib.host import Host
+from lib.pool import Pool
+
+from typing import Optional
+
+class ColoredFormatter(logging.Formatter):
```
olivierh-pro, line 36: Would you consider adding this class in specific module, so i can use it too in my `scripts/tools.py` ?

glehmann: Sure! I'll let you move it to a shared module once you'll use it, if you don't mind :)

olivierh-pro (approved):

casasnovas (approved):
Code lgtm and I've tested it on my xoa instance and it worked fine as per the instructions!

.github/workflows/docker.yml
```diff
@@ -0,0 +1,56 @@
+name: Build and Push Docker Image to GHCR
+
+on:
+  push:
+    branches:
+      - 'master'
```
rzr, line 6: We might want to stabilize it with tags soon or later

.github/workflows/docker.yml
```diff
@@ -0,0 +1,56 @@
...
+  pull_request:
+
+jobs:
+  compat-kit:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write # Required to push packages to GHCR
+    steps:
+      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8
```
rzr: nitpicking: pining hash is a good practice, but I tent to add versions in comments then can track what should be updated (or maybe dependabot can figure this out)

.github/workflows/docker.yml
```diff
@@ -0,0 +1,56 @@
...
+
+on:
+  push:
+    branches:
+      - 'master'
+  pull_request:
+
+jobs:
+  compat-kit:
+    runs-on: ubuntu-latest
```
rzr, line 11: I would avoid a floating base, but it seems a gh  common practice since AFAIK they are dropping support of older versions :/

.github/workflows/docker.yml
```diff
@@ -0,0 +1,56 @@
...
+          REF: ${{ github.ref }}
+          EVENT_NAME: ${{ github.event_name }}
+          REPOSITORY: ${{ github.repository }}
+          EVENT_NUMBER: ${{ github.event.number }}
+        run: |
+          if [[ "$REF" == "refs/heads/master" ]]; then
+            echo "tags=ghcr.io/$REPOSITORY/compat-kit:latest" >> $GITHUB_OUTPUT
+            echo "push=true" >> $GITHUB_OUTPUT
+          elif [[ "$EVENT_NAME" == "pull_request" ]]; then
+            echo "tags=ghcr.io/$REPOSITORY/compat-kit:pr-$EVENT_NUMBER" >> $GITHUB_OUTPUT
```
rzr, line 39: I would rather use a git describe id or just hash to avoid mixing up PR reversions

.github/workflows/docker.yml
```diff
@@ -0,0 +1,56 @@
...
+          else
+            echo "push=false" >> $GITHUB_OUTPUT
+          fi
+          echo "created=$(date -u +'%Y-%m-%dT%H:%M:%SZ')" >> $GITHUB_OUTPUT
+      - uses: docker/build-push-action@263435318d21b8e681c14492fe198d362a7d2c83
+        with:
+          context: ./
+          file: ./compat_kit/Dockerfile
+          push: ${{ steps.meta.outputs.push }}
+          tags: ${{ steps.meta.outputs.tags }}
```
rzr, line 50: Any plan to delete merged PRs ?

compat_kit/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM fedora:43
+RUN dnf install -y --setopt=install_weak_deps=False \
+    curl \
+    git \
+    iputils \
+    netcat \
```
rzr, line 6:
```suggestion
    netcat \
    make \
```

compat_kit/entrypoint.py
```diff
@@ -0,0 +1,687 @@
...
+                    remove_ssh_key(host)
+                except Exception as e:
+                    logging.warning(f"Failed to remove key from {host}: {e}")
+
+    # Always clean up temp directory, even if skip_cleanup is True
+    if state.temp_dir and state.temp_dir.exists():
+        try:
+            shutil.rmtree(state.temp_dir)
+        except Exception as e:
+            logging.warning(f"Failed to remove temporary directory {state.temp_dir}: {e}")
```
rzr, line 168: If there is any security concerns, I would trigger an error here and ask user to make sure there is no leftovers in

compat_kit/entrypoint.py
```diff
@@ -0,0 +1,687 @@
...
+    # Append the key to authorized_keys using sshpass
+    # This avoids the interactive prompts that ssh-copy-id causes
+    local_cmd([
+        'sshpass',
+        '-p', password,
+        'ssh',
+        '-o', 'StrictHostKeyChecking=no',
+        '-o', 'UserKnownHostsFile=/dev/null',
+        f'root@{host}',
+        f"echo '{public_key}' >> ~/.ssh/authorized_keys && chmod 600 ~/.ssh/authorized_keys",
```
rzr, line 235: Why not use ssh-copy-id instead ?

compat_kit/entrypoint.py
```diff
@@ -0,0 +1,687 @@
...
+    Raises:
+        ValueError: If pool constraint is violated
+    """
+    num_hosts = len(pool.hosts)
+
+    if second_host is None:
+        if num_hosts != 1:
+            raise ValueError(
+                f"Pool constraint violated: Master host's pool contains {num_hosts} host(s), "
+                f"but only 1 host was provided. The pool must contain only the specified hosts. "
```
rzr:
```suggestion
                f"but only 1 host was provided. The pool must contain only the specified host. "
```

unclear to me, is it "only  one host expected in pool" ?

compat_kit/entrypoint.py
```diff
@@ -0,0 +1,687 @@
...
+                    f"Pool constraint violated: Second host is already in master's pool, "
+                    f"but the pool contains {num_hosts} host(s). Only 2 hosts are expected "
+                    f"(master and second). Please eject the extra hosts before running the compatibility kit."
+                )
+        else:
+            # Second host not in pool - master must be alone
+            if num_hosts != 1:
+                raise ValueError(
+                    f"Pool constraint violated: Master host's pool contains {num_hosts} host(s), "
+                    f"but only the master host should be present before joining the second host. "
```
rzr:
```suggestion
                    f"but only the master host should be present before the second host is joining the pool. "
```

compat_kit/entrypoint.py
```diff
@@ -0,0 +1,687 @@
...
+
+        assert process.stdout is not None
+        with open(log_path, "wb") as f:
+            # Read from the pipe until the process finishes
+            for line in iter(process.stdout.readline, b''):
+                # 1. Write the original colorful line to the actual terminal
+                sys.stdout.buffer.write(line)
+                sys.stdout.buffer.flush()
+
+                # 2. Strip the codes and write the clean text to the file
```
rzr, line 384: question: Wouldn't it be better to make color output optional in the first place ?

compat_kit/entrypoint.py
```diff
@@ -0,0 +1,687 @@
...
+        return
+
+    test_args = [
+        'tests/storage/ext/test_ext_sr_intrapool_migration.py',
+        '-m', 'hostA2 and not hostB1',
+    ]
+    run_pytest(2, test_args, "test_kit_2.log")
+
+
+def set_hvm_fep(enabled: bool) -> None:
```
rzr, line 470: Would set default to True

compat_kit/entrypoint.py
```diff
@@ -0,0 +1,687 @@
...
+        try:
+            set_hvm_fep(False)
+            if not state.dry_run:
+                assert state.pool is not None
+                state.pool.master.reboot(verify=True)
+        except Exception as e:
+            logging.warning(f"Failed to disable hvm_fep and reboot: {e}")
+
+
+def print_summary() -> None:
```
rzr, line 513: suggestion: would add default param to make colors/utf-8 optional

compat_kit/entrypoint.py
```diff
@@ -0,0 +1,687 @@
...
+        run_phase_1()
+        run_phase_2()
+        run_phase_3()
+        print_summary()
+    except KeyboardInterrupt:
+        logging.error("Interrupted by user")
+        sys.exit(1)
+    except Exception as e:
+        logging.error(f"Error: {e}", exc_info=True)
+        sys.exit(1)
```
rzr: would use a different exit code

compat_kit/entrypoint.py
```diff
@@ -0,0 +1,687 @@
...
+
+
+def main() -> None:
+    """Parse arguments and run the compatibility test kit."""
+    default_vm_url = "https://nextcloud.vates.tech/index.php/s/MmEjo8qYo7Ccs2B/download"
```
rzr, line 616: is it public ? if not this has to be changed to something stable over time. or it has to be generated on the fly (With a script? ie vagrant?)

compat_kit/entrypoint.py
```diff
@@ -0,0 +1,687 @@
...
+    parser = argparse.ArgumentParser(description="XCP-ng Compatibility Test Kit - Automated testing on new hardware")
+    parser.add_argument('--master-host', help="IP or hostname of the pool master")
+    parser.add_argument("--second-host", help="IP or hostname of the second host (optional, for pool tests)")
+    parser.add_argument("--password", help="SSH root password (if not provided, will be prompted)")
+    parser.add_argument("--log-dir", default="/app/logs",
```
rzr: I would rather use a relative path (in HOME?) to allow out of docker use.

compat_kit/id_ed25519
```diff
@@ -0,0 +1,7 @@
+-----BEGIN OPENSSH PRIVATE KEY-----
```
rzr, line 1: Can't we put this in a secret ? maybe not if images have to be shared, I would generate it on the fly (like vm?) if possible, or add a comment that/why it is intentional and there is no security impact

compat_kit/README.md
```diff
@@ -0,0 +1,83 @@
+# XCP-ng Server Compatibility Test Kit
+
+## Requirements
+
+- **Docker** installed on your machine (or Docker Desktop for macOS/Windows).
+- **Hosts to test:**
+  - At least one x86_64 server with XCP-ng installed (the release you want to test),
+    [updated with the latest updates](https://xcp-ng.org/docs/updates.html).
+  - A disk for the system, and a separate, empty disk for testing.
+  - Optionally, a second identical host. If provided, will be automatically
```
rzr:
```suggestion
  - Optionally, a second identical host. If provided, host will be automatically
```

compat_kit/README.md
```diff
@@ -0,0 +1,83 @@
...
+    joined to the first to form a pool for storage migration and live migration
+    tests. **Both hosts must have the same SSH root password.**
+  - Internet access to download the test image.
+  - SSH access available (root user; password-based authentication).
+- **Time:** The test suite execution might take up to an hour, depending on the hardware.
+
+## Security Notes
+
+The test VM is configured with a known access key and is **strictly reserved for testing purposes only**.
```
rzr, line 19:
```suggestion
The test VM is configured with a known access key and is **strictly reserved for testing purposes only**.

Test VM should not be exposed to untrusted networks.
```

rzr (commented):
lgtm, I work prefer that we have a re-producible vm instead of a fixed asset. Has this been considered ? there is a risk it could be blocked in user's infra.

.github/workflows/docker.yml
glehmann: In practice, it's not changing that often, so we could use a tagged version.
We are using `ubuntu-latest` everywhere else though, so this change might be better suited for another PR :)

.github/workflows/docker.yml
glehmann: Yes, I think we should eventually

.github/workflows/docker.yml
glehmann: The idea is to only keep the image built with the last version of the PR, and to be able to do some cleanup based on the image tags, like removing the image when a PR is merged, …

.github/workflows/docker.yml
glehmann: yes!

compat_kit/Dockerfile
glehmann: why `make`? I don't think we need it in the image

compat_kit/entrypoint.py
glehmann: No there is no security concern, and the directory will be destroyed anyway when exiting the container.
This is more an informative message for the developer

compat_kit/entrypoint.py
glehmann: I tried, but I couldn't make it work with sshpass

compat_kit/entrypoint.py
glehmann: The pool must contain only the hosts provided by the user, or the script would fail.
I'll try to reformulate

compat_kit/entrypoint.py
glehmann: For what use case?

compat_kit/entrypoint.py
glehmann: yes it is. I agree we should use something stable—ideally on the same GH repository—but we don't have it yet

compat_kit/id_ed25519
glehmann: This must match the key in the VM image. I would prefer to generate it on the fly, but I don't think we have a way to start a VM and specify an authorized ssh key for now, as we would with cloudinit

compat_kit/Dockerfile
rzr: Well there are a makefile helper in repo, i thought we could use it to collect tasks that can be useful for developers (like the setup steps in the docker files)

compat_kit/id_ed25519
rzr: yea i figured this out, I would just add a comment with a reference to the image (and explicitly state it's not secure) file can be also renamed with unsecure suffix.

rzr, line 1:
```suggestion

This key is unsecure, it should be only used to connect to prebuild-demo vm images.

-----BEGIN OPENSSH PRIVATE KEY-----
```

rzr (approved):
Please consider the key comment (we never know) we could create a issue for removal or this key when a better solution is in place.

PR #437, opened 2026-03-26
storage: test large volumes

glehmann:

1. `master`
2. #436 → `master`
3. **"storage: test large volumes" (this PR) → #436**
4. #447 → #437
5. #446 → #447
6. #449 → #446
7. #450 → #449
8. #452 → #450
9. #453 → #452
10. #454 → #453
11. #461 → #454
12. #464 → #461
13. #470 → #464
14. #471 → #470
15. #481 → #471
16. #523 → #481
17. #497 → #481
18. #498 → #497
19. #500 → #498
20. #509 → #500

### DIFF
--- a/lib/common.py
+++ b/lib/common.py
@@ -40,6 +40,25 @@ MiB = KiB**2
 GiB = KiB**3
 TiB = KiB**4
 
+def parse_size(size_str: str) -> int:
+    """
+    Parse a size string like "2.5TiB", "1GiB" or "1024".
+    """
+    try:
+        return int(size_str)
+    except ValueError:
+        pass
+
+    size_str = size_str.strip()
+    for unit, multiplier in [('TiB', TiB), ('GiB', GiB), ('MiB', MiB), ('KiB', KiB)]:
+        if size_str.endswith(unit):
+            try:
+                return int(float(size_str[:-len(unit)].strip()) * multiplier)
+            except ValueError:
+                pass
+
+    raise ValueError(f"Cannot parse size: {size_str}")
+
 T = TypeVar("T")
 
 HostAddress: TypeAlias = str
--- a/tests/storage/conftest.py
+++ b/tests/storage/conftest.py
@@ -2,6 +2,11 @@ from __future__ import annotations
 
 import pytest
 
+import logging
+
+from lib import config
+from lib.common import randid
+from lib.host import Host
 from lib.vm import VM
 from tests.storage import install_randstream
 
@@ -19,3 +24,25 @@ def pytest_collection_modifyitems(config: pytest.Config, items: list[pytest.Item
 def storage_test_vm(running_unix_vm: VM) -> Generator[VM, None, None]:
     install_randstream(running_unix_vm)
     yield running_unix_vm
+
+@pytest.fixture()
+def temp_large_dir(host: Host) -> Generator[str, None, None]:
+    """Create a temporary directory on a large (NFS) disk on the primary host"""
+    temp_id = randid()
+    nfs_conf = config.sr_device_config("NFS_DEVICE_CONFIG")
+    if 'serverpath' in nfs_conf or 'server' in nfs_conf:
+        mount_path = f'/mnt/nfs_{temp_id}'
+        host.ssh(f'mkdir -p {mount_path}')
+        path = f'{mount_path}/temp_large_dir_{temp_id}'
+        host.ssh(f'mount -t nfs {nfs_conf["server"]}:{nfs_conf["serverpath"]} {mount_path}')
+        host.ssh(f'mkdir {path}')
+        yield path
+        host.ssh(f'rm -rf {path}')
+        host.ssh(f'umount {mount_path}')
+        host.ssh(f'rmdir {mount_path}')
+    else:
+        logging.warning("No NFS configuration found, using /tmp")
+        path = f'/tmp/temp_large_dir_{temp_id}'
+        host.ssh(f'mkdir {path}')
+        yield path
+        host.ssh(f'rm -rf {path}')
--- a/tests/storage/storage.py
+++ b/tests/storage/storage.py
@@ -4,8 +4,9 @@ import pytest
 
 import logging
 
+from lib import config
 from lib.commands import SSHCommandFailed
-from lib.common import Defer, GiB, strtobool, wait_for, wait_for_not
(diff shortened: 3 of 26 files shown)


lib/common.py
```diff
@@ -42,6 +42,25 @@
 GiB = KiB**3
 TiB = KiB**4
 
+def parse_size(size_str: str) -> int:
```
stormi, line 43: First commit: there are many things changed in this first commit and so little explained in the commit message. How volume sizes are handled, how we're supposed to use them, what had to be changed for this (I haven't seen everything yet, but I saw a randstream function added somewhere and it would be easier if the commit message would explain why), implementation choices (how are the disks generated and why)...

Also, I don't see explanations about the use of `--volume-size` in the README, where it would probably be useful.

tests/storage/storage.py
```diff
@@ -179,38 +180,58 @@ def install_randstream(vm: 'VM'):
...
+
+    The args string should contain the command and arguments to pass to randstream,
+    e.g. "generate /dev/xvdb".
+    """
+    output = vm.ssh(f'randstream -v {args}')
+    for line in output.splitlines():
+        if line.startswith('checksum: '):
+            return line.split(": ")[1].strip()
+    raise Exception(f"Could not find the checksum in the randstream output:\n{output}")
+
```
stormi, line 195: This looks like a change which would belong in a separate commit.

tests/storage/conftest.py
```diff
@@ -17,3 +24,25 @@ def pytest_collection_modifyitems(config, items):
 def storage_test_vm(running_unix_vm: VM):
     install_randstream(running_unix_vm)
     yield running_unix_vm
+
+@pytest.fixture(scope='package')
+def temp_large_dir(host: Host) -> Generator[str, None, None]:
```
stormi, line 29: Second commit : I don't understand the reason for the commit, the commit message doesn't tell. Stopping review here. I think this PR is a draft not flagged as such, a branch you use for the tests that you currently run under timing pressure, but not meant for review yet.

glehmann: With bigger vdis the exported image may not fit in /tmp. This fixture offers a way to have access to a temporary large directory that can be used for that, and probably for other usages.
The name felt clear enough to me, especially in the context of that PR, but I can add it to the commit description.

lib/common.py
glehmann: The randstream function was mostly added to keep the lines short enough to stay in the 120-character limit.

tests/storage/conftest.py
stormi: My comment is not just about the fixture. It applies to the commit itself, but I put it here so that it's attached to a part of the commit, as github doesn't allow commenting commit messages.

glehmann: I've updated the commit description—hopefully it's clearer now :)

stormi: Perfect :+1:

stormi (approved):

tests/storage/storage.py
```diff
@@ -219,36 +219,57 @@ def coalesce_integrity(vm: VM, vdi: VDI, vdi_op: CoalesceOperation, defer: Defer
...
+    randstream(vm, f'validate --expected-checksum {checksum} /root/data')
     vm.shutdown(verify=True)
 
-    xva_path = f'/tmp/{vm.uuid}.xva'
+    xva_path = f'{temp_large_dir}/{vm.uuid}.xva'
     defer(lambda: vm.host.ssh(f'rm -f {xva_path}'))
     vm.export(xva_path, compression)
     # check that the zero blocks are not part of the result. Most of the data is from the random stream, so
-    # compression has little effect. We just check the result is between 500 and 700 MiB
+    # compression has little effect. We just take into account the system sie
```
stormi: s/sie/size/

stormi (changes requested):
Sorry to be demanding, but commit 4 lacks a description of the changes, explaining how it works.

tests/storage/zfsvol/test_zfsvol_sr.py
stormi, line 1: Fifth commit: 

> Fix remaing hardcoded device name
The tests was using a hardcoded xvdb which worked because it matches the
default naming.

=> 

> Fix remaining hardcoded device name
The test was using a hardcoded xvdb which worked because it matches the
default naming.

stormi (approved):

tests/storage/storage.py
```diff
@@ -222,36 +222,57 @@ def coalesce_integrity(vm: VM, vdi: VDI, vdi_op: CoalesceOperation, defer: Defer
...
     vm.start()
     vm.wait_for_vm_running_and_ssh_up()
     install_randstream(vm)
 
-    # 500MiB, so we have some data to check and some empty spaces in the exported image
-    # TODO: resize the vm partitions and write more data in the disk (like 1/2 of the disk size)
-    #       to ensure we can export and import large xva images
-    randstream(vm, 'generate --size 500MiB /root/data')
-    randstream(vm, 'validate --expected-checksum 24e905d6 /root/data')
+    is_alpine = vm.ssh_with_result('apk --version').returncode == 0
```
stormi, line 238: You could use the method exposed by VM to get the package manager.

tests/storage/storage.py
```diff
@@ -222,36 +222,57 @@ def coalesce_integrity(vm: VM, vdi: VDI, vdi_op: CoalesceOperation, defer: Defer
...
+        vm.ssh('apk add gawk util-linux e2fsprogs-extra')
+        vm.ssh('wget https://raw.githubusercontent.com/canonical/cloud-utils/main/bin/growpart -O /usr/bin/growpart')
+        vm.ssh('chmod +x /usr/bin/growpart')
+        # TODO: maybe use `findmnt -no SOURCE /` from util-linux to get the blockdevice mounted on /
+        growpart_returncode = vm.ssh_with_result('growpart /dev/xvda 3').returncode
+        assert growpart_returncode in [0, 1] # growpart returns 1 if the size is already the expected one
+        vm.ssh('resize2fs /dev/xvda3')
+        stream_size = volume_size // 2
+    else:
+        stream_size = 500 * MiB
```
stormi, line 251: Let's improve the VM object so that it can grow partitions on a wider variety of OSes (RHEL-like, debian...)

Lankou66 (approved):
LGTM

Millefeuille42 (approved):

PR #438, opened 2026-03-26
storage: Snapshot the VM multiple times before exporting/importing

last-genius:
Adds a new with_snapshot parameter which, if enabled, snapshots the VM
twice before the export/import cycle.

This tests for the bug we are seeing now with QCOW2, where the full
chain of snapshots is not considered.

Leave holes in each snapshot by writing to independent sections to verify
all the snapshot's bitmaps are consulted.

---

This test passes with the current pre-build of xapi: [xapi-26.1.3-1.6~fixsnapexport.6.xcpng8.3](https://koji.xcp-ng.org/buildinfo?buildID=5455)

### DIFF
--- a/tests/storage/ext/test_ext_sr.py
+++ b/tests/storage/ext/test_ext_sr.py
@@ -92,8 +92,9 @@ class TestEXTSR:
 
     @pytest.mark.small_vm
     @pytest.mark.parametrize("compression", ["none", "gzip", "zstd"])
-    def test_xva_export_import(self, vm_on_ext_sr: VM, compression: XVACompression) -> None:
-        xva_export_import(vm_on_ext_sr, compression)
+    @pytest.mark.parametrize("with_snapshot", [True, False])
+    def test_xva_export_import(self, vm_on_ext_sr: VM, compression: XVACompression, with_snapshot: bool) -> None:
+        xva_export_import(vm_on_ext_sr, compression, with_snapshot)
 
     @pytest.mark.small_vm
     def test_vdi_export_import(self, storage_test_vm: VM, ext_sr: SR, image_format: ImageFormat) -> None:
--- a/tests/storage/ext/test_ext_sr.py
+++ b/tests/storage/ext/test_ext_sr.py
@@ -74,6 +74,10 @@ class TestEXTSR:
         xva_export_import(vm_on_ext_sr, compression, temp_large_dir, defer)
 
     @pytest.mark.small_vm
+    def test_xva_export_import_with_snapshot(self, vm_on_ext_sr: VM, temp_large_dir: str, defer: Defer) -> None:
+        xva_export_import(vm_on_ext_sr, 'zstd', temp_large_dir, defer, with_snapshot=True)
+
+    @pytest.mark.small_vm
     def test_vdi_export_import(self, storage_test_vm: VM, ext_sr: SR, image_format: ImageFormat, temp_large_dir: str,
                                defer: Defer) -> None:
         vdi_export_import(storage_test_vm, ext_sr, image_format, temp_large_dir, defer)
--- a/tests/storage/lvm/test_lvm_sr.py
+++ b/tests/storage/lvm/test_lvm_sr.py
@@ -149,8 +149,9 @@ class TestLVMSR:
 
     @pytest.mark.small_vm
     @pytest.mark.parametrize("compression", ["none", "gzip", "zstd"])
-    def test_xva_export_import(self, vm_on_lvm_sr: VM, compression: XVACompression) -> None:
-        xva_export_import(vm_on_lvm_sr, compression)
+    @pytest.mark.parametrize("with_snapshot", [True, False])
+    def test_xva_export_import(self, vm_on_lvm_sr: VM, compression: XVACompression, with_snapshot: bool) -> None:
+        xva_export_import(vm_on_lvm_sr, compression, with_snapshot)
 
     @pytest.mark.small_vm
     def test_vdi_export_import(self, storage_test_vm: VM, lvm_sr: SR, image_format: ImageFormat) -> None:
--- a/tests/storage/lvm/test_lvm_sr.py
+++ b/tests/storage/lvm/test_lvm_sr.py
@@ -142,6 +142,10 @@ class TestLVMSR:
         xva_export_import(vm_on_lvm_sr, compression, temp_large_dir, defer)
 
     @pytest.mark.small_vm
+    def test_xva_export_import_with_snapshot(self, vm_on_lvm_sr: VM, temp_large_dir: str, defer: Defer) -> None:
+        xva_export_import(vm_on_lvm_sr, 'zstd', temp_large_dir, defer, with_snapshot=True)
+
+    @pytest.mark.small_vm
     def test_vdi_export_import(self, storage_test_vm: VM, lvm_sr: SR, image_format: ImageFormat, temp_large_dir: str,
                                defer: Defer) -> None:
         vdi_export_import(storage_test_vm, lvm_sr, image_format, temp_large_dir, defer)
--- a/tests/storage/storage.py
+++ b/tests/storage/storage.py
@@ -211,6 +211,7 @@ def xva_export_import(vm: VM, compression: XVACompression, with_snapshot: bool) 
     vm.ssh("randstream generate -v --size 500MiB /root/data")
     vm.ssh("randstream validate -v --expected-checksum 24e905d6 /root/data")
 
+    snap1, snap2 = None, None
     if with_snapshot:
         snap1 = vm.snapshot()
         # Write new data to a particular sector after taking a snapshot
@@ -238,8 +239,9 @@ def xva_export_import(vm: VM, compression: XVACompression, with_snapshot: bool) 
         if imported_vm is not None:
             imported_vm.destroy()
         vm.host.ssh(f'rm -f {xva_path}')
-        if with_snapshot:
+        if snap1:
             snap1.destroy()
+        if snap2:
             snap2.destroy()
 
 def vdi_export_import(vm: VM, sr: SR, image_format: ImageFormat) -> None:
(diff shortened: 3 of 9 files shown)


glehmann (approved):
LGTM
It will conflict with #437

tests/storage/xfs/test_xfs_sr.py
```diff
@@ -80,9 +80,10 @@ def test_invalid_vdi_size(self, xfs_sr: SR, image_format: ImageFormat):
 
     @pytest.mark.small_vm
     @pytest.mark.parametrize("compression", ["none", "gzip", "zstd"])
-    def test_xva_export_import(self, vm_on_xfs_sr: VM, compression: XVACompression, temp_large_dir: str, defer: Defer) \
-            -> None:
-        xva_export_import(vm_on_xfs_sr, compression, temp_large_dir, defer)
+    @pytest.mark.parametrize("with_snapshot", [True, False])
```
stormi: I have some concerns regarding the multiplication of tests due to parametrization combinatory. 

@last-genius Do you think it's relevant to test with snapshot for each compression type? If yes, let's keep it as you've made it. Else, better create a new test_xva_export_import_with_snapshot not parametrized by "compression"?

last-genius: You're right, there's no need to test for each compression type, I've added a new function

tests/storage/ext/test_ext_sr.py
```diff
@@ -73,6 +73,10 @@ def test_xva_export_import(self, vm_on_ext_sr: VM, compression: XVACompression,
             -> None:
         xva_export_import(vm_on_ext_sr, compression, temp_large_dir, defer)
 
+    @pytest.mark.small_vm
+    def test_xva_export_import_with_snapshot(self, vm_on_ext_sr: VM, temp_large_dir: str, defer: Defer) -> None:
+        xva_export_import(vm_on_ext_sr, "none", temp_large_dir, defer, with_snapshot=True)
```
stormi: zstd compression might make the test faster and use less space?

stormi (approved):
Just one non-blocking comment.

tests/storage/ext/test_ext_sr.py
last-genius: Moved to zstd

tests/storage/ext/test_ext_sr.py
```diff
@@ -73,6 +73,10 @@ def test_xva_export_import(self, vm_on_ext_sr: VM, compression: XVACompression,
             -> None:
         xva_export_import(vm_on_ext_sr, compression, temp_large_dir, defer)
 
+    @pytest.mark.small_vm
+    def test_xva_export_import_with_snapshot(self, vm_on_ext_sr: VM, temp_large_dir: str, defer: Defer) -> None:
+        xva_export_import(vm_on_ext_sr, 'none', temp_large_dir, defer, with_snapshot=True)
```
stormi: Did you switch this one back to none on purpose?

stormi (approved):
Updated approval based on latest push.

tests/storage/storage.py
```diff
@@ -285,6 +298,10 @@ def xva_export_import(source_vm: VM, compression: XVACompression, temp_large_dir
     imported_vm.start()
     imported_vm.wait_for_vm_running_and_ssh_up()
     randstream(imported_vm, f'validate --expected-checksum {checksum} /root/data')
+    if snap1:
+        snap1.destroy()
+    if snap2:
+        snap2.destroy()
```
stormi: Sorry for not mentioning this earlier, but I think it would be more robust to use the `defer` fixture, as was done in line 232.

tests/storage/lvm/test_lvm_sr.py
```diff
@@ -139,7 +139,11 @@ def test_invalid_vdi_size(self, lvm_sr: SR, image_format: ImageFormat):
     @pytest.mark.parametrize("compression", ["none", "gzip", "zstd"])
     def test_xva_export_import(self, vm_on_lvm_sr: VM, compression: XVACompression, temp_large_dir: str, defer: Defer) \
             -> None:
-        xva_export_import(vm_on_lvm_sr, compression, temp_large_dir, defer)
+        xva_export_import(vm_on_lvm_sr, compression, temp_large_dir, defer, False)
+
+    @pytest.mark.small_vm
+    def test_xva_export_import_with_snapshot(self, vm_on_lvm_sr: VM, temp_large_dir: str, defer: Defer) -> None:
+        xva_export_import(vm_on_lvm_sr, 'zstd', temp_large_dir, defer, True)
```
Millefeuille42: Please pass booleans as key word arguments instead of positional (see other comment about this for how to enforce it).

tests/storage/storage.py
```diff
@@ -225,7 +225,8 @@ def coalesce_integrity(vm: VM, vdi: VDI, vdi_op: CoalesceOperation, defer: Defer
 
 XVACompression = Literal['none', 'gzip', 'zstd']
 
-def xva_export_import(source_vm: VM, compression: XVACompression, temp_large_dir: str, defer: Defer) -> None:
+def xva_export_import(source_vm: VM, compression: XVACompression, temp_large_dir: str,
+                      defer: Defer, with_snapshot=False) -> None:
```
Millefeuille42: Enforce booleans to be keyword arguments:
```suggestion
def xva_export_import(source_vm: VM, compression: XVACompression, temp_large_dir: str,
                      defer: Defer, *, with_snapshot=False) -> None:
```

Millefeuille42 (changes requested):
One detail + leftover comment from Sam, otherwise OK for me.

tests/storage/lvm/test_lvm_sr.py
stormi: Your comment sounds like a requirement. I understand that it's probably good practice (and it would be good to explain it when making such comments), but is that something we already enforce in this repository's coding practice?

tests/storage/storage.py
last-genius: Done

tests/storage/storage.py
last-genius: Moved to `defer`

tests/storage/lvm/test_lvm_sr.py
Millefeuille42: This is something we're trying to have on storage. However I should have made it a suggestion rather than a requirement, especially here. My bad.

For the context, using `*, ` enforces the next arguments to be used as kwargs only, this helps mainly with boolean arguments to help with clarity and readability.

Millefeuille42 (approved):

tests/storage/storage.py
```diff
@@ -225,10 +226,15 @@ def coalesce_integrity(vm: VM, vdi: VDI, vdi_op: CoalesceOperation, defer: Defer
 
 XVACompression = Literal['none', 'gzip', 'zstd']
 
-def xva_export_import(source_vm: VM, compression: XVACompression, temp_large_dir: str, defer: Defer) -> None:
+def xva_export_import(source_vm: VM, compression: XVACompression, temp_large_dir: str,
+                      defer: Defer, *, with_snapshot=False) -> None:
     # clone the vm, so we can resize the disk without affecting the vm from the fixture
     vm: VM | None = source_vm.clone()
+    snap1: Snapshot | None
+    snap2: Snapshot | None
```
stormi: Unless I'm mistaken, the variables are only type-annoted here, but not initialized.


```suggestion
    snap1: Snapshot | None = None
    snap2: Snapshot | None = None
```

last-genius: Fixed

tests/storage/storage.py
```diff
@@ -260,6 +266,17 @@ def xva_export_import(source_vm: VM, compression: XVACompression, temp_large_dir
...
     randstream(vm, f'validate --expected-checksum {checksum} /root/data')
+
+    if with_snapshot:
+        snap1 = vm.snapshot()
+        # Write new data to a particular sector after taking a snapshot
+        vm.ssh("randstream generate --seed 1 -v --position 300MiB --size 100MiB /root/data")
+        vm.ssh("randstream validate -v --expected-checksum f797ab33 /root/data")
+        snap2 = vm.snapshot()
+        vm.ssh("randstream generate --seed 2 -v --position 100MiB --size 100MiB /root/data")
+        vm.ssh("randstream validate -v --expected-checksum 60db5b9f /root/data")
```
glehmann:
```suggestion
        randstream(vm, "generate --seed 1 --position 300MiB --size 100MiB /root/data")
        randstream(vm, "validate --expected-checksum f797ab33 /root/data")
        snap2 = vm.snapshot()
        randstream(vm, "generate --seed 2 --position 100MiB --size 100MiB /root/data")
        randstream(vm, "validate --expected-checksum 60db5b9f /root/data")
```

To be consistent with the other usages.

tests/storage/storage.py
```diff
@@ -260,6 +266,17 @@ def xva_export_import(source_vm: VM, compression: XVACompression, temp_large_dir
...
+        checksum = '60db5b9f'
```
glehmann: Unfortunately, this won't work if we change the volume size, i.e. `--volume-size=10GiB`.
We can't just use a precomputed checksum anymore.

Also I don't think there is any guarantee that writing to an existing file would actually override its content with ext4, so you might as well write to another file.

A possible option is to always split the file write into 3 steps (and 3 files), independently of the `with_snapshot` value, also validate the three parts independently, and add the snapshots in between the writes when `with_snapshot` is `True`.

glehmann (changes requested):

tests/storage/storage.py
last-genius: Fixed

tests/storage/storage.py
last-genius: Moved to writing to 3 files instead

glehmann (approved):

PR #445, opened 2026-03-30
zfsvol: disable coalesce test

glehmann:
vhd-parent, which is used to detect when the coalesce is done, is
not available for that SR type.

This PR is part of a stack containing 12 PRs:

1. `master`
2. #436
3. #437
4. #447
5. #446
6. **"zfsvol: disable coalesce test" (this PR)**
7. #449
8. #450
9. #452
10. #453
11. #454
12. #461
13. #464

### DIFF
--- a/tests/storage/zfsvol/test_zfsvol_sr.py
+++ b/tests/storage/zfsvol/test_zfsvol_sr.py
@@ -63,6 +63,7 @@ class TestZfsvolVm:
 
     @pytest.mark.small_vm
     @pytest.mark.parametrize("vdi_op", ["snapshot"])  # "clone" requires a snapshot
+    @pytest.mark.skip("zfsvol doesn't provide vhd-parent")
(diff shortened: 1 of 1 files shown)


stormi:
The PR description is that of the first PR in the series, probably a copy and paste mistake?

tests/storage/zfsvol/test_zfsvol_sr.py
```diff
@@ -72,6 +72,7 @@ def test_snapshot(self, vm_on_zfsvol_sr):
 
     @pytest.mark.small_vm
     @pytest.mark.parametrize("vdi_op", ["snapshot"])  # "clone" requires a snapshot
+    @pytest.mark.skip("zfsvol doesn't provide vhd-parent")
```
stormi, line 66: I don't like pytest.skip but I don't have a better alternative to offer right now.

stormi (approved):

PR #449, opened 2026-03-31
storage: Avoid writing the whole device in coalesce tests

glehmann:
This is done by:
* writing in only a small piece of data at the beginning, in the middle
  and at the end of the device
* creating the snapshot/clone
* writing for data in an already used location and a non-used one
* triggering the coalesce
* finally verifying that the data we generated is accessible as expected

Still not necessarily clear to me:
* Are we losing something by not filling the drive?
* How much data needs to be written to the disk? For now the tests is writing 4 chunks of 2GiB of data max, independently of the actual device size.

This PR is part of a tree containing 19 PRs:

1. `master`
2. #436 → `master`
3. #437 → #436
4. #447 → #437
5. #446 → #447
6. **"storage: Avoid writing the whole device in coalesce tests" (this PR) → #446**
7. #450 → #449
8. #452 → #450
9. #453 → #452
10. #454 → #453
11. #461 → #454
12. #464 → #461
13. #470 → #464
14. #471 → #470
15. #481 → #471
16. #523 → #481
17. #497 → #481
18. #498 → #497
19. #500 → #498
20. #509 → #500

PR #447, opened 2026-03-30
host: include mdadm RAID devices in disk detection

glehmann:
This PR is part of a tree containing 19 PRs:

1. `master`
2. #436 → `master`
3. #437 → #436
4. **"host: include mdadm RAID devices in disk detection" (this PR) → #437**
5. #446 → #447
6. #449 → #446
7. #450 → #449
8. #452 → #450
9. #453 → #452
10. #454 → #453
11. #461 → #454
12. #464 → #461
13. #470 → #464
14. #471 → #470
15. #481 → #471
16. #523 → #481
17. #497 → #481
18. #498 → #497
19. #500 → #498
20. #509 → #500

PR #450, opened 2026-03-31
storage: avoid writing the whole device in migration tests

glehmann:

1. `master`
2. #436 → `master`
3. #437 → #436
4. #447 → #437
5. #446 → #447
6. #449 → #446
7. **"storage: avoid writing the whole device in migration tests" (this PR) → #449**
8. #452 → #450
9. #453 → #452
10. #454 → #453
11. #461 → #454
12. #464 → #461
13. #470 → #464
14. #471 → #470
15. #481 → #471
16. #523 → #481
17. #497 → #481
18. #498 → #497
19. #500 → #498
20. #509 → #500

### DIFF
--- a/tests/storage/storage.py
+++ b/tests/storage/storage.py
@@ -53,7 +53,7 @@ def cold_migration_then_come_back(vm: VM, prov_host: Host, dest_host: Host, dest
     vdi_name: str | None = None
     integrity_check = not vm.is_windows
     dev = ""
-    checksum = ""
+    checksums = ('', '', '')
 
     if integrity_check:
         # the vdi will be destroyed with the vm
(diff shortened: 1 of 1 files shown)


olivierh-pro (approved):

stormi:
Why?

(I assume it's for speed of execution, but it's not described)

IMO, live migration is something that we must also test with large disks with actual data (and, in the future, CPU/RAM and/or IO activity in the VM), not just some data and mostly zeroes. If we don't, then we won't test in production-like conditions. 

This whole series of PRs meant to reduce the amount of data after we've increased the size of the disks may need further planning work before implementing (part of this series of afterwards) when we get some air, as we must define the right balance. This very series would also have benefitted by a design review prior to implementing, but I understand the timing constraints.

stormi (approved):

Millefeuille42 (approved):

Lankou66 (approved):

PR #452, opened 2026-03-31
xva/vdi: only write a small data amount in large volumes for faster tests

glehmann:

1. `master`
2. #436 → `master`
3. #437 → #436
4. #447 → #437
5. #446 → #447
6. #449 → #446
7. #450 → #449
8. **"xva/vdi: only write a small data amount in large volumes for faster tests" (this PR) → #450**
9. #453 → #452
10. #454 → #453
11. #461 → #454
12. #464 → #461
13. #470 → #464
14. #471 → #470
15. #481 → #471
16. #523 → #481
17. #497 → #481
18. #498 → #497
19. #500 → #498
20. #509 → #500

### DIFF
--- a/tests/storage/storage.py
+++ b/tests/storage/storage.py
@@ -274,7 +274,7 @@ def xva_export_import(vm: VM, compression: XVACompression, temp_large_dir: str, 
         growpart_returncode = vm.ssh_with_result('growpart /dev/xvda 3').returncode
         assert growpart_returncode in [0, 1] # growpart returns 1 if the size is already the expected one
         vm.ssh('resize2fs /dev/xvda3')
-        stream_size = volume_size // 2
+        stream_size = min(volume_size // 2, 2 * GiB)
     else:
         stream_size = 500 * MiB
 
--- a/tests/storage/storage.py
+++ b/tests/storage/storage.py
@@ -311,16 +311,8 @@ def vdi_export_import(vm: VM, sr: SR, image_format: ImageFormat, temp_large_dir:
     defer(lambda: vm.disconnect_vdi(vdi_src) if vdi_src is not None and vdi_src.uuid in vm.vdis else None)
     dev = f'/dev/{vbd.param_get("device")}'
 
-    # the stream is 1/5 of the full one, truncated to a multiple of 32KiB, in order to
-    # be validable in a single command
-    stream_size = (config.volume_size // 5 // (32 * KiB)) * (32 * KiB)
-    stream_position = (config.volume_size // 2)
-
-    checksum1 = randstream(vm, f'generate --size {stream_size} {dev}')
-    # use a different seed to not write the same data (default seed is 0)
-    checksum2 = randstream(vm, f'generate --seed 1 --position {stream_position} --size {stream_size} {dev}')
-    randstream(vm, f'validate --size {stream_size} --expected-checksum {checksum1} {dev}')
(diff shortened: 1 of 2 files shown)


tests/storage/zfsvol/test_zfsvol_sr.py
```diff
@@ -91,51 +91,7 @@ def test_vdi_export_import(
...
-        vm.host.xe('vdi-import', {'uuid': vdi_dest.uuid, 'filename': image_path, 'format': image_format})
-        vbd = vm.connect_vdi(vdi_dest)
-        defer(lambda: vm.disconnect_vdi(vdi_dest))
-        dev = f'/dev/{vbd.param_get("device")}'
-
-        randstream(vm, f'validate --size {stream_size} --expected-checksum {checksum1} {dev}')
-        randstream(
-            vm, f'validate --position {stream_position} --size {stream_size} --expected-checksum {checksum2} {dev}'
-        )
+        vdi_export_import(vm, sr, image_format, temp_large_dir, defer)
```
stormi, line 92: Please describe why there was a special case before and why it isn't needed anymore.

tests/storage/storage.py
```diff
@@ -271,7 +271,7 @@ def xva_export_import(vm: VM, compression: XVACompression, temp_large_dir: str,
         growpart_returncode = vm.ssh_with_result('growpart /dev/xvda 3').returncode
         assert growpart_returncode in [0, 1] # growpart returns 1 if the size is already the expected one
         vm.ssh('resize2fs /dev/xvda3')
-        stream_size = volume_size // 2
+        stream_size = min(volume_size // 2, 2 * GiB)
```
stormi, line 277: IMO, this cap, as well as all the other caps that you put into place to speed up the tests, should be configurable, up to even having a way to disable them and write the full disk. 

I don't have a good feeling about never testing with full disks. 2 GiB of data inside a 16 TB disk, it's just one 8000th of the disk. It's not the same tests anymore.

tests/storage/zfsvol/test_zfsvol_sr.py
glehmann: I don't know why it wasn't working before. I just noticed the exported image size is as expected, and that we don't need to maintain a special case anymore.
Maybe @xcp-ng/storage can provide more details?

tests/storage/storage.py
glehmann: I've added a full disk test in a follow-up commit, and I was thinking to add tests with full disk write for export/import.
Do you prefer the flag option to change the size?

stormi: If we don't have such a flag, with the current code, we'll not test, for example live migration with lots of data. Whatever the chosen path (flag, additional tests), we need a plan for additional slow tests that manipulate large disks with a lot of data actually written.

Using a flag looks like a possible way, but there are probably others.

tests/storage/zfsvol/test_zfsvol_sr.py
stormi: Ok, let's not make it a blocking comment of mine.

stormi (approved):

dinhngtu:
It seems exclusively storage-related so I've removed the H&K review request. You might want to check the codeowners file?

Millefeuille42 (approved):

Lankou66 (approved):

PR #453, opened 2026-03-31
storage: test full device write

glehmann:
With most test now writing only a small amount of data to the devices,
we need these new test to validate that we can write on the whole
device.

1. `master`
2. #436 → `master`
3. #437 → #436
4. #447 → #437
5. #446 → #447
6. #449 → #446
7. #450 → #449
8. #452 → #450
9. **"storage: test full device write" (this PR) → #452**
10. #454 → #453
11. #461 → #454
12. #464 → #461
13. #470 → #464
14. #471 → #470
15. #481 → #471
16. #523 → #481
17. #497 → #481
18. #498 → #497
19. #500 → #498
20. #509 → #500

### DIFF
--- a/tests/storage/ext/test_ext_sr.py
+++ b/tests/storage/ext/test_ext_sr.py
@@ -16,6 +16,7 @@ from tests.storage import (
     ImageFormat,
     XVACompression,
     coalesce_integrity,
+    full_vdi_write,
     try_to_create_sr_with_missing_device,
     vdi_export_import,
     vdi_is_open,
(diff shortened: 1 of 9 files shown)


stormi:
> With most test now writing only a small amount of data to the devices,
we need these new test to validate that we can write on the whole
device.

~~It's good to tell us in the PR, but the commit message also should tell that.~~ Oops, it does. Looks like I don't know how to use GH at this time of the day.

tests/storage/ext/test_ext_sr.py
```diff
@@ -103,6 +104,10 @@ def test_vdi_export_import(
     ):
         vdi_export_import(storage_test_vm, ext_sr, image_format, temp_large_dir, defer)
 
+    @pytest.mark.small_vm
```
stormi, line 105: Back when I created `small_vm`, I also created `big_vm`, indicating that we should also run the same test with a big VM (but we never actually used that in jobs) and `multi_vm` for tests that should run on a variety of VMs.

It's not exactly appropriate anymore, since these storage tests start from a small VM and grow it to the required sizee now, but the naming became confusing. We'll have to think about its evolution at some point.

stormi (approved):

Lankou66 (approved):

Millefeuille42 (approved):

PR #455, opened 2026-04-01
Tools/update - add nested mark in inventory

olivierh-pro:
Add `nested = true|false` in inventory to prepare snapshot creation.

This new "nested" mark in inventory allow me to differentiate physical hosts from nested hosts to perform snapshot creation accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -584,6 +584,8 @@ For each pool target :
 2. Get attached secondary hosts of the pool
   * Repeat step `1.` for each secondary
 
+It is possible to mark target as *"nested"* (XCP-ng inside VM). *See inventory example below*
+
 **Inventory file**
 
 `update` command can read an inventory file in [TOML v1.0.0](https://toml.io/en/v1.0.0) format:
@@ -603,6 +605,7 @@ Take a look at an example inventory file:
 # my_inventory.toml
 
 [all]
+nested = true
 enablerepos = ["xcp-ng-base"]
 
 [servers]
--- a/lib/tools/inventory.py
+++ b/lib/tools/inventory.py
@@ -7,9 +7,9 @@ from pathlib import Path
 
 from lib.common import HostAddress
 
-def load_inventory(inventory_path: Path) -> dict[str, dict[str, list[str]]]:
+def load_inventory(inventory_path: Path) -> dict[str, dict[str, list[str] | bool]]:
     """Create an inventory object from loaded inventory file."""
-    inventory: dict[str, dict[str, list[str]]] = {}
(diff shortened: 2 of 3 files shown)


lib/tools/inventory.py
```diff
@@ -18,15 +18,20 @@ def load_inventory(inventory_path: Path) -> dict[str, dict[str, list[str]]]:
...
         host = {
+            "nested": nested,
             "enablerepos": repos or all.get("enablerepos", [])
         }
         inventory[server] = host
 
     return inventory
 
-def into_inventory(hosts: list[HostAddress], enablerepos: list[str]) -> dict[HostAddress, dict[str, list[str]]]:
+def into_inventory(hosts: list[HostAddress], enablerepos: list[str], nested: bool) -> dict[HostAddress, dict[str, list[str]]]:
```
rzr: may set nested to default value

olivierh-pro: `nested` is set to False by default. It does not appear as default here but the implementation set this variable to False.

lib/tools/inventory.py
```diff
@@ -18,23 +18,31 @@ def load_inventory(inventory_path: Path) -> dict[str, dict[str, list[str]]]:
...
     Basically, it is used as compatibility when we don't want inventory from file.
     """
-    inventory: dict[HostAddress, dict[str, list[str]]] = {}
+    inventory: dict[HostAddress, dict[str, list[str] | bool]] = {}
 
     for h in hosts:
-        host = {
+        host: dict[str, list[str] | bool] = {
+            "nested": nested or False,
             "enablerepos": enablerepos or [],
```
rzr, line 46: I would sort alphabatically (e before n)

README.md
```diff
@@ -603,6 +605,7 @@ Take a look at an example inventory file:
 # my_inventory.toml
 
 [all]
+nested = true
```
rzr, line 608: same

rzr (approved):

glehmann:
> Add `nested = true|false` in inventory to prepare snapshot creation.
> 
> This new "nested" mark in inventory allow me to differentiate physical hosts from nested hosts to perform snapshot creation accordingly.

But we'll need more info than just `nested` to create the snapshot, so maybe it would be simpler to just define the host it's on?

IMO, this PR on its own is too small to understand where this is going. Maybe include an actual feature in the PR, or link it to another PR that implements the feature?

olivierh-pro:
Discussed with @glehmann (related to [this](https://github.com/xcp-ng/xcp-ng-tests/pull/455#issuecomment-4307343223)). I will merge this work with another existing PR (#468).

I close this PR

PR #457, opened 2026-04-02
tools/update: move update_host into Host class

olivierh-pro:
Instead of an external function, update should live inside Host class.

With this PR, I move function `update_host` from module `lib/tools/update.py` into class `Host`. IMO, this change fits well the existing "object model" in `lib/`.

### DIFF
--- a/lib/host.py
+++ b/lib/host.py
@@ -450,6 +450,26 @@ class Host:
 
         return self.ssh(base_command)
 
+    def update(self, enablerepos: list[str] = [], reboot: bool = True):
+        """Updates current host.
+
+        An helper function that wraps update tasks on current host.
+
+        :param list[str] enablerepos:
+            Repositories to enable when updating.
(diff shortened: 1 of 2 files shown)


lib/host.py
```diff
@@ -450,6 +450,23 @@ def yum_update(self, enablerepos: List[str] = []):
...
+
+        :param list[str] enablerepos:
+            Repositories to enable when updating.
+        """
+        logging.info(f"[{self}] Updating...")
+
+        self.yum_clean_metadata()
+        self.yum_update(enablerepos=enablerepos)
+        # Everything's ok, just reboot
+        self.reboot(verify=True)
```
```suggestion
        if reboot:
            self.reboot(verify=True)
```

And add it as default param of function

rzr (approved):

glehmann (approved):

stormi (approved):

PR #459, opened 2026-04-03
tools/update - add a new arg to handle hosting pool (physical)

olivierh-pro:
This PR changes the structure of the "python dict inventory".

Move all hosts inside a subkey named `hosts`.

```python
# before
{'host-1': ..., 'host-2': ...}

# after
{'hosts': {'host-1': ..., 'host-2': ...}}
```
Add a new cli arg to provide the address of hosting pool's master host (physical): `-P/--hosting-pool`

**What is the hosting pool's master host?**

This is the host where live my nested hosts. In a nested context, "nested hosts" are VM which are hosted by an Host (a physical one). This information is needed to perform a snapshot (incoming PR #468).

image

* Orange: my own nested lab. My hosts (which are VM in real life) must be snapshotted.
* Blue: The Real host where my hosts live. This is the hosting pool of my hosts (A1, A2, B1) and **this is where I can perform the snapshot** (in nested context).

This is the reason why I've added the key `hosting_pool` in inventory. I needed a starting point to perform a snapshot.

### DIFF
--- a/README.md
+++ b/README.md
@@ -673,6 +673,7 @@ Take a look at an example inventory file:
 
 [all]
 repositories = ["xcp-ng-base"]
+hosting_pool = "A"
 
 [hosts]
 
@@ -681,12 +682,13 @@ repositories = ["xcp-ng-base"]
 [hosts."ip_or_hostname-2"]
 
 repositories = ["xcp-ng-updates"]
+hosting_pool = "B"
 ```
 
 > [!IMPORTANT]
-> Config values under `servers` override values under `all`. For instance, the above inventory would produce
+> Config values under `hosts` override values under `all`. For instance, the above inventory would produce
 > the following python dict:
 >
-> `{'ip_or_hostname-1': {'repositories': ['xcp-ng-base']}, 'ip_or_hostname-2': {'repositories': ['xcp-ng-updates']}}`
+> `{'ip_or_hostname-1': {'repositories': ['xcp-ng-base'], 'hosting_pool': 'A'}, 'ip_or_hostname-2': {'repositories': ['xcp-ng-updates'], 'hosting_pool': 'B'}}`
 >
 > Using *enablerepo flag* `-e` with inventory is still possible, it won't be used though.
--- a/README.md
+++ b/README.md
@@ -671,7 +671,7 @@ Take a look at an example inventory file:
 ```toml
 # my_inventory.toml
 
-[all]
+[default]
 repositories = ["xcp-ng-base"]
 hosting_pool = "A"
 
@@ -686,9 +686,10 @@ hosting_pool = "B"
 ```
 
 > [!IMPORTANT]
-> Config values under `hosts` override values under `all`. For instance, the above inventory would produce
+> * `default` is applied to all hosts
+> * Config values under `hosts` override values under `default`. For instance, the above inventory would produce
 > the following python dict:
 >
 > `{'ip_or_hostname-1': {'repositories': ['xcp-ng-base'], 'hosting_pool': 'A'}, 'ip_or_hostname-2': {'repositories': ['xcp-ng-updates'], 'hosting_pool': 'B'}}`
 >
-> Using *enablerepo flag* `-e` with inventory is still possible, it won't be used though.
+> * Using *enablerepo flag* `-e` with inventory is still possible, it won't be used though.
--- a/README.md
+++ b/README.md
@@ -641,17 +641,17 @@ uv run scripts/tools.py -h
 This command performs an update operation on remote targets.
 
 ```bash
-uv run scripts/tools.py update -H primary1 primary2
+uv run scripts/tools.py update -H master1 master2
 ```
 
 For each pool target :
 
-1. Update master (primary) host of the pool:
+1. Update master host of the pool:
   * Clean cached metadata
   * Update with repository manager (yum): Optionally enables repositories
   * Reboot
-2. Get attached secondary hosts of the pool
-  * Repeat step `1.` for each secondary
+2. Get other hosts of the pool
+  * Repeat step `1.` for each host
 
 **Inventory file**
 
@@ -692,4 +692,4 @@ hosting_pool = "B"
 >
 > `{'ip_or_hostname-1': {'repositories': ['xcp-ng-base'], 'hosting_pool': 'A'}, 'ip_or_hostname-2': {'repositories': ['xcp-ng-updates'], 'hosting_pool': 'B'}}`
 >
-> * Using *enablerepo flag* `-e` with inventory is still possible, it won't be used though.
+> * When `--inventory` flag is present, repos passed to `-e` flag won't be considered.
--- a/lib/pool.py
+++ b/lib/pool.py
@@ -17,7 +17,7 @@ from typing import Any, Callable, Iterable
 class Pool:
     """Pool
 
-    :raises NotAMasterHostError: if initial host is not a master (primary)
+    :raises NotAMasterHostError: if initial host is not a master
     """
     xe_prefix = "pool"
 
@@ -300,4 +300,4 @@ class Pool:
         return self.master.xe('network-list', {'name-label': network_name}, minimal=True)
 
 class NotAMasterHostError(Exception):
-    """Host must be a master (primary)."""
+    """Host must be a master."""
--- a/lib/tools/cli.py
+++ b/lib/tools/cli.py
@@ -17,7 +17,7 @@ def _command_update(args: argparse.Namespace) -> None:
     if args.inventory:
         inventory = load_inventory(args.inventory)
     else:
-        inventory = into_inventory(args.hosts, args.repos)
+        inventory = into_inventory(args.hosts, args.repos, args.hosting_pool)
 
     update_pools(inventory)
 
@@ -52,6 +52,12 @@ def cli() -> None:
         action="append",
         dest="repos",
         help="repositories to enable when updating",
+    )
+    subparser_cmd_update.add_argument(
+        "-P",
+        "--hosting-pool",
+        type=HostAddress,
+        help="Address (hostname|ip) of hosting pool's master host (nested context)",
     )
     subparser_cmd_update.set_defaults(func=_command_update)
 
--- a/lib/tools/inventory.py
+++ b/lib/tools/inventory.py
@@ -13,25 +13,29 @@ class HostConfig(TypedDict):
     repositories: list[str]
 
 
-Inventory: TypeAlias = dict[HostAddress, HostConfig]
+HostConfigs: TypeAlias = dict[HostAddress, HostConfig]
+
+class Inventory(TypedDict):
+    hosts: HostConfigs
 
 
 def load_inventory(inventory_path: Path) -> Inventory:
     """Create an inventory object from loaded inventory file."""
-    inventory: Inventory = {}
-
     with open(inventory_path, "rb") as f:
         data = tomllib.load(f)
 
     all = data.get("all", {})
     hosts = data.get("hosts", [])
 
-    for server, config in hosts.items():
+    inventory_hosts: HostConfigs = {}
+    for h, config in hosts.items():
         repos = config.get("repositories", [])
         host: HostConfig = {"repositories": repos or all.get("repositories", [])}
-        inventory[server] = host
+        inventory_hosts[h] = host
 
-    return inventory
+    return {
+        "hosts": inventory_hosts,
+    }
 
 
 def into_inventory(hosts: list[HostAddress], repositories: list[str]) -> Inventory:
@@ -39,10 +43,11 @@ def into_inventory(hosts: list[HostAddress], repositories: list[str]) -> Invento
 
     Basically, it is used as compatibility when we don't want inventory from file.
     """
-    inventory: Inventory = {}
-
+    inventory_hosts: HostConfigs = {}
     for h in hosts:
         host: HostConfig = {"repositories": repositories or []}
-        inventory[h] = host
+        inventory_hosts[h] = host
 
-    return inventory
+    return {
+        "hosts": inventory_hosts,
+    }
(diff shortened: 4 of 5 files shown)


stormi:
@olivierh-pro you should describe in the PR what the concept of a parent host means. I also see a commit without a description which would deserve such an explanation: [tools/cli: add new arg for parent host](https://github.com/xcp-ng/xcp-ng-tests/pull/459/changes/29508d63885f84dbdfb3b2d07d1f45f9319651c0)

Last, I see so many force-pushes to this PR that I never know if we must review it or it's still evolving. **Update**: but seeing as the last notification was asking the platform team for review, now I know :)

olivierh-pro:
@stormi I've added more information in PR and commit.

For the "force push spam", two possible explanations:
* I constantly integrate feedback from my reviewers into the current branch
* If the branch in part of "stacked branches", I always rebase current branch on previous work to be up to date

README.md
```diff
@@ -601,6 +601,7 @@ Take a look at an example inventory file:
 
 ```toml
 # my_inventory.toml
+parent = "ip_or_hostname" # master host
```
rzr: question: master or primary ? One should be decided and stick

lib/tools/inventory.py
```diff
@@ -13,36 +13,43 @@ class HostConfig(TypedDict):
...
 
-    return inventory
+    return {
+        "hosts": inventory_hosts,
+        "parent": data.get("parent", None),
+    }
 
 
-def into_inventory(hosts: list[HostAddress], repositories: list[str]) -> Inventory:
+def into_inventory(hosts: list[HostAddress], repositories: list[str], parent: HostAddress) -> Inventory:
```
rzr: I still find the name confusing,  since the return type has changed 

suggestion: I would make constructor since you introduced a class type.

Same for the load method

lib/tools/cli.py
```diff
@@ -53,6 +53,12 @@ def cli() -> None:
         dest="repos",
         help="repositories to enable when updating",
     )
+    subparser_cmd_update.add_argument(
+        "-P",
+        "--parent-host",
+        type=HostAddress,
+        help="Address (hostname|ip) of the parent master host (works with '--nested')",
```
rzr: suggestion: I would group  --nested  and --parent with a single option --nested-parent

lib/tools/tasks/update.py
```diff
@@ -23,24 +24,25 @@ def update_pools(inventory: dict) -> None:
...
         for p in pools:
-            executor.submit(p.master.update, inventory[p.master.hostname_or_ip]["enablerepos"])
+            executor.submit(p.master.update, inventory_hosts[p.master.hostname_or_ip]["repositories"])
 
     # secondary hosts
     with ThreadPoolExecutor() as executor:
         for p in pools:
             # omit first item because it is a primary (master)
-            for h in p.hosts[1:]:
+            for secondary in p.hosts[1:]:
```
rzr: if you use secondary here, please use primary elsewhere,  then the other s  word can be removed from codebase for consistency

README.md
olivierh-pro: Actually, neither master nor primary.

The snapshot operation can be performed either on master(primary) or on secondary, even if we usually perform `xe commands` on a master, it is possible.

The concept of "parent" here represents an artifact (an entity) where my VM (nested host) can live. Please, take a look at my diagram on PR description.

Maybe the word "parent" is a bad choice but I don't have enough inspiration to name it correctly.

Would you suggest something else?

lib/tools/inventory.py
olivierh-pro: I am very confortable with the idea of introducing a new class here. My motivation is to handle easily typechecking.

At the beginning, Inventory was a dict and typing a dict is (IMO) a plague. So, I wrapped it in a Custom type.

In my opinion, it is efficient enough at the moment.

lib/tools/cli.py
olivierh-pro: `--nested` arg is not introduced yet. It is a the next PR (#468).

So I will rework nested arg in next PR.

lib/tools/tasks/update.py
olivierh-pro: What alternative to "secondary" would you suggest?

A host that is not a master is a secondary.

lib/tools/cli.py
rzr: I am suggesting to rename "--parent-host" to "--nested-parent" (and remove the mention of "--nested" in help because If I understand correctly it's duplicating with the "--nested-parent" , no nested without parents right ?

lib/tools/tasks/update.py
rzr: secondary is fine here , but there is master/slave? elsewhere in the code, so I would use primary  (in place of master) to align the secondary you are introducing with this line.

README.md
rzr: My comment was about the "# master host" comment, I would use primary/secondary everywhere (and make sure we don't have trace of master/slave) but this has to be decided for once and applied to the whole code base (maybe a ticket to be created)

lib/tools/cli.py
olivierh-pro: `nested-parent` makes me think of that is the parent which is nested.

What do you think about `controller` instead ? So we can give explanation below:

> A snapshot targets a VM (nested host) and is performed on the controller, which see the VM in its scope

lib/tools/tasks/update.py
olivierh-pro: I prefer to use "master (primary)" since the word "master" is used everywhere in XCP-ng context (for example, in xapi or orchestra)

lib/tools/cli.py
rzr: yes, controller is a good name, we need to make clear it only for nested

lib/tools/tasks/update.py
rzr: so for the logic if you use master then you have to use slave.. , because secondary without primary is confusing IMHO

lib/tools/cli.py
ydirson: Yes selecting a name not easy. "nest" is what had settled with (with option `--nest` in pytest)

IMO "controller" is too generic, it could apply to anything (like "manager")

lib/pool.py
```diff
@@ -26,7 +26,7 @@ def __init__(self, master_hostname_or_ip: HostAddress) -> None:
         if not master.is_master():
             raise NotAMasterHostError(f"Host {master_hostname_or_ip} is not a master host. Pool not created.")
         self.master = master
-        self.hosts = [master]
+        self.hosts: list[Host] = [master]
```
ydirson: That type would seem to be perfectly derived by type-checkers already, is it really useful?

README.md
ydirson: Linked to my previous comment, `parent` at this place in the inventory seems to assume we'll only ever manage nested-in-xcp-ng hosts.  It would be good to be specify it in a more future-proof way, so we don't have to change the inventory syntax in incompatible ways in the near future.

lib/tools/tasks/update.py
```diff
@@ -36,13 +36,13 @@ def update_pools(inventory: Inventory) -> None:
 
     with ThreadPoolExecutor() as executor:
         for p in pools:
-            executor.submit(p.master.update, inventory_hosts[p.master.hostname_or_ip]["enablerepos"])
+            executor.submit(p.master.update, inventory_hosts[p.master.hostname_or_ip]["repositories"])
```
ydirson, line 40: IIUC this is a fix because da07780e2838cff5c604341c3b4503df1321de1f ("add a typed inventory") made an incomplete change wrt that key. It would be good to explain and point to that commit.
As a side note, that commit claims to be about something else entirely.

README.md
ydirson: "pool master" would be the most descriptive given the current naming in the code. When switching to "primary/secondary" I fear we'd have to switch to the more verbose "pool primary host". Yuck.

lib/tools/inventory.py
ydirson: I'm not sure a ctor would be the best (there can be only one), but it's customary to add class methods acting as specialized ctors.

Also, even though I've not used them much, I wonder if using dataclasses for all those custom data types

ydirson (changes requested):

lib/tools/tasks/update.py
olivierh-pro: There is already a commit to fix this : 6e46b93fe85c95d87dabfa36e15f4e2a6c82e9a1

Note: the above commit SHA may change with a rebase operation.

lib/tools/inventory.py
olivierh-pro: I don't see any reason to use Dataclasses here. My need has always been to type a dict.

In that case, we can also mention Pydantic too.

lib/tools/tasks/update.py
ydirson: Yes it is that commit I'm commenting, suggesting to add the details about the commit it fixes.

lib/pool.py
olivierh-pro: Yes it is. We have seen with @glehmann that the typehint in IDE was incorrect when trying to access this attribute.

```python
secondaries = my_pool.hosts[1:] # this is supposed to be inferred `list[str]` but it is not
```

So yes, it need this to improve typehint in IDE. :)

lib/tools/tasks/update.py
olivierh-pro: Done! I've added more information into commit message.

README.md
olivierh-pro: Done. Added a new commit to remove old occurrences of "secondary" or "primary".

Chosen couple: master/slave.

stormi: Although we don't hunt for every occurrence of word `slave` in the code, we tend to avoid adding more occurrences. The only mention I can find in the tests is a place where it's related to the actual name of a XAPI plugin. 

Nowadays, "pool master" and "pool coordinator" (not "primary". Coordinator, that's what XAPI and XenServer use, and it's appropriate) are equivalent for us.

They also replaced "pool slave" with "pool member", which is not the most unambiguous wording they could have chosen, so that's not one we use in docs without adding "other" before it: "other pool members". We tend to use either "other hosts/pool members" or "secondary hosts/pool members". Secondary pool member is already present in README.md.

Please don't use "slave". I suggest `master` (as `coordinator` is not used in this repo yet, contrarily to `master`) / `secondary` here

rzr: I am skeptical about using secondary without primary.

see https://github.com/xcp-ng/xcp-ng-tests/pull/459#discussion_r3199593027

glehmann (approved):

lib/tools/tasks/update.py
rzr: I still find confusing to use master/secondary pair, but let's move on, I am approving until a better naming is decided project wide.

rzr (approved):

lib/pool.py
olivierh-pro: @ydirson good for you?

lib/tools/tasks/update.py
ydirson: It is good to mention the commit (as `da07780e2838cff5c604341c3b4503df1321de1f (tools/update: add a typed inventory)' ) so it can be looked up quickly (e.g. `gitk` will hyperlink it)

lib/pool.py
ydirson: If this is specific to one tool, it should be made explicit:
- in the commit message, which should always make it obvious why a change is done
- in a comment mentioning the tool and version, so we can get an idea while readying the code when it will be acceptable to remove that workaround

README.md
```diff
@@ -617,9 +617,10 @@ hosting_pool = "B"
...
 > [!IMPORTANT]
-> Config values under `hosts` override values under `all`. For instance, the above inventory would produce
+> * `default` is applied to all hosts
+> * Config values under `hosts` override values under `default`. For instance, the above inventory would produce
 > the following python dict:
 >
 > `{'ip_or_hostname-1': {'repositories': ['xcp-ng-base'], 'hosting_pool': 'A'}, 'ip_or_hostname-2': {'repositories': ['xcp-ng-updates'], 'hosting_pool': 'B'}}`
 >
-> Using *enablerepo flag* `-e` with inventory is still possible, it won't be used though.
+> * Using *enablerepo flag* `-e` with inventory is still possible, it won't be used though.
```
ydirson: Although not directly linked to the PR contents, that wording feels awkward to me, and can likely be improved while we're touching this area.  Does it mean that when `-e` is passed it is simply ignored?

lib/tools/tasks/update.py
```diff
@@ -18,7 +18,7 @@ def update_pools(inventory: Inventory) -> None:
 
         Every non-master hosts in inventory will be ignored
 
-    *Update master hosts declared in inventory first, then, update secondary hosts attached to each master.*
+    *Update pools' master hosts declared in inventory first, then, update other hosts for each pool.*
```
ydirson: "Update each pool's master host" could be a smoother read

ydirson (changes requested):
submitted a few other change requests in individual comments

lib/tools/tasks/update.py
olivierh-pro: Done!

README.md
olivierh-pro: Done!

lib/pool.py
olivierh-pro: Done!

I've removed this change because for any reason I  cannot explain, typechecking is OK.

lib/tools/tasks/update.py
olivierh-pro: Done!

README.md
ydirson: I'm OK with the reformulation, but that new commit should be squashed so we don't have 2 changes successive of the same text in the PR

lib/tools/tasks/update.py
ydirson: Please don't use short sha1's, we cannot predict when they become ambiguous

olivierh-pro: Done

README.md
olivierh-pro: Done

ydirson (approved):

PR #460, opened 2026-04-07
misc type related enhancements

glehmann:
See the commit descriptions!

1. `master`
2. **"misc type related enhancements" (this PR)**
3. #476

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,11 +1,11 @@
 # Makefile to run static code checkers locally
 # Equivalent to the GitHub Actions workflow in .github/workflows/code-checkers.yml
 
-.PHONY: all mypy pyright ruff ruff-fix flake8 autopep8 autopep8-fix
+.PHONY: all pyright ruff ruff-fix flake8 autopep8 autopep8-fix
 
 # By default, only run the non-fix version of the hooks so it doesn't modify any files
 # It runs on all files managed by git (untracked files are not checked)
-check: ruff autopep8 flake8 mypy pyright
+check: ruff autopep8 flake8 pyright
 
 # Use the `fix` directive to let autopep8 auto-format the code and ruff sort the imports
 # It runs on all files managed by git (untracked files are not modified)
@@ -14,7 +14,7 @@ fix: ruff-fix autopep8-fix
 data.py:
 	@test -r data.py || echo "File 'data.py' does not exist. Refer to https://github.com/xcp-ng/xcp-ng-tests#configuration." && exit 1
 
-mypy pyright ruff: data.py
+pyright ruff: data.py
 
-ruff ruff-fix autopep8 autopep8-fix flake8 mypy pyright:
+ruff ruff-fix autopep8 autopep8-fix flake8 pyright:
 	uv run prek -a $@
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,10 @@ fix: ruff-fix autopep8-fix
 data.py:
 	@test -r data.py || echo "File 'data.py' does not exist. Refer to https://github.com/xcp-ng/xcp-ng-tests#configuration." && exit 1
 
-pyright ruff: data.py
+vm_data.py:
+	@test -r vm_data.py || echo "File 'vm_data.py' does not exist. Refer to https://github.com/xcp-ng/xcp-ng-tests#configuration." && exit 1
+
+pyright ruff: data.py vm_data.py
 
 ruff ruff-fix autopep8 autopep8-fix flake8 pyright:
 	uv run prek -a $@
--- a/conftest.py
+++ b/conftest.py
@@ -2,12 +2,12 @@ from __future__ import annotations
 
 import pytest
 
-import argparse
 import dataclasses
 import itertools
 import logging
 import os
 import tempfile
+from argparse import Action, ArgumentParser, Namespace
 from collections import defaultdict
 
 import git
@@ -44,7 +44,7 @@ from lib.xo import xo_cli
 # need to import them in the global conftest.py so that they are recognized as fixtures.
 from pkgfixtures import formatted_and_mounted_ext4_disk, sr_disk_wiped
 
-from typing import Any, Dict, Generator, Iterable, List, Optional
+from typing import Any, Dict, Generator, Iterable, List, Optional, Sequence
 
 # Do we cache VMs?
 try:
@@ -53,8 +53,9 @@ except ImportError:
     CACHE_IMPORTED_VM = False
 assert CACHE_IMPORTED_VM in [True, False]
 
-class SplitCommaAction(argparse.Action):
-    def __call__(self, parser, namespace, values, option_string=None):
+class SplitCommaAction(Action):
+    def __call__(self, parser: ArgumentParser, namespace: Namespace, values: str | Sequence[Any] | None,
+                 option_string: str | None = None) -> None:
         items = getattr(namespace, self.dest, None)
         if items is None:
             items = []
--- a/conftest.py
+++ b/conftest.py
@@ -44,7 +44,7 @@ from lib.xo import xo_cli
 # need to import them in the global conftest.py so that they are recognized as fixtures.
 from pkgfixtures import formatted_and_mounted_ext4_disk, sr_disk_wiped
 
-from typing import Any, Dict, Generator, Iterable, List, Optional, Sequence
+from typing import Any, Dict, Generator, Iterable, Sequence
 
 # Do we cache VMs?
 try:
@@ -171,7 +171,7 @@ def pytest_generate_tests(metafunc: pytest.Metafunc) -> None:
 
 # Used to group tests together whenever possible and limit parametrized fixture
 # "context switching" (needless teardown and setup of SRs, for example)
-SCHEDULING_AXES: List[str] = [
+SCHEDULING_AXES: list[str] = [
     "image_format",
 ]
 
@@ -224,7 +224,7 @@ def pytest_collection_modifyitems(items: list[pytest.Item], config: pytest.Confi
     # -----------------------------------------------------
     # Build execution matrix: axis -> leaf package -> items
     # -----------------------------------------------------
-    axis_ordering: Dict[str | None, int] = defaultdict(int)
+    axis_ordering: dict[str | None, int] = defaultdict(int)
     # "None" gets the same order value as the first real axis, on purpose,
     # so that we may better retain initial test order.
     # For example, if we start with this test order:
@@ -254,7 +254,7 @@ def pytest_collection_modifyitems(items: list[pytest.Item], config: pytest.Confi
         grouped[axis_order][package].append(item)
 
     # Flatten back to a list of items
-    new_items: List[pytest.Item] = [
+    new_items: list[pytest.Item] = [
         item
         for axis_order in sorted(grouped) # apply axis_ordering here
         for package in grouped[axis_order]
@@ -269,7 +269,7 @@ def pytest_collection_modifyitems(items: list[pytest.Item], config: pytest.Confi
 # FIXME we may have to move this into lib/ if fixtures in sub-packages
 # want to make use of this feature
 
-PHASE_REPORT_KEY = pytest.StashKey[Dict[str, pytest.TestReport]]()
+PHASE_REPORT_KEY = pytest.StashKey[dict[str, pytest.TestReport]]()
 @pytest.hookimpl(wrapper=True, tryfirst=True)
 def pytest_runtest_makereport(
     item: pytest.Item, call: pytest.CallInfo[Any]
--- a/conftest.py
+++ b/conftest.py
@@ -54,7 +54,7 @@ except ImportError:
 assert CACHE_IMPORTED_VM in [True, False]
 
 class SplitCommaAction(Action):
-    def __call__(self, parser: ArgumentParser, namespace: Namespace, values: str | Sequence[Any] | None,
+    def __call__(self, parser: ArgumentParser, namespace: Namespace, values: str | Sequence[str] | None,
                  option_string: str | None = None) -> None:
         items = getattr(namespace, self.dest, None)
         if items is None:
--- a/conftest.py
+++ b/conftest.py
@@ -174,7 +174,7 @@ SCHEDULING_AXES: List[str] = [
     "image_format",
 ]
 
-def get_axis(item: pytest.Item) -> Optional[str]:
+def get_axis(item: pytest.Item) -> str | None:
     callspec = getattr(item, "callspec", None)
     if callspec is None:
         return None
@@ -223,7 +223,7 @@ def pytest_collection_modifyitems(items: list[pytest.Item], config: pytest.Confi
     # -----------------------------------------------------
     # Build execution matrix: axis -> leaf package -> items
     # -----------------------------------------------------
-    axis_ordering: Dict[Optional[str], int] = defaultdict(int)
+    axis_ordering: Dict[str | None, int] = defaultdict(int)
     # "None" gets the same order value as the first real axis, on purpose,
     # so that we may better retain initial test order.
     # For example, if we start with this test order:
--- a/jobs.py
+++ b/jobs.py
@@ -604,7 +604,11 @@ BROKEN_TESTS = [
 ]
 
 VmDef = str | tuple[str, str]
-VMSDef = dict[str, dict[str, VmDef | list[VmDef]]]
+
+
+class VMSDef(TypedDict):
+    single: dict[str, VmDef]
+    multi: dict[str, list[VmDef]]
 
 # Returns the vm filename or None if a host_version is passed and matches the one specified
 # with the vm filename in vm_data.py. ex: ("centos6-32-hvm-created_8.2-zstd.xva", "8\.2\..*")
@@ -642,7 +646,7 @@ def get_vm_or_vms_refs(handle: str, host_version: str | None = None) -> str | li
 
     VMS = cast(VMSDef, VMS_untyped)
     category, key = handle.split("/")
-    if category not in VMS or key not in VMS[category]:
+    if category not in ("single", "multi") or key not in VMS[category]:
         print(f"ERROR: Could not find VMS['{category}']['{key}'] in vm_data.py, or it's empty.")
         print("You need to update your local vm_data.py.")
         print("You may also bypass this error by providing your own --vm parameter(s).")
--- a/lib/basevm.py
+++ b/lib/basevm.py
@@ -2,13 +2,14 @@ from __future__ import annotations
 
 import logging
 
-from typing import TYPE_CHECKING, List, Literal, overload
+from lib.common import _param_add, _param_clear, _param_get, _param_remove, _param_set
+from lib.sr import SR
+
+from typing import TYPE_CHECKING, Literal, overload
 
 if TYPE_CHECKING:
     from lib.host import Host
 
-from lib.common import _param_add, _param_clear, _param_get, _param_remove, _param_set
-from lib.sr import SR
 
 class BaseVM:
     """ Base class for VM and Snapshot. """
@@ -61,7 +62,7 @@ class BaseVM:
     def _disk_list(self) -> str:
         raise NotImplementedError()
 
-    def vdi_uuids(self, sr_uuid: str | None = None) -> List[str]:
+    def vdi_uuids(self, sr_uuid: str | None = None) -> list[str]:
         output = self._disk_list()
         if output == '':
             return []
--- a/lib/commands.py
+++ b/lib/commands.py
@@ -10,7 +10,7 @@ import tempfile
 import lib.config as config
 from lib.netutil import wrap_ip
 
-from typing import TYPE_CHECKING, Generic, List, Literal, TypeVar, overload
+from typing import TYPE_CHECKING, Generic, Literal, TypeVar, overload
 
 if TYPE_CHECKING:
     from lib.common import HostAddress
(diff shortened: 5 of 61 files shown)


olivierh-pro (approved):

rzr (approved):

rzr (commented):
Can the checks be fixed in this PR ?

Lankou66 (approved):

Makefile
```diff
@@ -8,7 +8,10 @@ all: ruff autopep8 flake8 mypy pyright
 data.py:
 	@test -r data.py || echo "File 'data.py' does not exist. Refer to https://github.com/xcp-ng/xcp-ng-tests#configuration." && exit 1
 
-mypy pyright ruff: data.py
+vm_data.py:
+	@test -r vm_data.py || echo "File 'vm_data.py' does not exist. Refer to https://github.com/xcp-ng/xcp-ng-tests#configuration." && exit 1
+
+mypy pyright ruff: data.py vm_data.py
```
ydirson, line 20: Actually, the files we want to check here are the `-dist` ones

jobs.py
ydirson, line 654: This feels strange. It looks like a type assertion, but after a cast to the type we then assert.  Also, doing a type cast on `category` makes it "have 2 types", which is not very clear (and I recall seeing rejected by type-checkers in the not-so-distant past)

conftest.py
```diff
@@ -51,8 +51,9 @@
     CACHE_IMPORTED_VM = False
 assert CACHE_IMPORTED_VM in [True, False]
 
-class SplitCommaAction(argparse.Action):
-    def __call__(self, parser, namespace, values, option_string=None):
+class SplitCommaAction(Action):
+    def __call__(self, parser: ArgumentParser, namespace: Namespace, values: str | Sequence[Any] | None,
+                 option_string: str | None = None) -> None:
```
ydirson, line 58: `Sequence[any]` cannot be `Sequence[str]`?

tests/install/conftest.py
```diff
@@ -6,19 +6,16 @@
 import logging
 import os
 import tempfile
-import xml.etree.ElementTree as ET
```
ydirson, line 9: This and `Any` are not part of the `TYPE_CHECKING` suppression. I'm surprised they did not get reported as "unused import" originally, any idea why?

lib/efi.py
```diff
@@ -24,7 +24,7 @@
 from typing import Any, Iterable, Literal, Self, cast
 
 class _EfiGlobalTempdir:
-    _instance = None
+    _instance: TemporaryDirectory[str] | None = None
```
ydirson, line 27: This one actually gets the kind of typing this commit removes, making it look suspicious.

lib/efi.py
```diff
@@ -440,7 +440,6 @@ def auth(self) -> str:
     def sign_with(self, signer: Self) -> None:
         """Sign this object, using another EFIAuth object as signer."""
         owner_cert = signer.owner_cert()
-        assert owner_cert is not None
```
ydirson, line 443: Is it supposed to be part of this commit?
Aside from this, I assume (in the absence of description, and because of the PR topic) that this removal was done because the `owner_cert()` sig does not allow `None`, are we sure we don't need it to catch unexpected situations or to make the code readable?

tests/install/test.py
```diff
@@ -191,7 +191,6 @@ def _test_firstboot(self, create_vms: list[VM], mode: str, *,
             assert len(ips) == 1
             host_vm.ip = ips[0]
             ip = host_vm.ip
-            assert ip is not None
```
ydirson, line 197: `VM.ip` is still typed `str | None`, why removing this assertion?

tests/storage/storage.py
```diff
@@ -198,12 +198,10 @@ def coalesce_integrity(vm: VM, vdi: VDI, vdi_op: CoalesceOperation, defer: Defer
     spans = partially_populate_device(vm, dev, vdi_size, 4, skip_spans=[1])
     # make sure we can read that exact data before the snapshot/clone
     validate_partially_populated_device(vm, dev, spans)
-    new_vdi: VDI | None = None
     match vdi_op:
         case 'clone': new_vdi = vdi.clone()
         case 'snapshot': new_vdi = vdi.snapshot()
     defer(lambda: new_vdi.destroy() if new_vdi is not None else None)
```
ydirson, line 208: IIUC `new_vdi` may be dereferenced while unbound here, don't we need to set it to `None` on the default match branch?

tests/storage/storage.py
```diff
@@ -198,12 +198,10 @@ def coalesce_integrity(vm: VM, vdi: VDI, vdi_op: CoalesceOperation, defer: Defer
...
-    assert vdi is not None
```
ydirson, line 210: Is this really meant to be part of this commit?
From its original commit I'm not sure what the role of this assertion is.

tests/storage/storage.py
```diff
@@ -224,9 +222,8 @@ def coalesce_integrity(vm: VM, vdi: VDI, vdi_op: CoalesceOperation, defer: Defer
 
 def xva_export_import(source_vm: VM, compression: XVACompression, temp_large_dir: str, defer: Defer) -> None:
     # clone the vm, so we can resize the disk without affecting the vm from the fixture
-    vm: VM | None = source_vm.clone()
+    vm = source_vm.clone()
     defer(lambda: vm.destroy() if vm is not None else None)
-    assert vm is not None
```
ydirson, line 238: if `vm` cannot be `None` (which seems to be the case) why do we need to test that condition in deferred call?
Note this really looks like cleanup of unnecessary logic than being on topic for this commit

tests/storage/storage.py
```diff
@@ -285,9 +282,8 @@ def xva_export_import(source_vm: VM, compression: XVACompression, temp_large_dir
     randstream(imported_vm, f'validate --expected-checksum {checksum} /root/data')
 
 def vdi_export_import(vm: VM, sr: SR, image_format: ImageFormat, temp_large_dir: str, defer: Defer) -> None:
-    vdi_src: VDI | None = sr.create_vdi(image_format=image_format, virtual_size=config.volume_size)
+    vdi_src = sr.create_vdi(image_format=image_format, virtual_size=config.volume_size)
     defer(lambda: vdi_src.destroy() if vdi_src is not None else None)
-    assert vdi_src is not None
```
ydirson, line 306: same as above, looks like incomplete cleanup rather than on-topic change

ydirson (changes requested):

Makefile
glehmann: Why? The code is loading the non-`-dist` ones, and the type checkers need them.

jobs.py
glehmann: It's just that the type checker don't understand that `category not in ("single", "multi")` means that it can later consider `category` as `Literal["single", "multi"]`.
Maybe we can rewrite the condition in a way that better match the way the static checkers work.

conftest.py
glehmann: Maybe. Shouldn't it require to be of the same type as declared in `Action`?

tests/install/conftest.py
glehmann: We don't run the unused import check in conftests.py files, because pytest actually needs these imports for some fixture, even though they don't seem used for the linter.

lib/efi.py
glehmann: For some reason, mypy is not happy without it:

```
  lib/efi.py:38: error: Incompatible types in assignment (expression has type
  "TemporaryDirectory[str]", variable has type "None")  [assignment]
                  self._instance = TemporaryDirectory()
                                   ^~~~~~~~~~~~~~~~~~~~
  lib/efi.py:40: error: Incompatible return value type (got "None", expected
  "TemporaryDirectory[str]")  [return-value]
              return self._instance
                     ^~~~~~~~~~~~~~
  Found 2 errors in 1 file (checked 214 source files)

File: lib/efi.py (hunk 5 above)
Comment (glehmann) [reply lib/efi.py:443]: The assert is already done in `owner_cert()`.

File: tests/install/test.py (hunk 6 above)
Comment (glehmann) [reply tests/install/test.py:197]: Because `host_vm.ip = ips[0]` is set just above, and `ips` is of type `list[str]` so `ip` is guaranteed to be of type `str`.
Also this assert was added in ec14611993b59de61ca5f0dcd14f67ca0f918923 to please the type checkers, so it's seems a good idea to remove it now that the type checker better analyzes the types.

File: tests/storage/storage.py (hunk 7 above)
Comment (glehmann) [reply tests/storage/storage.py:208]: We don't because `vdi_op` is of type `Literal['clone', 'snapshot']`, so there is no default branch.

File: tests/storage/storage.py (hunk 8 above)
Comment (glehmann) [reply tests/storage/storage.py:210]: Yes! The assert was there because `vdi` was of type `VDI | None`, so the methods using `vdi` later would know it's actually of type `VDI`.
With type narrowing, the type checker already knows that, at that point, the type is only `VDI`.
And BTW, the type checkers are clever enough to know that the type of `vdi` may actually be `VDI | None` when the lambda the line just above will be used.
image

File: tests/storage/storage.py (hunk 9 above)
Comment (glehmann) [reply tests/storage/storage.py:238]: Sorry, I partially replied to this in the previous comment.

`vdi` might be `None` at the time this lambda is run, and the type checkers properly infer that.
Line 216, we do:
`    new_vdi = None`

File: tests/storage/storage.py (hunk 10 above)
Comment (glehmann) [reply tests/storage/storage.py:306]: Same as above, we do
```py
    vdi_src = None
```
on line 274, so `vdi_src` might be `None` when the lambda is actually executed, and the type checkers properly infer the type as `VDI | None`.

image

jobs.py
glehmann: Actually pyright understands that, but not mypy.

Pyright:
image
mypy:
```
  jobs.py:625: error: TypedDict key must be a string literal; expected one of
  ("single", "multi")  [literal-required]
      ...   if category not in ("single", "multi") or key not in VMS[category]:
                                                                     ^~~~~~~~
  jobs.py:632: error: TypedDict key must be a string literal; expected one of
  ("single", "multi")  [literal-required]
          vms_unfiltered = VMS[category][key]
                               ^~~~~~~~
  Found 2 errors in 1 file (checked 214 source files)
  ```

Comment (glehmann) [reply jobs.py:654]: We could use an assertion, but if `category` is not `"single"` or `"multi"`, it would fail before we get a chance to print the message to the user.
I'm not sure how to guide mypy in the right direction in a less intrusive way.

File: Makefile (hunk 1 above)
Comment (ydirson) [reply Makefile:20]: Because the files that will be merged are the `-dist` ones.

File: jobs.py
Comment (ydirson) [reply jobs.py:654]: > Actually pyright understands that, but not mypy.

Are there actual situations where `mypy` catches defects that `pyright` does not? `mypy` has known deficiencies, and when we start to do changes in the code just so `mypy` can be happy, while `pyright` has no issue with it, I find it counter-productive. It might be better overall to just drop `mypy`...

Comment (ydirson) [reply jobs.py:654]: > We could use an assertion, but if `category` is not `"single"` or `"multi"`, it would fail before we get a chance to print the message to the user.

`assert XXX, "detailed message"` is the usual approach I take

File: conftest.py (hunk 2 above)
Comment (ydirson) [reply conftest.py:58]: That would be just type narrowing (not sure my arpgparse uses `Any` here, likely for some more advanced usage?)

File: lib/efi.py (hunk 4 above)
Comment (ydirson) [reply lib/efi.py:27]: That error is quite understandable, but it is strange it would start to appear just now, with only very little code changing around this?

File: tests/install/conftest.py (hunk 3 above)
Comment (ydirson) [reply tests/install/conftest.py:9]: Ah, yes, that one. It would be great when we'll have `pytest` plugins for type checkers...

File: jobs.py
Comment (glehmann) [reply jobs.py:654]: We would have to put the assert before the `if`, because `category` is used in the condition.
But that would defeat the intent of that `if` block, which is to provide a nice message to the user about a misconfiguration in his `vm_data.py`

> It might be better overall to just drop mypy...

Indeed, since the project is fully typehinted, I haven't seen mypy report useful problems that weren't already reported by pyright. If there is no objection, I'd be happy to keep a single type checker!

File: conftest.py (hunk 2 above)
Comment (glehmann) [reply conftest.py:58]: Indeed, it works.

File: jobs.py
Comment (vxgmichel) [reply jobs.py:654]: > I'd be happy to keep a single type checker!

I'm +1 on selecting a single type checker, and then focus on tuning it if we have to.

**Review by dinhngtu — APPROVED** [review]

File: tests/storage/lvmohba/test_lvmohba_sr.py [hunk 11]
```diff
@@ -53,36 +54,38 @@ def test_snapshot(self, vm_on_lvmohba_sr):
 
     @pytest.mark.small_vm
     @pytest.mark.parametrize("vdi_op", ["snapshot", "clone"])
-    def test_coalesce(self, storage_test_vm: 'VM', vdi_on_lvmohba_sr: 'VDI', vdi_op: CoalesceOperation, defer: Defer):
+    def test_coalesce(
+        self, storage_test_vm: 'VM', vdi_on_lvmohba_sr: 'VDI', vdi_op: CoalesceOperation, defer: Defer
```
semarie: should these (string) types replaced by real types ?

Makefile
semarie: for me, it makes sense to check for `data.py` and `vm_data.py`. the check are orthogonal about checking the typage inside `-dist` variants.

by having them as dependencies here, we are emitting a readable error with documentation url about configuring xcp-ng-tests. without it, users will only have a random python error about some missing files while importing.

semarie (approved):
LGTM.
few nits, but I think the PR should be committed (after dealing with the current conflicts) and the polishing finished in tree.

tests/storage/lvmohba/test_lvmohba_sr.py
glehmann: Good catch! I've also updated the ruff config to catch the quoted types even when the module doesn't import future annotations

Makefile
ydirson: The static checks are not supposed to be influenced by local configuration.  This encourages to modify the local files to get the checks to pass, without providing a way to check that the files that will actually be committed are correct.

jobs.py
ydirson: Actually, from recent experience on other projects, while `pyright` in strict mode does also better than `mypy` in strict mode, in standard mode it reports less.  And we cannot switch on strict mode (where `pyright` is definitely better). But we can likely enable pyright strict mode while a selection of checks (possibly different for `lib` and `tests`) and actually drop `mypy`.  WDYT @glehmann ?

lib/efi.py
ydirson: that sounds like one more vote to drop `mypy`?

jobs.py
ydirson: Rationale for holding this PR on this point is, if we start adding such cruft because of mypy, and we actually end up dropping it, we'll likely be stuck with it (unless we actually add comments that will allow us to notice we can remove it)

glehmann:
I updated this PR to:
- remove mypy entirely
- not introduce mypy specific changes

glehmann:
rebased on master to fix a conflict

Makefile
glehmann: I agree that the check shouldn't be influenced by an ignored file. In fact, I think it's one more reason not to use a Python file for the configuration.

We are already doing that for data.py, though, so this is just applying the same behavior to vm_data.py. Could we keep that discussion for another PR?

glehmann:
Updated to fix missing typehints introduced by recent PRs and fix conflicts

lib/common.py
```diff
@@ -133,15 +133,15 @@ def expand_scope_relative_nodeid(scoped_nodeid: str, scope: str, ref_nodeid: str
     return "::".join(itertools.chain(base, (scoped_nodeid,)))
 
 @lru_cache(maxsize=None)
-def _get_type_adapter(tp: Any) -> TypeAdapter[Any]:
+def _get_type_adapter(tp: Any) -> TypeAdapter[Any]:  # noqa: ANN401
```
ydirson, line 137: A way to quickly see what `ANN401` is about would be nice, maybe a note in the commit message?

tests/storage/linstor/test_linstor_sr.py
```diff
@@ -139,28 +135,26 @@ def host_and_vm_with_corrupted_vdi_on_linstor_sr(
...
+        vdi_host.ssh(
+            shlex.join([
+                "dd",
+                "if=/dev/urandom",
+                f"of={lv_path}",
+                "bs=4096",
+                # Lower values seem to go undetected sometimes
+                "count=10000",  # ~40MB
+            ])
+        )
```
ydirson: It is still quite annoying to see a commit with a given topic do tons of code reformatting, it forces the reader to filter out those changes.  This file literally has more formatting changes than type hints changes.

lib/common.py
ydirson, line 137: This commit message still mentions work done for `mypy`, for which we just dropped support

tests/storage/linstor/test_linstor_sr.py
```diff
@@ -1,10 +1,14 @@
...
+
 import pytest
 
 import json
 import logging
 import shlex
 import time
 
+from pydantic import BaseModel, Field, TypeAdapter
+
```
ydirson, line 11: That change is arguably more than just adding missing type hints, sounds like a sign it would need its separate commit?

tests/storage/linstor/test_linstor_sr.py
```diff
@@ -16,31 +20,50 @@
...
 
 # Requirements:
 # - two or more XCP-ng hosts >= 8.2 with additional unused disk(s) for the SR
 # - access to XCP-ng RPM repository from the host
 
 
-def get_drbd_status(host: Host, resource: str):
+class DrbdPeerDevice(BaseModel):
+    out_of_sync: int = Field(default=0, alias="out-of-sync")
+    peer_disk_state: str = Field(default="", alias="peer-disk-state")
```
ydirson, line 32: Do we really need aliases?  I can't see them used

tests/storage/linstor/test_linstor_sr.py
```diff
@@ -16,31 +20,50 @@
...
+
+
+class DrbdConnection(BaseModel):
+    name: str = ""
+    peer_devices: list[DrbdPeerDevice] = Field(default=[], alias="peer-devices")
+
+
+class DrbdResource(BaseModel):
+    name: str = ""
+    connections: list[DrbdConnection] = Field(default=[], alias="connections")
```
ydirson: alias with same name as field?

pyproject.toml
```diff
@@ -148,5 +143,5 @@ python-preference = "managed"
 
 [tool.pytest-language-server]
 # pytest-language-server shows some internal variables, like host, as undeclared fixture.
-# An undeclared fixture would be cought by pyright/mypy. Just disable that diagnostic.
+# An undeclared fixture would be cought by pyright. Just disable that diagnostic.
```
ydirson: While at it, `s/cought/caught/`

conftest.py
```diff
@@ -54,7 +54,7 @@
 assert CACHE_IMPORTED_VM in [True, False]
 
 class SplitCommaAction(Action):
-    def __call__(self, parser: ArgumentParser, namespace: Namespace, values: str | Sequence[Any] | None,
+    def __call__(self, parser: ArgumentParser, namespace: Namespace, values: str | Sequence[str] | None,
```
ydirson, line 57: This change is not related to the rest of this commit

lib/commands.py
```diff
@@ -188,40 +188,40 @@ def _ssh(
...
 # as pytest prints the whole function definition that raised the SSHCommandFailed exception
 @overload
-def ssh(hostname_or_ip: HostAddress, cmd: str, *, check: bool = True,
-        simple_output: Literal[True] = True,
-        suppress_fingerprint_warnings: bool = True, background: Literal[False] = False,
-        decode: Literal[True] = True, options: List[str] = [], multiplexing: bool = True) -> str:
+def ssh(hostname_or_ip: HostAddress, cmd: str, *, check: bool = ...,
+        simple_output: Literal[True] = ...,
+        suppress_fingerprint_warnings: bool = ..., background: Literal[False] = ...,
+        decode: Literal[True] = ..., options: List[str] = ..., multiplexing: bool = ...) -> str:
```
ydirson: Is there a check that would report deviations from this convention?

conftest.py
```diff
@@ -44,7 +44,7 @@
 # need to import them in the global conftest.py so that they are recognized as fixtures.
 from pkgfixtures import formatted_and_mounted_ext4_disk, sr_disk_wiped
 
-from typing import Any, Dict, Generator, Iterable, List, Optional, Sequence
+from typing import Any, Dict, Generator, Iterable, Sequence
```
ydirson: Remaining `Dict` here, you'll find a similar `Tuple` occurrence on a line untouched by this commit

lib/basevm.py
```diff
@@ -2,13 +2,14 @@
...
+from lib.common import _param_add, _param_clear, _param_get, _param_remove, _param_set
+from lib.sr import SR
+
+from typing import TYPE_CHECKING, Literal, overload
 
 if TYPE_CHECKING:
     from lib.host import Host
 
-from lib.common import _param_add, _param_clear, _param_get, _param_remove, _param_set
-from lib.sr import SR
```
ydirson, line 11: Didn't we have a check on import ordering already?  That looks like something that could/should be batch-fixed separately?

lib/vdi.py
```diff
@@ -13,15 +13,15 @@
...
+from typing import TYPE_CHECKING, Callable, Literal, TypeAlias, TypeVar, overload
 
 if TYPE_CHECKING:
     from lib.host import Host
     from lib.sr import SR
 
 R = TypeVar("R")
 
-ImageFormat = Literal['qcow2', 'raw', 'vhd']
+ImageFormat: TypeAlias = Literal['qcow2', 'raw', 'vhd']
```
ydirson, line 24: That one seems to rather belong to "Fix missing type annotations"?

ydirson (changes requested):

jobs.py
ydirson: With `mypy` gone we could do better here?

dinhngtu (approved):
LGTM for the Windows part.

vxgmichel (approved):

jobs.py
glehmann: It's already a better version, where we don't have to cast first. What would you do better?

lib/common.py
glehmann: ruff lsp provides that via a very convenient hover tooltip, but I can add it in the commit message

lib/vdi.py
glehmann: fixed

lib/basevm.py
glehmann: The check doesn't cover cases where imports appear after the `if TYPE_CHECKING:` block.
I've added a flake8 plugin to test that in https://github.com/xcp-ng/xcp-ng-tests/pull/696. This is the only place where the order is not right.

conftest.py
glehmann: fixed

lib/commands.py
glehmann: Not out of the box. I've added a flake8 plugin for that in https://github.com/xcp-ng/xcp-ng-tests/pull/696

conftest.py
glehmann: fixed

pyproject.toml
glehmann: fixed

tests/storage/linstor/test_linstor_sr.py
glehmann: fixed

tests/storage/linstor/test_linstor_sr.py
glehmann: done

lib/common.py
glehmann: fixed

tests/storage/linstor/test_linstor_sr.py
glehmann: There were a few unwanted reformatted line, on lines that were not edited otherwise. I removed those changes.

tests/storage/linstor/test_linstor_sr.py
glehmann: they map the hyphenated json keys from `drbdsetup status --json` (`out-of-sync`, `peer-disk-state`, `peer-devices`) to the Python fields. They are used implicitly by pydantic's `validate_json` in `get_drbd_status()`, so they don't show up anywhere in the code.

ydirson: That sounds useful to explain somewhere, maybe in a introductory comment before the type definitions?

glehmann: Ok, but that's just basic pydantic usage

ydirson: The fact they come from `drbdsetup`, I mean

PR #454, opened 2026-03-31
storage: test that we can't create a vdi over its max allowed size

glehmann:
Signed-off-by: Gaëtan Lehmann <gaetan.lehmann@vates.tech>

This PR is part of a tree containing 19 PRs:

1. `master`
2. #436 → `master`
3. #437 → #436
4. #447 → #437
5. #446 → #447
6. #449 → #446
7. #450 → #449
8. #452 → #450
9. #453 → #452
10. **"storage: test that we can't create a vdi over its max allowed size" (this PR) → #453**
11. #461 → #454
12. #464 → #461
13. #470 → #464
14. #471 → #470
15. #481 → #471
16. #523 → #481
17. #497 → #481
18. #498 → #497
19. #500 → #498
20. #509 → #500

PR #461, opened 2026-04-07
storage: add jobs for large volume tests

glehmann:
Mark LVM and lvmoiscsi tests that allocate large VDIs with disk_space_allocation,
and all test_full_vdi_write tests with disk_throughput_intensive, to allow running
large-VDI jobs only on hardware that can handle the respective disk requirements.

Add storage-main-16tb-{thin,thick,full-write} and storage-migrations-16tb-{thin,thick}
jobs that use these markers to split 16TiB volume testing by disk speed requirement.

1. `master`
2. #436 → `master`
3. #437 → #436
4. #447 → #437
5. #446 → #447
6. #449 → #446
7. #450 → #449
8. #452 → #450
9. #453 → #452
10. #454 → #453
11. **"storage: add jobs for large volume tests" (this PR) → #454**
12. #464 → #461
13. #470 → #464
14. #471 → #470
15. #481 → #471
16. #523 → #481
17. #497 → #481
18. #498 → #497
19. #500 → #498
20. #509 → #500

### DIFF
--- a/jobs.py
+++ b/jobs.py
@@ -101,6 +101,62 @@ JOBS: dict[str, JobData] = {
         "markers": "(small_vm or no_vm) and not reboot and not quicktest and not unused_4k_disks",
         "name_filter": "not migration and not linstor",
     },
+    "storage-main-large-thin": {
+        "description":
+            "same as storage-main with 3TiB VDIs, no large gzip XVAs creation and no test requiring 3TiB allocation",
+        "requirements": [
+            "A pool with at least 3 hosts.",
+            "An additional free disk on every host.",
+            "Configuration in data.py for each remote SR that will be tested.",
+            "A small VM that can be imported on the SRs.",
+        ],
+        "nb_pools": 1,
+        "params": {
+            "--vm": "single/small_vm",
+            "--volume-size": "3TiB",
+        },
+        "paths": ["tests/storage"],
+        "markers": "(small_vm or no_vm) and not reboot and not quicktest and not unused_4k_disks"
+                   " and not thick_provisioned and not disk_throughput_intensive",
+        "name_filter": "not migration and not linstor and not gzip",
+    },
+    "storage-main-large-thick": {
+        "description":
+            "same as storage-main with 3TiB VDIs, no gzip XVAs and tests requiring 3TiB allocation",
+        "requirements": [
+            "A pool with at least 3 hosts.",
+            "An additional free disk on every host.",
+            "Configuration in data.py for each remote SR that will be tested.",
+            "A small VM that can be imported on the SRs.",
+        ],
+        "nb_pools": 1,
+        "params": {
+            "--vm": "single/small_vm",
+            "--volume-size": "3TiB",
+        },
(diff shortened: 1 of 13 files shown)


jobs.py
```diff
@@ -90,6 +90,62 @@
         "markers": "(small_vm or no_vm) and not reboot and not quicktest and not unused_4k_disks",
         "name_filter": "not migration and not linstor",
     },
+    "storage-main-16tb-thin": {
+        "description":
+            "same as storage-main with 16TiB VDIs, no gzip XVAs and no test requiring 16TiB allocation",
```
stormi: No gzip XVAs means we can't work with an initial VM which is a gzipped XVA, or that you exclude tests that would create such XVAs?

jobs.py
```diff
@@ -90,6 +90,62 @@
         "markers": "(small_vm or no_vm) and not reboot and not quicktest and not unused_4k_disks",
         "name_filter": "not migration and not linstor",
     },
+    "storage-main-16tb-thin": {
```
stormi: We'll likely want similar jobs with different sizes. How about naming it `storage-large-thin` with a default value for `--volume-size` (I'd go with something that we can easily run in our lab, maybe 3 TB?) and then we'd override the volume size for larger disks?

pytest.ini
```diff
@@ -32,6 +32,10 @@ markers =
     multi_vms: tests that it would be good to run on a variety of VMs (includes `small_vm` but excludes `big_vm`).
     debian_uefi_vm: tests that require a Debian UEFI VM
 
+    # * Disk-related markers
+    disk_space_allocation: tests that allocate a large VDI (disk won't be fully written, slow disk is OK).
```
klmp200: I think it would be better to have a `thick_provisionned` marker or something like this, it would be less ambiguous and more flexible.

klmp200 (changes requested):

pytest.ini
stormi: You mean, instead of `disk_space_allocation` (which doesnt mean the same thing), or in addition to it?

jobs.py
glehmann: It means we exclude tests that create gzip xvas

stormi: Thanks. I suggest rewording it slightly to avoid the ambiguity.

pytest.ini
glehmann: We are using `disk_space_allocation` in the `storage-main-16tb-thick` job `jobs.py`, to select those tests and run them on a host with enough actual space.
Using a `thick_provisioned` marker on the whole class would be easier to understand, would better match the job name, and would reduce the chance of forgetting to tag a new test.

glehmann: I've updated the PR with that marker

stormi (changes requested):
The commit message doesn't match the change anymore. I stopped the review at this stage, will re-review when asked.

Lankou66 (approved):

Millefeuille42 (approved):

PR #468, opened 2026-04-10
tools/update - create snapshot for nested hosts

olivierh-pro:
With this PR, `tools/update` can now create snapshots for nested hosts we are working on.

### DIFF
--- a/lib/tools/tasks/snapshot.py
+++ b/lib/tools/tasks/snapshot.py
@@ -1,0 +1,28 @@
+"""Snapshot tasks.
+"""
+from concurrent.futures import ThreadPoolExecutor
+from datetime import date
+
+from lib.host import Host
+from lib.vm import VM
+
+from .. import logger
+
+def create_snapshot(host: Host, vm_uuids: list[str]):
+    """Create a snapshot for a list of VMs in host.
+
+    :param `lib.Host` host:
+        The target parent which hosts the VMs.
+    :param list[str] vm_uuids:
+        uuids of target VMs.
+    """
+    # init VMs list
+    vms = [VM(uuid, host) for uuid in vm_uuids]
+
+    snapshot_name = f"utd-{date.today().strftime('%Y%m%d')}"
+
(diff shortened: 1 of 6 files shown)


lib/tools/tasks/snapshot.py
```diff
@@ -0,0 +1,28 @@
...
+
+from lib.host import Host
+from lib.vm import VM
+
+from .. import logger
+
+def create_snapshot(host: Host, vm_uuids: list[str]):
+    """Create a snapshot for a list of VMs in host.
+
+    :param `lib.Host` host:
```
glehmann: Do we need to put the type there? It's already in the function signature.

glehmann (approved):

lib/tools/tasks/snapshot.py
olivierh-pro: You're right! It's not necessary!

lib/tools/tasks/snapshot.py
```diff
@@ -0,0 +1,28 @@
...
+"""
+from concurrent.futures import ThreadPoolExecutor
+from datetime import date
+
+from lib.host import Host
+from lib.vm import VM
+
+from .. import logger
+
+def create_snapshot(host: Host, vm_uuids: list[str]):
```
rzr:
```suggestion
def create_snapshot(host: Host, vm_uuids: list[str], name: None):
```
And generate a string if none, later

lib/tools/tasks/update.py
```diff
@@ -47,3 +57,9 @@ def update_pools(inventory: Inventory) -> None:
                 # repos are the same as for the master host
                 repos = inventory_hosts[p.master.hostname_or_ip]["repositories"]
                 executor.submit(other_host.update, repos)
+
+    # Snapshot creation
+    for hosting_pool, nested in nested_hosts.items():
+        pool = Pool(hosting_pool) # mandatory for getting an host instance
+        vm_uuids = [h.get_system_uuid() for h in nested]
+        create_snapshot(pool.master, vm_uuids)
```
rzr: If you want to track snapshots with a same name, it can be added here.

rzr (approved):

PR #469, opened 2026-04-10
Add Windows tools onboarding tests

dinhngtu:
XenClean implements an "onboarding" command that lets users of tools
from other vendors automatically transition to XCP-ng tools. The
XenClean process returns an exit code depending on whether any tools are
installed, whether the VM is ready for XCP-ng tools installation, etc.

Add a test suite for this functionality.

On top of that, disable secure boot in enable_testsign. (I've made this error often enough when creating images that it's easier to change it from within the tests)

### DIFF
--- a/lib/windows/__init__.py
+++ b/lib/windows/__init__.py
@@ -63,6 +63,16 @@ def wait_for_vm_running_and_ssh_up_without_tools(vm: VM):
 
 
 def enable_testsign(vm: VM, rootcert: Union[str, None]):
+    assert vm.is_running()
+
+    if strtobool(vm.param_get("platform", "secureboot")):
+        logging.info("Disable secure boot on test image")
+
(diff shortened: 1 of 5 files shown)


lib/windows/__init__.py
```diff
@@ -63,6 +63,14 @@ def wait_for_vm_running_and_ssh_up_without_tools(vm: VM):
 
 
 def enable_testsign(vm: VM, rootcert: Union[str, None]):
```
stormi, line 65: Not directly related to this PR, but it might be useful to document that the VM described by the `vm` parameter must be running.

stormi (approved):

lib/windows/__init__.py
dinhngtu: Thanks, added assert

PR #472, opened 2026-04-13
Implement non-default markers

dinhngtu:
Some markers (e.g. reboot/flaky) represent tests that shouldn't be run in the default suite. As a result, we have to negate them over and over in jobs.py. Add a mechanism to skip them by default. These tests can be manually enabled using a pytest argument.

Add a "slow" marker to represent slow tests for detecting rare issues.

Add a couple initial slow tests to catch previously-encountered issues.

Add some typing to conftest.py while we're at it.

### DIFF
--- a/jobs.py
+++ b/jobs.py
@@ -468,7 +468,22 @@ JOBS = {
             "--vm[]": "multi/limits",
         },
         "paths": ["tests/limits"],
-    }
+    },
+    "slow": {
+        "description": "tests that may take a long time",
+        "requirements": [
+            "Will vary depending on the tests included.",
+            "Use the collect command to get the list of tests "
+            + "and check the requirements written at the top of the test files.",
+        ],
+        "nb_pools": 1,
+        "params": {
+            "--enable-slow": True,
+            "--vm[]": "multi/slow",
+        },
+        "paths": ["tests"],
+        "markers": "slow",
+    },
 }
 
 # List used by the 'check' action: tests listed here will not raise a check error
--- a/jobs.py
+++ b/jobs.py
@@ -32,7 +32,7 @@ JOBS = {
             "tests/xapi_plugins",
             "tests/install/test_fixtures.py",
         ],
-        "markers": "(small_vm or no_vm) and not flaky and not reboot and not complex_prerequisites",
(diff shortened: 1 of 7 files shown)


dinhngtu:
`parametrize` made the same test show up 100 times in `collect` so I scrapped it and went with a normal for loop instead.

stormi:
> `parametrize` made the same test show up 100 times in `collect` so I scrapped it and went with a normal for loop instead.

parametrize often looks like a good idea until we actually try to use it :)

glehmann (commented):
Looks mostly good to me, but it is missing some documentation in the README.md.

If we add more excluded makers by default, we might want to add a `--enable-all`, but I don't think that's necessary for now.

dinhngtu:
Added documentation. Also rearranged pytest.ini to explicitly list the non-default markers.

jobs.py
```diff
@@ -468,7 +468,22 @@
             "--vm[]": "multi/limits",
         },
         "paths": ["tests/limits"],
-    }
+    },
+    "slow": {
```
stormi, line 472: We can start with that, but I think we'll quickly see the need for a way to be more specific about which slow tests to run, depending on their requirements.

dinhngtu: Indeed I ran into this when trying to add `tests/guest_tools/win` into the slow suite.

stormi (approved):
I have doubts about some of the choices, but nothing that we can't improve later.

glehmann (approved):
Please make sure to update `vm_data.py` in the `jenkins-ci/ci-configuration` repository before merging this PR

stormi (changes requested):
On second thought I don't like the implementation. It will cause annoyances when once wants to run a specific test from a failed job, or execute tests manually, only to discover that pytest skipped parts of the tests when it's too late already.

To me, the implementation should be limited to jobs.py: automatically adding "not XXX" to the `-m` switch, in order to exclude some markers by default, unless they're specified in the job definition or manually via `-m` on the command line.

dinhngtu:
Using a for loop made per-test setup/teardown more complicated compared to using a fixture, so I'm having second thoughts...

PR #475, opened 2026-04-15
Fix XAPI session leak by vdi_is_open

Nambrok:
In the case of `vdi_is_open` tests, we are using a script to obtain the sr_ref of XAPI. This script doesn't free the XAPI session it obtained.
```
Apr 15 09:46:55 r620-x1 xapi: [debug||115 db_gc|DB GC D:d2304dcf287c|db_gc_util] Session.destroy _ref=OpaqueRef:77699499-92e7-ad96-2703-fc53adaae642 uuid=e8a67d26-9aae-4451-ea87-66a3180141fb trackid=3c4e5b67d27d1b17ff45f3e41b478632 (last active 20260414T07:46:32Z): Timed out session in group 'originator:xcp-ng-tests session' because of its age
```
This add the logout to the script to avoid leaking the session.

### DIFF
--- a/tests/storage/storage.py
+++ b/tests/storage/storage.py
@@ -139,7 +139,10 @@ def get_xapi_session():
     return session
 
 session = get_xapi_session()
-sr_ref = session.xenapi.SR.get_by_uuid(\"{sr.uuid}\")
+try:
+    sr_ref = session.xenapi.SR.get_by_uuid(\"{sr.uuid}\")
+finally:
+    session.xenapi.logout()
 print(sr_ref)
 """
 


stormi (approved):

tests/storage/storage.py
```diff
@@ -141,6 +141,7 @@ def get_xapi_session():
 session = get_xapi_session()
 sr_ref = session.xenapi.SR.get_by_uuid(\"{sr.uuid}\")
 print(sr_ref)
+session.xenapi.logout()
```
Wescoeur:
```suggestion
try:
    sr_ref = session.xenapi.SR.get_by_uuid(\"{sr.uuid}\")
finally:
    session.xenapi.logout()
print(sr_ref)
```

Wescoeur (changes requested):

tests/storage/storage.py
Nambrok: Done

Wescoeur (approved):

tests/storage/storage.py
stormi: I guess it wouldn't work with a `with` ? (does session.xenapi logout when you destroy the object?)

glehmann:
nitpicking: "session leak" seems scarier than necessary. I'd rather call it "dangling session"

Nambrok:
> nitpicking: "session leak" seems scarier than necessary. I'd rather call it "dangling session"

Leak completely describe the situation though ^^

glehmann:
It's not leaked in the sense that someone can't get the token or whatever represents the session and do something not allowed. The session is still open for the server, but nobody can use it.
I guess you used it as in "memory leak", I read it as in "secret leak" :)

stormi:
Can we merge?

PR #476, opened 2026-04-15
Add ty type checker configuration

glehmann:
Add ty configuration. ty provides fast type checking that complements
pyright/mypy, while ruff handles missing type hints detection.

ty is not yet reporting as much problems as mypy or pyright, but it is
much, much faster, and is already interesting to use interactively.

This change only aims at making ty usable in this project, without
replacing mypy or pyright. As such it is not added in the dev dependencies
and must be installed manually, for example with

1. `master`
2. #460
3. **"chore: add ty type checker and reorder make commands for faster feedback" (this PR)**

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -33,6 +33,15 @@ repos:
         require_serial: true
         entry: "uv run flake8"
         types: ["python"]
+      - id: "ty"
+        name: "ty"
+        language: "system"
+        require_serial: true
+        entry: "uv run ty check ."
+        types: ["python"]
+        stages: ["manual"]
+        pass_filenames: false
+        always_run: true
       - id: "pyright"
         name: "pyright"
         language: "system"
--- a/lib/config.py
+++ b/lib/config.py
@@ -1,10 +1,10 @@
 from lib.common import GiB
 
-ignore_ssh_banner = False
-ssh_output_max_lines = 20
-volume_size = 1 * GiB
-write_volume_cap = 2 * GiB
-write_volume_align = 1
+ignore_ssh_banner: bool = False
+ssh_output_max_lines: int = 20
+volume_size: int = 1 * GiB
+write_volume_cap: int = 2 * GiB
+write_volume_align: int = 1
 
 def sr_device_config(datakey: str, *, required: list[str] = []) -> dict[str, str]:
     import data  # import here to avoid depending on this user file for collecting tests
--- a/scripts/install_xcpng.py
+++ b/scripts/install_xcpng.py
@@ -18,7 +18,7 @@ from packaging import version
 # flake8: noqa: E402
 sys.path.append(f"{os.path.abspath(os.path.dirname(__file__))}/..")
 from lib import pxe
-from lib.commands import SSHCommandFailed, ssh
+from lib.commands import SSHCommandFailed, local_cmd, ssh
 from lib.common import is_uuid, wait_for
 from lib.host import Host, host_data
 from lib.pool import Pool
@@ -72,7 +72,7 @@ def generate_answerfile(directory: str, installer: str, hostname_or_ip: str, tar
             raise Exception(f"Unknown action: `{action}`")
 
 def is_ip_active(ip: str) -> bool:
-    return not os.system(f"ping -c 3 -W 10 {ip} > /dev/null 2>&1")
+    return local_cmd(["ping", "-c", "3", "-W", "10", ip], check=False).returncode == 0
 
 def is_ssh_up(ip: str) -> bool:
     try:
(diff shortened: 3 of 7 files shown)


rzr (approved):
Do you want to commit lock file on each changes ? i would only do this on release (using calver.org ?)

glehmann:
> Do you want to commit lock file on each changes ? i would only do this on release (using calver.org ?)

Yes, it's meant to be used that way, to ensure that all developers are using the same dependency version.

olivierh-pro:
If we use `ty` as type checker, should it be useful to add the related config for LSP in README ?

glehmann:
We get errors with the newest versions. It's not ready yet for prime time. I'm moving this PR to draft

ydirson (commented):
The `chore:` prefix is not applicable to this repo, and I'm not sure it would be correct in that standard ;)
I suggest `checks: ` would be more informative.

.pre-commit-config.yaml
```diff
@@ -33,6 +33,13 @@ repos:
         require_serial: true
         entry: "uv run flake8"
         types: ["python"]
+      - id: "ty"
+        name: "ty"
+        language: "system"
+        require_serial: true
+        entry: "uv run ty check ."
+        types: ["python"]
+        stages: ["manual"]
```
olivierh-pro, line 42: The official "ty pre-commit" recommends using : 

```
pass_filenames: false
always_run: true
```

Réf: https://github.com/astral-sh/ty-pre-commit/blob/main/.pre-commit-hooks.yaml

Maybe those parameters are in conflict with "manual" behaviour?

pyproject.toml
```diff
@@ -18,22 +18,22 @@ dependencies = [
 
 [dependency-groups]
 dev = [
```
semarie, line 23: shouldn't `ty` added as dependency too ? else it will not be available via `uv` ?

semarie (changes requested):
globally LGTM, but some concern about `ty` dependency : does it should be added in `uv` dependency (or it is too soon for now) ?

pyproject.toml
stormi: This is explained in the PR description.

semarie (approved):

olivierh-pro (approved):

stormi (approved):

scripts/install_xcpng.py
```diff
@@ -18,7 +18,7 @@
 # flake8: noqa: E402
 sys.path.append(f"{os.path.abspath(os.path.dirname(__file__))}/..")
 from lib import pxe
-from lib.commands import SSHCommandFailed, ssh
+from lib.commands import SSHCommandFailed, local_cmd, ssh
```
ydirson, line 21: Why is this part of this PR? It should already be part of #359

lib/config.py
```diff
@@ -1,10 +1,10 @@
...
-ignore_ssh_banner = False
-ssh_output_max_lines = 20
-volume_size = 1 * GiB
-write_volume_cap = 2 * GiB
-write_volume_align = 1
+ignore_ssh_banner: bool = False
+ssh_output_max_lines: int = 20
+volume_size: int = 1 * GiB
+write_volume_cap: int = 2 * GiB
+write_volume_align: int = 1
```
ydirson, line 7: I'm a bit disappointed if we have to add superfluous annotations for `ty`, while we just decided to drop `mypy` to avoid adding superfluous annotations.  I'd argue it is possibly too early for a `ty` config if this needs to go in with it.
Also, those changes are not even mentioned in the commit message introducing them.

.pre-commit-config.yaml
```diff
@@ -33,6 +33,15 @@ repos:
         require_serial: true
         entry: "uv run flake8"
         types: ["python"]
+      - id: "ty"
```
ydirson, line 36: One possibility to install `ty` aside `uv tool install` is `uv pip install`, don't know if we want to mention it?
Also, such instructions could belong to the README?

ydirson (changes requested):

scripts/install_xcpng.py
glehmann: https://github.com/xcp-ng/xcp-ng-tests/pull/476#issuecomment-4981682311

.pre-commit-config.yaml
glehmann: I don't think it belongs in the README yet, and when it does, it will also be time to add it as a dependency, so we might not need it in the README at all :)

lib/config.py
glehmann: Yes that's not ideal, but that's a quite small amount of changes.
We could also decide to *not* add those annotations. It would still make ty useful as a LSP, but would show some false errors when run as a cli.

scripts/install_xcpng.py
ydirson: Since ty support is accessory, what about boosting #359 first?

Wescoeur (approved):

PR #471, opened 2026-04-13
storage: free space for XVA import by destroying source VM first

glehmann:
The test wasn't working properly with tick provisioning: in that case
the space from the source VM may be required to be able to import the
image.

Signed-off-by: Gaëtan Lehmann <gaetan.lehmann@vates.tech>

This PR is part of a tree containing 19 PRs:

1. `master`
2. #436 → `master`
3. #437 → #436
4. #447 → #437
5. #446 → #447
6. #449 → #446
7. #450 → #449
8. #452 → #450
9. #453 → #452
10. #454 → #453
11. #461 → #454
12. #464 → #461
13. #470 → #464
14. **"storage: free space for XVA import by destroying source VM first" (this PR) → #470**
15. #481 → #471
16. #523 → #481
17. #497 → #481
18. #498 → #497
19. #500 → #498
20. #509 → #500

PR #470, opened 2026-04-13
storage: add comprehensive tests for lvmohba storage repositories

glehmann:
Signed-off-by: Gaëtan Lehmann <gaetan.lehmann@vates.tech>

This PR is part of a tree containing 19 PRs:

1. `master`
2. #436 → `master`
3. #437 → #436
4. #447 → #437
5. #446 → #447
6. #449 → #446
7. #450 → #449
8. #452 → #450
9. #453 → #452
10. #454 → #453
11. #461 → #454
12. #464 → #461
13. **"storage: add comprehensive tests for lvmohba storage repositories" (this PR) → #464**
14. #471 → #470
15. #481 → #471
16. #523 → #481
17. #497 → #481
18. #498 → #497
19. #500 → #498
20. #509 → #500

PR #464, opened 2026-04-08
skip large volume tests for zvol and nfsv4

glehmann:
They are slow and currently out of scope.

This PR is part of a tree containing 19 PRs:

1. `master`
2. #436 → `master`
3. #437 → #436
4. #447 → #437
5. #446 → #447
6. #449 → #446
7. #450 → #449
8. #452 → #450
9. #453 → #452
10. #454 → #453
11. #461 → #454
12. **"skip large volume tests for zvol and nfsv4" (this PR) → #461**
13. #470 → #464
14. #471 → #470
15. #481 → #471
16. #523 → #481
17. #497 → #481
18. #498 → #497
19. #500 → #498
20. #509 → #500

PR #481, opened 2026-04-16
storage: limit data written per VDI with --write-volume-cap

glehmann:
Refactor stream management to support arbitrary span distributions.
The new StreamSpan dataclass encapsulates stream metadata and operations,
while partially_populate_device() now supports configurable span counts
and skip patterns.

This required to upgrade randstream to the version 0.6.0 which supports
arbitrary position.
 
The default cap is 2GiB.

1. `master`
2. #436 → `master`
3. #437 → #436
4. #447 → #437
5. #446 → #447
6. #449 → #446
7. #450 → #449
8. #452 → #450
9. #453 → #452
10. #454 → #453
11. #461 → #454
12. #464 → #461
13. #470 → #464
14. #471 → #470
15. **"storage: limit data written per VDI with --write-volume-cap" (this PR) → #471**
16. #523 → #481
17. #497 → #481
18. #498 → #497
19. #500 → #498
20. #509 → #500

### DIFF
--- a/lib/config.py
+++ b/lib/config.py
@@ -3,6 +3,7 @@ from lib.common import GiB
 ignore_ssh_banner = False
 ssh_output_max_lines = 20
 volume_size = 1 * GiB
+write_volume_cap = 2 * GiB
 
 def sr_device_config(datakey: str, *, required: list[str] = []) -> dict[str, str]:
     import data  # import here to avoid depending on this user file for collecting tests
--- a/tests/storage/nfs/test_nfs_sr.py
+++ b/tests/storage/nfs/test_nfs_sr.py
@@ -140,9 +140,9 @@ class TestNFSSR:
     @pytest.mark.usefixtures('vm_ref')
     @pytest.mark.parametrize('dispatch_nfs', ['vm_on_nfs_sr', 'vm_on_nfs4_sr'], indirect=True)
     @pytest.mark.parametrize("compression", ["none", "gzip", "zstd"])
-    def test_xva_export_import(self, dispatch_nfs: VM, compression: XVACompression, temp_large_dir: str, defer: Defer) \
-            -> None:
-        if "NFS4" in dispatch_nfs.vdis[0].sr.get_name_label() and config.volume_size > 20 * GiB:
+    def test_xva_export_import(self, dispatch_nfs: VM, compression: XVACompression, temp_large_dir: str,
+                               defer: Defer) -> None:
+        if "NFS4" in dispatch_nfs.vdis[0].sr.get_name_label() and config.write_volume_cap > 20 * GiB:
             pytest.skip("Skipping NFSv4 large VDI test (known performance issue)")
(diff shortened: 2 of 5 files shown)


lib/config.py
```diff
@@ -3,6 +3,7 @@
 ignore_ssh_banner = False
 ssh_output_max_lines = 20
 volume_size = 1 * GiB
+write_volume_cap = 2 * GiB
```
rzr, line 6: is it twice volume size ?

tests/storage/nfs/test_nfs_sr.py
```diff
@@ -140,17 +140,17 @@ def test_invalid_vdi_size(self, nfs_sr: SR, image_format: ImageFormat):
     @pytest.mark.usefixtures('vm_ref')
     @pytest.mark.parametrize('dispatch_nfs', ['vm_on_nfs_sr', 'vm_on_nfs4_sr'], indirect=True)
     @pytest.mark.parametrize("compression", ["none", "gzip", "zstd"])
-    def test_xva_export_import(self, dispatch_nfs: VM, compression: XVACompression, temp_large_dir: str, defer: Defer) \
-            -> None:
-        if "NFS4" in dispatch_nfs.vdis[0].sr.get_name_label() and config.volume_size > 20 * GiB:
+    def test_xva_export_import(self, dispatch_nfs: VM, compression: XVACompression, temp_large_dir: str,
+                               defer: Defer) -> None:
+        if "NFS4" in dispatch_nfs.vdis[0].sr.get_name_label() and config.write_volume_cap > 20 * GiB:
```
rzr, line 145: Sorry if asked previously but why magic numbers are duplicated, and not a constant introduced?

tests/storage/storage.py
```diff
@@ -27,35 +26,13 @@ def try_to_create_sr_with_missing_device(sr_type, label, host) -> None:
...
-
-
 def cold_migration_then_come_back(vm: VM, prov_host: Host, dest_host: Host, dest_sr: SR) -> None:
     """ Storage migration of a shutdown VM, then migrate it back. """
     prov_sr = vm.get_sr()
     vdi_name: str | None = None
     integrity_check = not vm.is_windows
     dev = ""
-    checksums = ('', '', '')
+    spans: list[StreamSpan] = []
```
rzr, line 35: What about using sets in place of list ?

tests/storage/nfs/test_nfs_sr.py
glehmann: Because I didn't feel a need for it, I suppose.

Indeed, I think we've already had this discussion :)

tests/storage/storage.py
glehmann: That's the type returned by the function that creates them.
And it's convenient to use a list as return type because they are not all equivalent when using `skip_span`.

Lankou66 (approved):

Millefeuille42 (approved):

PR #491, opened 2026-04-21
Introduce TOML-based configuration to replace data.py

glehmann:
The previous data.py approach had several shortcomings:
- No base configuration: users had to start from scratch or copy data.py-dist
- No environment management: working with multiple infra required to manually
  copy the data.py files
- No clear default and customization separation: the data.py, once copied
  from data.py-dist and customized, was mixing the defaults and the
  customizations

The new TOML config system addresses this by:
- Providing a base config.toml maintained by the project with sensible defaults,
  which can be updated without user action
- Supporting per-environment override files (config.<name>.toml) so users only
  specify what differs from the defaults, making their customizations explicit

Options have been grouped by sections for clarity.

Configuration sections and their contents:
- [host]: default XAPI credentials (user, password)
- [hosts]: per-host credential overrides
- [network]: management network name
- [pxe]: PXE server settings (arp_server)
- [vm]: VM settings (default_sr, images, equivalents, def_url)
- [install]: installer settings (answerfiles, iso_remaster, isos)
- [guest_tools]: guest tools packages and installer ISOs (win, other, installed, download_url)
- [ssh]: SSH client settings (pubkey, output_max_lines, ignore_banner)
- [storage]: storage backend configs (nfs, nfs4, nfs_iso, cifs_iso, cephfs, moosefs, lvmoiscsi)
- Root keys: objects_name_prefix, dns_server

Also includes config-schema.json, lib/config_loader.py (Pydantic models),
and scripts/migrate_data_py.py to help users migrate from data.py.

1. `master`
2. #396 → `master`
3. #431 → #396
4. **"Introduce TOML-based configuration to replace data.py" (this PR) → #396**


config-schema.json
```diff
@@ -0,0 +1,312 @@
...
+              "description": "Local directory to cache downloaded ISOs"
+            },
+            "definitions": {
+              "type": "object",
+              "description": "ISO version definitions keyed by version name",
+              "additionalProperties": {
+                "type": "object",
+                "properties": {
+                  "path": { "type": "string" },
+                  "net-url": { "type": "string" },
```
olivierh-pro: Maybe you should choose a convention "snake_case" or "kebab-case", what do you think?

stormi:
@glehmann Moving away from data.py (and probably vm_data.py) to something better is clearly needed. Do you have a documented list of requirements (which maybe some team members would want to add to) and summary of the proposed alternative to address them?

glehmann:
> @glehmann Moving away from data.py (and probably vm_data.py) to something better is clearly needed. Do you have a documented list of requirements (which maybe some team members would want to add to) and summary of the proposed alternative to address them? 

I tried to document that in the file CONFIG.md in this PR: https://github.com/xcp-ng/xcp-ng-tests/pull/491/changes#diff-9a8f890c7271a9e9bdabd6026ecd9f8aa10fe23a6f9edd7cc09a24bf73eb0725

CONFIG.md
```diff
@@ -0,0 +1,549 @@
...
+model and its main consumers.
+
+### Root values
+
+- `objects_name_prefix` prefixes the labels of XAPI objects created by tests.
+- `dns_server` identifies the DNS server used by relevant VM tests.
+- `volume_size` and `write_volume_cap` control storage-test volume sizes and
+  write limits.
+- `write_volume_align` controls the block alignment used by volume-write
+  tests.
```
stormi: Do they need to be root values rather than under [storage]?

stormi:
> > @glehmann Moving away from data.py (and probably vm_data.py) to something better is clearly needed. Do you have a documented list of requirements (which maybe some team members would want to add to) and summary of the proposed alternative to address them?
> 
> I tried to document that in the file CONFIG.md in this PR: https://github.com/xcp-ng/xcp-ng-tests/pull/491/changes#diff-9a8f890c7271a9e9bdabd6026ecd9f8aa10fe23a6f9edd7cc09a24bf73eb0725

Thanks! I've read it, and I think the overlay approach works for the requirements I can think of at the moment. I'll keep it in mind and tell you if I realize that there's something that wouldn't fit cleanly in the picture.

Have other team members read it yet?

glehmann:
> Have other team members read it yet?

@vxgmichel even wrote part of it (sorry Vincent, for not mentioning you earlier).
I haven't shown it to the other team members yet

CONFIG.md
glehmann: Indeed that would be better in the `[storage]` section :+1:

lib/config.toml
```diff
@@ -0,0 +1,278 @@
+"$schema" = "../config-schema.json"
+# Configuration file for XCP-ng tests
+# To be customized for your environment
+
+# The following prefix will be added to the `name-label` parameter of XAPI objects
+# that the tests will create or import, such as VMs and SRs.
+# Default value: [your login/user]
+# Prefix for test-created objects in XAPI (empty = use username)
+objects_name_prefix = ""
```
gduperrey: I would tend to suggest a three-letter code or something standard (like the first letter of the first name plus the last name, for example). This is mainly to avoid ending up with a lot of pseudonyms that are hard to link to a specific person, as is already the case with GitHub pseudonyms and PRs.

lib/config.toml
```diff
@@ -0,0 +1,278 @@
...
+
+# Block size to align span positions to when writing in volumes.
+write_volume_align = 1
+
+[host]
+# Default user and password to connect to a host through XAPI
+# Note: this won't be used for SSH.
+# You need to have an SSH key into the hosts' /root/.ssh/authorized_keys.
+default_user = "root"
+default_password = ""
```
gduperrey: I'm not sure about the naming of this category, as it simply lists default credentials. The term "host" doesn't seem the most logical to me here, especially since there is a "hosts" category below where this can be overridden.

lib/config.toml
```diff
@@ -0,0 +1,278 @@
...
+
+[hosts]
+# Hosts to run tests/tools against. Each key is the address (hostname|ip) of
+# the master host of a pool. Settings may also override the global defaults
+# below (repositories, disabled_repositories, hosting_pool) per host.
+# Connection settings (user/password/skip_xo_config) default to the `host`
```
gduperrey: Perhaps explain what "skip_xo_config" is in a single sentence?

lib/config.toml
```diff
@@ -0,0 +1,278 @@
...
+[guest_tools.installed.vendor]
+# Vendor device-specific guest tools
+vendor_device = true
+upgradable = false
+
+[ssh]
+# Public keys for a private key available to the test runner
+pubkey = """ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIMnN/wVdQqHA8KsndfrLS7fktH/IEgxoa533efuXR6rw XCP-ng CI
+ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDKz9uQOoxq6Q0SQ0XTzQHhDolvuo/7EyrDZsYQbRELhcPJG8MT/o5u3HyJFhIP2+HqBSXXgmqRPJUkwz9wUwb2sUwf44qZm/pyPUWOoxyVtrDXzokU/uiaNKUMhbnfaXMz6Ogovtjua63qld2+ZRXnIgrVtYKtYBeu/qKGVSnf4FTOUKl1w3uKkr59IUwwAO8ay3wVnxXIHI/iJgq6JBgQNHbn3C/SpYU++nqL9G7dMyqGD36QPFuqH/cayL8TjNZ67TgAzsPX8OvmRSqjrv3KFbeSlpS/R4enHkSemhgfc8Z2f49tE7qxWZ6x4Uyp5E6ur37FsRf/tEtKIUJGMRXN XCP-ng CI"""
+# Maximum number of SSH output lines to log before truncating
```
gduperrey: Even though they are only public keys, it bothers me that these public keys, which are internal to our company, are being pushed publicly to the Internet via GitHub like this.

README.md
```diff
@@ -147,10 +147,122 @@ For Guest UEFI Secure Boot tests, the requirements are:
...
+loads `lib/config.toml` first, then merges `../../inventory/cgt1-qcow2.toml` on top.
+The `include` key inside any config file is resolved against the file's own
+directory first, then against the main xcp-ng-tests directory, so an overlay
+in another directory can reuse files from the tests repository.
+
+This allows you to:
+- Commit `lib/config.toml` with project defaults to version control
+- Create `config.local.toml` locally (ignored by git) for your standard environment
+- Create `config.prod.toml`, `config.ci.toml`, etc. for other environments and select with `--config=prod` (or `XCPNG_CONFIG=prod`)
+- Point `--config` (or `XCPNG_CONFIG`) at an overlay file kept anywhere on disk
```
gduperrey: I'm not sure I follow here. Is there a risk that this file contains information specific to our infrastructure?

I have to say, I don't find that very clear.

README.md
```diff
@@ -237,8 +350,6 @@ We wanted the job definitions to be in this git repository, that's why the job d
 
 To use `./jobs.py`, you also need to populate `vm_data.py` to define the VM groups that are necessary to run jobs (unless `--vm` is provided on the command line to override the defaults).
```
gduperrey: Question: If we rework data.py, wouldn't we also want to move away from vm_data.py by integrating it into a single configuration file? Would that be possible?

README.md
```diff
@@ -659,47 +780,45 @@ For each pool target :
...
 ```toml
-# my_inventory.toml
+# lib/config.toml
 
-[default]
+[tools.update]
 repositories = ["xcp-ng-base"]
 disabled_repositories = ["epel"]
-hosting_pool = "A"
+hosting_pool = "10.30.0.50"
```
gduperrey: Please assign an IP address such as 10.0.0.2, just as you would for the other IP models.

README.md
```diff
@@ -659,47 +780,45 @@ For each pool target :
...
 [hosts]
-
-[hosts."ip_or_hostname-1"]
-
-[hosts."ip_or_hostname-2"]
-
-repositories = ["xcp-ng-updates"]
-hosting_pool = "B"
+"10.30.0.56" = {}
+"10.30.0.59" = { repositories = ["xcp-ng-updates"] }  # overrides "[tools.update]"
```
gduperrey: Please assign an IP address such as 10.0.0.2/3, just as you would for the other IP models.

README.md
```diff
@@ -659,47 +780,45 @@ For each pool target :
...
-> * `default` is applied to all hosts
-> * Config values under `hosts` override values under `default`. For instance, the above inventory would produce
-> the following python dict:
+> [!NOTE]
+> * `tools.update` is applied to all hosts
+> * Values set in a `[hosts]` entry override `tools.update`. The example above
+>   would produce the following python dict:
 >
-> `{'ip_or_hostname-1': {'repositories': ['xcp-ng-base'], 'disabled_repositories': ['epel'], 'hosting_pool': 'A'}, 'ip_or_hostname-2': {'repositories': ['xcp-ng-updates'], 'hosting_pool': 'B'}}`
+> `{'10.30.0.56': {'repositories': ['xcp-ng-base'], 'disabled_repositories': ['epel'], 'hosting_pool': '10.30.0.50'}, '10.30.0.59': {'repositories': ['xcp-ng-updates'], 'disabled_repositories': ['epel'], 'hosting_pool': '10.30.0.50'}}`
```
gduperrey: Please assign an IP address such as 10.0.0.2, just as you would for the other IP models.

gduperrey (changes requested):

gduperrey:
Would it be possible for you to walk me through how it works during a call?

I’m struggling to see how the benefits you’ve outlined actually fit together compared to the old configuration file. At the end of the day, it’s still a configuration file that needs to be filled out.

And any new user will still have to fill it out, without our internal files and information ending up on the Internet.

lib/config.toml
p0sql: Correct. Even if the public key is not a cryptographic secret, its exposure provides additional information to an adversary and facilitates some operations of recognition and correlation. It gives details on how the key pair was generated (especially with which algorithm). And if our private key were to leak, an adversary could notably use the public key we provide to verify that it matches the compromised private key before testing it on the affected servers: https://nmap.org/nsedoc/scripts/ssh-publickey-acceptance.html.
IMHO, it would be better to keep it in a vault rather than exposing it publicly.

README.md
p0sql: Storing internal ip address may disclose information about our internal infrastructure. Even if it is not directly exploitable from the internet it could be useful to an adversary in the context of his recon activities. Moreover, deleting the ip address from the code is not necessarily enough since it can remain accessible in the commit history.

PR #498, opened 2026-04-24
Enhance block device management

glehmann:
Some block devices may be used to build other devices (mdadm, lvm) or be accessible on multiple hosts, but only usable on a single one at a time (fc, iscsi).

1. `master`
2. #436 → `master`
3. #437 → #436
4. #447 → #437
5. #446 → #447
6. #449 → #446
7. #450 → #449
8. #452 → #450
9. #453 → #452
10. #454 → #453
11. #461 → #454
12. #464 → #461
13. #470 → #464
14. #471 → #470
15. #481 → #471
16. #523 → #481
17. #497 → #481
18. **"Enhance block device management" (this PR) → #497**
19. #500 → #498
20. #509 → #500

### DIFF
--- a/.github/workflows/unit.yml
+++ b/.github/workflows/unit.yml
@@ -1,0 +1,19 @@
+name: Run unit tests
+
+on: [push]
+
+permissions: {}
+
+jobs:
+  unit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+        with:
+          persist-credentials: false
+      - uses: ./.github/actions/uv-setup/
+        with:
+          dev: false
+      - name: Create a dummy data.py
+        run: cp data.py-dist data.py
+      - run: pytest tests/unit/
--- a/conftest.py
+++ b/conftest.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import pytest
 
 import argparse
+import dataclasses
 import itertools
 import logging
 import os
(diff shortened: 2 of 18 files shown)


Millefeuille42 (approved):

lib/host.py
```diff
@@ -669,48 +669,121 @@ def management_pif(self) -> PIF:
...
+                size=int(r['size']),
+                log_sec=int(r.get('log-sec', '0') or '0'),
+                type='mpath',
+                available=len(mountpoint) == 0,
+            ))
+
+        self.block_devices_info = sorted(devices, key=lambda d: d.name)
+        logging.debug("blockdevs found: %s", [d.name for d in self.block_devices_info])
+
+    def _disk_is_available_local(self, disk: str) -> bool:
```
stormi: Declare it before the method that uses it?

conftest.py
```diff
@@ -435,6 +436,42 @@ def _host_disks(host: Host, hosts_cli_disks: list[DiskDevName] | None) -> Iterab
     ret = {host: list(_host_disks(host, cli_disks.get(host.hostname_or_ip)))
            for host in pools_hosts_by_name_or_ip.values()
            }
+    # Cross-host deduplication: a LUN in use on any host (same WWN) is unavailable on all hosts.
```
stormi, line 439: Comment about the commit message: I read it but it didn't explain what problem it solves. Only how it solves it.

Lankou66 (approved):

Millefeuille42 (approved):

.github/workflows/unit.yml
```diff
@@ -0,0 +1,19 @@
...
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+        with:
+          persist-credentials: false
+      - uses: ./.github/actions/uv-setup/
+        with:
+          dev: false
+      - name: Create a dummy data.py
+        run: cp data.py-dist data.py
+      - run: pytest tests/unit/*.py
```
vxgmichel: Maybe follow the [convention for python test discovery](https://docs.pytest.org/en/stable/explanation/goodpractices.html#test-discovery) by renaming `tests/unit/rescan_block_devices_info.py` to `tests/unit/test_rescan_block_devices_info.py`?

```suggestion
      - run: pytest tests/unit
```

vxgmichel (approved):
One suggestion, LGTM otherwise 👍

.github/workflows/unit.yml
glehmann: I voluntarily kept them without the test_ prefix, so they are not run with the other tests, to not add more to the output in jenkins or in our terminal.

Do you think that's a problem?

stormi: Alternatively, we could put then under `/unit` so that they're easier to run from CLI directly without having to think about the *.py stuff.

vxgmichel: Apparently passing a test directory explicitly takes priority over `--ignore`, so another possible approach would be add the following.
```diff
diff --git a/pytest.ini b/pytest.ini
index 3deaa2c..bf08f83 100644
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,5 +1,5 @@
 [pytest]
-addopts = -ra --maxfail=1 --ignore=contrib/
+addopts = -ra --maxfail=1 --ignore=contrib/ --ignore=tests/unit
 markers =
     # *** Markers that change test behaviour ***
     default_vm: mark a test with a default VM in case no --vm parameter was given.
```
This effectively disables the unit tests by default. 

But @stormi suggestion of using a `unit` directory outside of `tests` is probably more explicit. It would require defining `testpaths` though:
```diff
diff --git a/pytest.ini b/pytest.ini
index 3deaa2c..a94601d 100644
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,5 +1,6 @@
 [pytest]
-addopts = -ra --maxfail=1 --ignore=contrib/
+testpaths = tests
+addopts = -ra --maxfail=1
 markers =
```

glehmann: > Apparently passing a test directory explicitly takes priority over `--ignore`

I did that. Running the unit tests is now done with just `pytest tests/unit`.

PR #499, opened 2026-04-24
lib/host: add method to get system uuid

olivierh-pro:
System uuid is the uuid of the host scoped into its physical parent host. It is used in a context of nested hosts.

If the current host is a nested host, snapshot creation must be done through its physical parent host (because the "nested host" is actually a VM).

This PR add a new method in host class to get the system uuid (using `dmidecode -s system-uuid`).

With this uuid, we can then perform a snapshot of the ~~VM~~ nested host.

This work is a preliminiary work for future code changes (#468)

### DIFF
--- a/lib/host.py
+++ b/lib/host.py
@@ -449,6 +449,28 @@ class Host:
         else:
             return self.xe('vm-list', {'uuid': vm_uuid}, minimal=True) == vm_uuid
 
+    def get_system_uuid(self) -> str:
+        """Return system uuid of current host.
+
+        Intended for driving current host from its "parent host" in a **nested context**.
+
+        .. note::
+            If the current host is nested, it means it is not a physical host. It is a VM living inside a real host.::
+
+                [PH: Physical Host] -> [VM: emulation of an XCP-ng host] -> [vm: a vm inside nested host]
+                                       |      current working host     |
+
+            So we need system-uuid of current working host (`VM`) which is
+            the uuid seen in physical host's (`PH`) scope.
+
+        Performs the following command::
+
+            dmidecode -s system-uuid
+
+        ref: `dmidecode(8) <https://man.archlinux.org/man/dmidecode.8.en#s>__`
+        """
+        return self.ssh("dmidecode -s system-uuid").lower().strip()
+
     def yum_clean_metadata(self) -> str:
         """Quietly removes cached metadata on target.
 


lib/host.py
```diff
@@ -449,6 +449,31 @@ def pool_has_vm(self, vm_uuid: str, vm_type: str = 'vm') -> bool:
...
+        Performs the following command::
+
+            dmidecode -s system-uuid
+
+        ref: `dmidecode(8) <https://man.archlinux.org/man/dmidecode.8.en#s>__`
+        """
+        ssh_result = self.ssh_with_result("dmidecode -s system-uuid")
+        system_uuid = str(ssh_result.stdout)
+
+        return system_uuid.lower().rstrip('\n')
```
```suggestion
        return self.ssh("dmidecode -s system-uuid").lower().strip()
```

Or do you want the dmidecode to silently fail?

olivierh-pro: Done!

glehmann (approved):

lib/host.py
```diff
@@ -449,6 +449,28 @@ def pool_has_vm(self, vm_uuid: str, vm_type: str = 'vm') -> bool:
...
+            So we need system-uuid of current working host (`VM`) which is
+            the uuid seen in physical host's (`PH`) scope.
+
+        Performs the following command::
+
+            dmidecode -s system-uuid
+
+        ref: `dmidecode(8) <https://man.archlinux.org/man/dmidecode.8.en#s>__`
+        """
+        return self.ssh("dmidecode -s system-uuid").lower().strip()
```
rzr, line 472:
```suggestion
        return self.ssh("dmidecode -s system-serial-number").lower().strip()
```

I've found a bug while testing it, using SN is aligned to XE uuid (sys uuid has first bytes flipped)


- good (system-serial-number) : 4d8820d1-b103-839e-e3ea-04bfad19e77c
- bad (system-uuid) : d120884d-03b1-9e83-e3ea-04bfad19e77c

cc: @ydirson

rzr (changes requested):

lib/host.py
glehmann: @olivierh-pro I thought we tested it manually, didn't we?

rzr: Forget my suggestion, I think uuid is correct semantically speaking, I think the bug is elsewhere, I am narrowing the problem since I have a reproducer

rzr (approved):

PR #503, opened 2026-04-27
Avoid destroying/unloading the resources before entering the debugger

glehmann:
Same as #436, but for tests not in storage/

### DIFF
--- a/tests/xen/test_ring0.py
+++ b/tests/xen/test_ring0.py
@@ -4,6 +4,7 @@ import logging
 import secrets
 import time
 
+from lib.common import Defer
 from lib.host import Host
 from lib.vm import VM
 
@@ -35,15 +36,13 @@ def host_without_livepatch_loaded(host: Host) -> Host:
     return host
 
 
-def do_execute_xst(host: Host, modname: str, testname: str | None = None) -> None:
+def do_execute_xst(host: Host, modname: str, defer: Defer, testname: str | None = None) -> None:
     if testname is None:
         testname = modname
     host.ssh(f"modprobe xst_{modname}")
-    try:
-        host.ssh(f"echo 1 > /sys/kernel/debug/xst/{testname}/run")
-        host.ssh(f"grep -q 'status: pass' /sys/kernel/debug/xst/{testname}/results")
-    finally:
-        host.ssh(f"modprobe -r xst_{modname}", check=False)
+    defer(lambda: host.ssh(f"modprobe -r xst_{modname}", check=False))
+    host.ssh(f"echo 1 > /sys/kernel/debug/xst/{testname}/run")
+    host.ssh(f"grep -q 'status: pass' /sys/kernel/debug/xst/{testname}/results")
 
 
 @pytest.mark.reboot  # host_with_ring0_tests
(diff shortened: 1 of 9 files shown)


semarie (approved):
LGTM
it will be a good improvement for readability and debugging, thanks.

psafont (approved):
Nice change, it reduces indentation while managing resources, similar to OCaml's  pattern of using `let@ resource = with_resource params in ...`

olivierh-pro (approved):

tests/xen/test_ring0.py
```diff
@@ -52,53 +51,49 @@ class TestRing0Tests:
...
 
-    def test_xst_get_user_pages(self, host: Host) -> None:
+    def test_xst_get_user_pages(self, host: Host, defer: Defer) -> None:
         host.ssh("modprobe xst_get_user_pages")
-        try:
-            host.ssh("/usr/bin/gup_test")
-        finally:
-            host.ssh("modprobe -r xst_get_user_pages", check=False)
+        host.ssh("/usr/bin/gup_test")
+        defer(lambda: host.ssh("modprobe -r xst_get_user_pages", check=False))
```
dinhngtu: Shouldn't the defer move up?

tests/network/test_vif_management.py
```diff
@@ -33,33 +34,29 @@ def test_vif_management(self, running_unix_vm: VM) -> None:
...
+                pass
+            else:
+                raise exc
+
+        # unplug the VIF
+        vif_new.unplug()
+        assert count_interfaces(vm) == n_interfaces
 
         # check one less VIF
         assert len(vm.vifs()) == n_vif
```
dinhngtu: Because the length of `vm.vifs()` is checked after `vif_new.destroy()`, I think converting to defer will change the semantics

dinhngtu (changes requested):
I wonder if there are fixtures that can be trivially replaced by `defer`, and will that net us any benefit in terms of code cleanliness?

tests/xen/test_ring0.py
glehmann: Indeed!

tests/network/test_vif_management.py
glehmann: You're right, I haven't been careful enough when converting this one

dinhngtu: Since defer() is test-scoped while looking like it's block-scoped (unlike C++ RAII), I think we'll run into similar problems in the future, especially if we start looping tests.

glehmann: The semantics should be good, because `vif_new.destroy()` is still there, but the destroy in defer would fail, because the new vif is already destroyed

dinhngtu: Oh, I didn't see that. In that case, why not keep the try/finally since it better represents the desired semantics anyway?

glehmann: Yes, I'm wondering if `defer` actually helps here.
The whole point is to avoid destroying the resources before entering the debugger, so we can inspect them and actually debug some problems.
But in that case, I'm not sure it helps.

I will remove that change. We can convert the test to use `defer` later if we ever need it.

glehmann: done

dinhngtu (approved):

PR #504, opened 2026-04-27
Add cpu policy helper and test

TSnake41:
Add a CPU policy tooling to be able to check for certain CPU feature.
The test_cpu_policy also logs hosts CPU policies for debugging purposes (especially around live migrations).

Required by #393

### DIFF
--- a/lib/cpu_policy.py
+++ b/lib/cpu_policy.py
@@ -1,0 +1,93 @@
+from dataclasses import dataclass
+
+from lib.host import Host
+
+from typing import Literal
+
+NO_SUBLEAF = 0xffffffff
+
+@dataclass(frozen=True)
+class CpuidRegisters:
+    eax: int
+    ebx: int
+    ecx: int
+    edx: int
+
+class CpuPolicy:
+    """A specific CPU policy (PV, HVM, ...)"""
+    # (leaf, subleaf) -> (eax, ebx, ecx, edx)
+    cpuid: dict[tuple[int, int], CpuidRegisters]
+    # msr -> val
+    msr: dict[int, int]
+
+    def __init__(self, cpuid: dict[tuple[int, int], CpuidRegisters], msr: dict[int, int]):
+        self.cpuid = cpuid
+        self.msr = msr
+
+class HostCpuPolicy:
+    """All CPU policies of a host (CPUID, specific MSRs)"""
+    policies: dict[str, CpuPolicy] = {}
+
+    def __init__(self, host: Host):
+        text = host.ssh("xen-cpuid --policy")
+        current_policy: CpuPolicy | None = None
+        mode: Literal["cpuid", "msr"] | None = None
+
+        lines = text.splitlines()
+
+        for line in lines:
+            line = line.rstrip()
+
+            # ---- Policy header ----
+            if "policy:" in line:
+                # Example: "Raw policy: 32 leaves, 2 MSRs"
+                name, rest = line.split("policy:", 1)
+                name = name.strip()
+
+                self.policies[name] = CpuPolicy(dict(), dict())
+                current_policy = self.policies[name]
+                mode = None
+                continue
+
+            if current_policy is None:
+                continue
+
+            # ---- Section switches ----
+            if line.strip() == "CPUID:":
+                mode = "cpuid"
+                continue
+
+            if line.strip() == "MSRs:":
+                mode = "msr"
+                continue
+
+            # Skip table headers
+            if "leaf" in line or "index" in line or not line.strip():
+                continue
+
+            # ---- CPUID parsing ----
+            if mode == "cpuid":
+                # Example:
+                # 00000004:00000003 -> 1c03c163:02c0003f:00001fff:00000006
+                left, right = line.split("->")
+                leaf_hex, subleaf_hex = left.strip().split(":")
+                eax, ebx, ecx, edx = right.strip().split(":")
+
+                key = (int(leaf_hex, 16), int(subleaf_hex, 16))
+                current_policy.cpuid[key] = CpuidRegisters(
+                    int(eax, 16),
+                    int(ebx, 16),
+                    int(ecx, 16),
+                    int(edx, 16)
+                )
+                continue
+
+            # ---- MSR parsing ----
+            if mode == "msr":
+                # Example:
+                # 0000010a -> 400000000c000000
+                idx_hex, val_hex = line.split("->")
+                idx = int(idx_hex.strip(), 16)
+                val = int(val_hex.strip(), 16)
+                current_policy.msr[idx] = val
+                continue
--- a/tests/xen/test_cpu_policy.py
+++ b/tests/xen/test_cpu_policy.py
@@ -1,0 +1,23 @@
+import pytest
+
+import logging
+
+from lib.cpu_policy import NO_SUBLEAF, HostCpuPolicy
+from lib.host import Host
+
+# Xen CPU Policy gathering test
+#
+# # Requirements:
+# - XCP-ng host
+
+class TestCpuPolicy:
+    def test_cpu_policy(self, host: Host) -> None:
+        cpu_policy = HostCpuPolicy(host)
+
+        for name, policy in cpu_policy.policies.items():
+            for ((leaf, subleaf), regs) in policy.cpuid.items():
+                eax, ebx, ecx, edx = regs.eax, regs.ebx, regs.ecx, regs.edx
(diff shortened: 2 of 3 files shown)


tests/misc/test_cpu_policy.py
```diff
@@ -0,0 +1,16 @@
+import pytest
+
+import logging
+
+from lib.cpu_policy import NO_SUBLEAF, HostCpuPolicy
+
+class TestCpuPolicy:
+    def test_cpu_policy(self, host):
```
glehmann:
```suggestion
    def test_cpu_policy(self, host: Host) -> None:
```

To ensure the type checkers check the function.

glehmann (commented):
The test doesn't seem actually to validate anything. Am I missing something?

TSnake41:
> The test doesn't seem actually to validate anything. Am I missing something?

There's nothing to validate aside that it doesn't errors.

lib/cpu_policy.py
```diff
@@ -0,0 +1,83 @@
...
+
+class HostCpuPolicy:
+    """All CPU policies of a host (CPUID, specific MSRs)"""
+    policies: dict[str, CpuPolicy]
+
+    def __init__(self, host: Host):
+        self.policies = dict()
+        text = host.ssh("xen-cpuid --policy")
+        current_policy: CpuPolicy = None
+        mode = None  # None | "cpuid" | "msr"
```
dinhngtu: I think you can use typing.Literal here?

lib/cpu_policy.py
```diff
@@ -0,0 +1,83 @@
...
+    def __init__(self, cpuid: dict[(int, int), int], msr: dict[int, int]):
+        self.cpuid = cpuid
+        self.msr = msr
+
+class HostCpuPolicy:
+    """All CPU policies of a host (CPUID, specific MSRs)"""
+    policies: dict[str, CpuPolicy]
+
+    def __init__(self, host: Host):
+        self.policies = dict()
```
dinhngtu: `policies: dict[str, CpuPolicy] = {}` would be enough IMO

jobs.py
```diff
@@ -469,6 +469,12 @@ class JobData(TypedDict):
             "--vm[]": "multi/limits",
         },
         "paths": ["tests/limits"],
+    },
+    "cpu-policy": {
+        "description": "Tests displaying CPU policies",
+        "nb_pools": 1,
+        "params": {},
+        "paths": ["tests/misc/test_cpu_policy.py"]
     }
```
dinhngtu: Given that there's no special requirement in this job, it can be put into `tests/xen` rather than as a separate job in `misc`.

dinhngtu (changes requested):

lib/cpu_policy.py
```diff
@@ -0,0 +1,83 @@
+from lib.host import Host
+
+NO_SUBLEAF = 0xffffffff
+
+class CpuPolicy:
+    """A specific CPU policy (PV, HVM, ...)"""
+    # (leaf, subleaf) -> (eax, ebx, ecx, edx)
+    cpuid: dict[(int, int), (int, int, int, int)]
```
glehmann:
```suggestion
    cpuid: dict[tuple[int, int], tuple[int, int, int, int]]
```

lib/cpu_policy.py
```diff
@@ -0,0 +1,83 @@
...
+    # msr -> val
+    msr: dict[int, int]
+
+    def __init__(self, cpuid: dict[(int, int), int], msr: dict[int, int]):
```
glehmann:
```suggestion
    def __init__(self, cpuid: dict[tuple[int, int], tuple[int, int, int, int]], msr: dict[int, int]):
```

lib/cpu_policy.py
```diff
@@ -0,0 +1,83 @@
...
+        self.msr = msr
+
+class HostCpuPolicy:
+    """All CPU policies of a host (CPUID, specific MSRs)"""
+    policies: dict[str, CpuPolicy]
+
+    def __init__(self, host: Host):
+        self.policies = dict()
+        text = host.ssh("xen-cpuid --policy")
+        current_policy: CpuPolicy = None
```
glehmann:
```suggestion
        current_policy: CpuPolicy | None = None
```

lib/cpu_policy.py
dinhngtu: @glehmann, is the current recommendation to use native types for typing or the ones from the typing module?

lib/cpu_policy.py
glehmann:
```suggestion
        mode: Literal["cpuid", "msr"] | None = None
```

lib/cpu_policy.py
glehmann: `tuple[int, int, int, int]` might be better represented with

```python

@dataclass(frozen=True)
class CpuidRegisters:
    eax: int
    ebx: int
    ecx: int
    edx: int
```

and used as

```python
def __init__(self, cpuid: dict[tuple[int, int], CpuidRegisters], msr: dict[int, int]):
```
```python
                current_policy.cpuid[key] = CpuidRegisters(
                    int(eax, 16),
                    int(ebx, 16),
                    int(ecx, 16),
                    int(edx, 16)
                )
```

Each element can then be accessed by name:

```python
current_policy.cpuid[key].eax
```

dinhngtu (changes requested):
Please fix the checkers first.

lib/cpu_policy.py
glehmann: > @glehmann, is the current recommendation to use native types for typing or the ones from the typing module?

No, we had diverging opinions on that last time we discussed it.
I'm proposing to go with builtin types in #460

glehmann (approved):

tests/xen/test_cpu_policy.py
```diff
@@ -0,0 +1,18 @@
+import pytest
+
+import logging
+
+from lib.cpu_policy import NO_SUBLEAF, HostCpuPolicy
+from lib.host import Host
+
+class TestCpuPolicy:
+    def test_cpu_policy(self, host: Host) -> None:
```
stormi: This seems to be a test that doesn't test anything. I understand from the commit message that its goal is to log data about the CPUs, but I have no idea when and why this data is supposed to be gathered. Well, no, by following the link to another PR from this PR's description I understand that it's related to https://github.com/xcp-ng/xcp-ng-tests/pull/393, but this explains why the lib was added (which I would probably have done in the other PR), not the test itself.

I assume that you rely on the fact that you put the test in the `xen` directory, and that, for the time being, it will get selected by the current selection criteria for the `xen` job in `jobs.py`, but this seems fragile to me. 

I don't think we should have tests whose sole objective is to log data without testing anything in particular. But if we do, they it should be made very clear in the test description.

tests/xen/test_cpu_policy_collection.py
```diff
@@ -0,0 +1,18 @@
+import pytest
+
+import logging
+
+from lib.cpu_policy import NO_SUBLEAF, HostCpuPolicy
+from lib.host import Host
+
```
stormi, line 7: There's a "requirements" comment here, in test files.

tests/xen/test_cpu_policy.py
TSnake41: It does log thing, but it also does tests, as if we can't gather this information (this is mostly going to happen in a change in xen-cpuid) the test would now fail, instead of making all the tests that depends on cpu policy fail.

stormi: The tests that depend on it would also fail, if executed first.

I guess we can keep it as it is, but it would be good to add explanations as a comment, and rename the test to indicate what it really tests: we don't test the CPU policies, here, we test CPU policies collection.

tests/xen/test_cpu_policy.py
```diff
@@ -0,0 +1,23 @@
...
+from lib.cpu_policy import NO_SUBLEAF, HostCpuPolicy
+from lib.host import Host
+
+# Xen CPU Policy gathering test
+#
+# # Requirements:
+# - XCP-ng host
+
+class TestCpuPolicy:
+    def test_cpu_policy(self, host: Host) -> None:
```
glehmann:
```suggestion
    def test_cpu_policy_collection(self, host: Host) -> None:
```

glehmann (approved):

dinhngtu (approved):

tests/xen/test_cpu_policy.py
stormi: I've been asked for a re-review but this hasn't been addressed. No comment, no renaming of the test. I'll go ahead and do it myself so that we can merge.

stormi (approved):

dinhngtu (approved):

glehmann (approved):

PR #506, opened 2026-04-28
tools/update: add a typed inventory

olivierh-pro:
The dict object representing "Inventory" is growing, so it it more convenient to use TypedDict instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -603,7 +603,7 @@ Take a look at an example inventory file:
 # my_inventory.toml
 
 [all]
-enablerepos = ["xcp-ng-base"]
+repositories = ["xcp-ng-base"]
 
 [servers]
 
@@ -611,13 +611,13 @@ enablerepos = ["xcp-ng-base"]
 
 [servers."ip_or_hostname-2"]
 
-enablerepos = ["xcp-ng-updates"]
+repositories = ["xcp-ng-updates"]
 ```
 
 > [!IMPORTANT]
 > Config values under `servers` override values under `all`. For instance, the above inventory would produce
 > the following python dict:
 >
-> `{'ip_or_hostname-1': {'enablerepos': ['xcp-ng-base']}, 'ip_or_hostname-2': {'enablerepos': ['xcp-ng-updates']}}`
+> `{'ip_or_hostname-1': {'repositories': ['xcp-ng-base']}, 'ip_or_hostname-2': {'repositories': ['xcp-ng-updates']}}`
 >
 > Using *enablerepo flag* `-e` with inventory is still possible, it won't be used though.
--- a/README.md
+++ b/README.md
@@ -605,11 +605,11 @@ Take a look at an example inventory file:
 [all]
 repositories = ["xcp-ng-base"]
 
-[servers]
+[hosts]
 
-[servers."ip_or_hostname-1"]
+[hosts."ip_or_hostname-1"]
 
-[servers."ip_or_hostname-2"]
+[hosts."ip_or_hostname-2"]
 
 repositories = ["xcp-ng-updates"]
 ```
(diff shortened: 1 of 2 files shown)


glehmann (approved):

lib/tools/inventory.py
```diff
@@ -1,15 +1,24 @@
...
 import tomllib
 from pathlib import Path
 
 from lib.common import HostAddress
 
-def load_inventory(inventory_path: Path) -> dict[str, dict[str, list[str]]]:
+from typing import TypeAlias, TypedDict
+
+class Server(TypedDict):
+    enablerepos: list[str]
```
rzr: suggestion: maybe repos would be enough here ?

lib/tools/inventory.py
```diff
@@ -19,24 +28,21 @@ def load_inventory(inventory_path: Path) -> dict[str, dict[str, list[str]]]:
...
-            "enablerepos": repos or all.get("enablerepos", [])
-        }
+        host: Server = {"enablerepos": repos or all.get("enablerepos", [])}
         inventory[server] = host
 
     return inventory
 
-def into_inventory(hosts: list[HostAddress], enablerepos: list[str]) -> dict[HostAddress, dict[str, list[str]]]:
+
+def into_inventory(hosts: list[HostAddress], enablerepos: list[str]) -> Inventory:
```
rzr: suggestion: align to verb_noun pattern  eg: create_inventory , or just verb if in a class like create

lib/tools/inventory.py
```diff
@@ -1,15 +1,24 @@
...
 
 import tomllib
 from pathlib import Path
 
 from lib.common import HostAddress
 
-def load_inventory(inventory_path: Path) -> dict[str, dict[str, list[str]]]:
+from typing import TypeAlias, TypedDict
+
+class Server(TypedDict):
```
rzr: unsure about this typename, suggestion: HostConfiguration ?

lib/tools/inventory.py
olivierh-pro: Done!

lib/tools/inventory.py
olivierh-pro: Nice one! Done!

I created a new commit to rename accordingly

lib/tools/inventory.py
olivierh-pro: This method is intended for converting cli args into an inventory, so the name looks good to me.

glehmann (approved):

lib/tools/inventory.py
```diff
@@ -1,42 +1,48 @@
...
 
-def into_inventory(hosts: list[HostAddress], enablerepos: list[str]) -> dict[HostAddress, dict[str, list[str]]]:
+
+def into_inventory(hosts: list[HostAddress], enablerepos: list[str]) -> Inventory:
     """Create an inventory object from arguments.
 
     Basically, it is used as compatibility when we don't want inventory from file.
     """
-    inventory: dict[HostAddress, dict[str, list[str]]] = {}
+    inventory: Inventory = {}
```
rzr, line 42: I am bit confused about what into is supposed to mean, but let me explain more:

An object is created here right ? so the function is a constructor, without being too pendentic it looks like a "Creational pattern", if you can't create a proper class Inventory constructor, then use create as function verb... 

But now I think I understand more your logic, into is a cast operator to do some type conversion, right ? that would work too but i find it easier to use verbs in function names.

README.md
```diff
@@ -605,11 +605,11 @@ Take a look at an example inventory file:
 [all]
 enablerepos = ["xcp-ng-base"]
```
rzr: I would use repositories here too to make it less yum centric

rzr (approved):
minor suggestions for improvements but otherwise it looks good

PR #509, opened 2026-04-29
partially_populate_device: align span positions to block size for better performance

glehmann:
Add an optional `align` parameter (default: 4KiB) to align span positions
to a given block size. Only positions are aligned (rounded down); span sizes
are reduced if needed to avoid overlapping the next span's position. When
config.write_volume_cap is large enough, spans are contiguous and cover the
full device with no gaps. Without alignment, write throughput drops from
~700MiB/s to ~200MiB/s on large disks.

1. `master`
2. #436 → `master`
3. #437 → #436
4. #447 → #437
5. #446 → #447
6. #449 → #446
7. #450 → #449
8. #452 → #450
9. #453 → #452
10. #454 → #453
11. #461 → #454
12. #464 → #461
13. #470 → #464
14. #471 → #470
15. #481 → #471
16. #523 → #481
17. #497 → #481
18. #498 → #497
19. #500 → #498
20. **"partially_populate_device: align spans to block size for better performance" (this PR) → #500**

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -129,6 +129,13 @@ def pytest_addoption(parser: pytest.Parser) -> None:
         help="Maximum amount of data written to a volume."
              " Accepts sizes like '1GiB', '2.5TiB', or symbolic values 'VHD_MAX', 'QCOW2_MAX'."
     )
+    parser.addoption(
+        "--write-volume-align",
+        action="store",
+        default="1",
+        help="Block size to align span positions to when writing in volumes."
+             " Accepts sizes like '512', '4KiB', '1MiB'. A value of 1 is equivalent to no alignment."
+    )
 
 def pytest_configure(config: pytest.Config) -> None:
     global_config.ignore_ssh_banner = config.getoption('--ignore-ssh-banner')
@@ -141,6 +148,9 @@ def pytest_configure(config: pytest.Config) -> None:
     write_volume_cap = config.getoption('--write-volume-cap')
     assert write_volume_cap is not None
     global_config.write_volume_cap = parse_size(write_volume_cap)
+    write_volume_align = config.getoption('--write-volume-align')
+    assert write_volume_align is not None
+    global_config.write_volume_align = parse_size(write_volume_align)
 
 def pytest_generate_tests(metafunc: pytest.Metafunc) -> None:
     if "vm_ref" in metafunc.fixturenames:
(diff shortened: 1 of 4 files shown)


Lankou66 (approved):

tests/storage/storage.py
```diff
@@ -369,6 +369,64 @@ def validate(self, vm: VM, dev: str) -> None:
         expected_flags = f'--expected-checksum {self.checksum}' if self.checksum is not None else ''
         randstream(vm, f'validate {expected_flags} --position {self.position} --size {self.size} {dev}')
 
+def compute_span_layout(dev_size: int, total_size: int, num_spans: int, align: int) -> list[tuple[int, int]]:
```
Millefeuille42: I'd rather have the argument called `block_size`. Align makes it ambiguous whether it's a flag or a value.

Millefeuille42 (approved):

tests/storage/storage.py
glehmann: done!

conftest.py
```diff
@@ -129,6 +129,13 @@ def pytest_addoption(parser: pytest.Parser) -> None:
         help="Maximum amount of data written to a volume."
              " Accepts sizes like '1GiB', '2.5TiB', or symbolic values 'VHD_MAX', 'QCOW2_MAX'."
     )
+    parser.addoption(
```
stormi, line 132: Would it make sense to also accept a value that means "don't align", and to add a work item so that we'd make sure we test this kind of scenario for performance benchmarks?

glehmann: `--write-volume-align=1`, the default value, means no alignment.
I think we should run the tests with:
-  no alignment (the default value) for the small write caps (also the default value). This is slower than with alignment, but still fast because of the small amount of data written.
- `--write-volume-align=2MiB` for the large write caps, like `--write-volume-cap=QCOW2_MAX`, to make these tests run much faster.

stormi: I read `default="4KiB"` here, and I think argparse will apply this value even if the parameter is not provided on command line. Am I wrong?

stormi: I also more than suggest to indicate what this value means in the help for the command line option.

glehmann: Indeed, that wasn't supposed to be `4KiB`. I see other things in previous PRs that are not at the expected version—I'll update the code.

PR #516, opened 2026-05-05
Add test-ssh-public-key-auth.sh script

vxgmichel:
This PR adds a script to test that older OpenSSH client versions can still connect to an XCP-ng server using public key authentication.

### Usage

```shell
± scripts/test-ssh-public-key-auth.sh 1.2.3.4 7.2_p2-r5 rsa 
Enter root@10.30.36.1 password: 
Pulling Docker image sig9/alpine-openssh-client:7.2_p2-r5...
Generating SSH key pair...
Copying public key to root@1.2.3.4 ...
Verifying SSH connection...
SSH connection successful, cleaning up...
✅ Test passed

```

### Context

Due to the removal of the `ssh-rsa` signature algorithm in OpenSSH 9.8, older clients (version < 7.2) are no longer able to use public key authentication with `ssh-rsa` keys. Howerver, newer OpenSSH versions (7.2 and above) should still be able to authenticate using `ssh-rsa` keys, as well as `ed25519` keys. This script can be used to verify this assumption by running a matrix of tests with different OpenSSH client versions and key types against a specified host. In practice, it should be run manually against a test XCP-ng server whenever the OpenSSH RPM package is updated, to ensure that we maintain compatibility with older clients.

### Requirements

It requires Docker or Podman to be installed.

### Why here? 

It was decided during the daily meeting that this was the right place for test scripts that are not integrated in the CI.

### Why not integrated in the CI?

Having to depend on Docker/Podman has been deemed to heavy for a script that can be run manually whenever a new OpenSSH RPM package is built. An alternative to consider in the future would be to use a VM image from our lab instead of a docker container.

### DIFF
--- a/scripts/test-ssh-public-key-auth.sh
+++ b/scripts/test-ssh-public-key-auth.sh
@@ -1,0 +1,109 @@
+#!/bin/bash
+
+# SSH Public Key Authentication Test
+#
+# This script tests that older OpenSSH client versions can still connect to an
+# XCP-ng server using public key authentication
+#
+# Usage:
+#   test-ssh-public-key-auth.sh <host> <version> <keytype>
+#
+# Context:
+# Due to the removal of the ssh-rsa signature algorithm in OpenSSH 9.8, older
+# clients (version < 7.2) are no longer able to use public key authentication
+# with ssh-rsa keys. Howerver, newer OpenSSH versions (7.2 and above) should
+# still be able to authenticate using ssh-rsa keys, as well as ed25519 keys.
+#
+# This script can be used to verify this assumption by running:
+#
+#     $ scripts/test-ssh-public-key-auth.sh HOST 7.2_p2-r5 rsa
+#
+# It requires Docker or Podman to be installed.
+
+# -e: Exit immediately if a command exits with a non-zero status.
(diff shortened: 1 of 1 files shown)


scripts/test-ssh-public-key-auth.sh
```diff
@@ -0,0 +1,207 @@
...
+    status "Pulling Docker image $image..."
+    if ! silent_unless_fail $CONTAINER_ENGINE pull "$image"; then
+        return 1
+    fi
+
+    status "Testing SSH connection to $USER@$HOST using $image"
+    rm -f "$key_name" "$key_name.pub"
+
+    status "Generating SSH key pair..."
+    if ! silent_unless_fail $CONTAINER_ENGINE run --rm -v "$(pwd):/data" "$image" -c "cd /data && ssh-keygen -t $key_type -f $key_name -C '$comment' -N ''"; then
```
rzr:
```suggestion
    if ! silent_unless_fail $CONTAINER_ENGINE run --rm -v "${PWD}:/data" "$image" -c "cd /data && ssh-keygen -t $key_type -f $key_name -C '$comment' -N ''"; then
```

PR #518, opened 2026-05-07
Add a Host.wait_for_xapi_enabled method

vxgmichel:
From `XCPNG-1061`:
>  `xapi-wait-init-complete` is the right tool, we only use it in one place, should move the code to a helper, audit all other `is_enabled` uses, and replace the relevant ones with the helper

This PR introduces four methods:
- `Host.wait_for_host_down` method
   - use the `ping` logic that was already present in `Host.reboot`
   - refactored with `local_cmd`
- `Host.wait_for_host_up` method
   - use the `ping` logic that was already present in `Host.reboot`
   - refactored with `local_cmd` 
- `Host.wait_for_ssh_reachable` method
   - use the `nc` logic that was already present in `Host.reboot`
   - refactored with `local_cmd` 
- `Host.wait_for_xapi_enabled`
   -  relies on the `xapi-wait-init-complete` command

It also goes through all the existing occurrences of `host.is_enabled`, and updates them according to the context.

## Tests

I've tested the following:
- the `main` job: `uv run jobs.py run main`
- the `test_pool` module: `uv run pytest tests/misc/test_pool.py`

### DIFF
--- a/lib/host.py
+++ b/lib/host.py
@@ -24,6 +24,7 @@ from lib.common import (
     strtobool,
     to_xapi_bool,
     wait_for,
+    wait_for_not,
 )
 from lib.netutil import wrap_ip
 from lib.pif import PIF
@@ -530,7 +531,36 @@ class Host:
         logging.info("Restart toolstack on host %s" % self)
         self.ssh('xe-toolstack-restart')
         if verify:
-            wait_for(self.is_enabled, "Wait for host enabled", timeout_secs=30 * 60)
+            self.wait_for_xapi_enabled()
+
+    def wait_for_host_down(self, timeout_secs: int = 2 * 60) -> None:
+        wait_for_not(
+            lambda: commands.local_cmd(["ping", "-c1", self.hostname_or_ip], check=False).returncode == 0,
+            "Wait for host down",
+            timeout_secs=timeout_secs,
+            retry_delay_secs=2,
+        )
+
+    def wait_for_host_up(self, timeout_secs: int = 10 * 60) -> None:
+        wait_for(
+            lambda: commands.local_cmd(["ping", "-c1", self.hostname_or_ip], check=False).returncode == 0,
+            "Wait for host up",
+            timeout_secs=timeout_secs,
+            retry_delay_secs=10,
+        )
+
+    def wait_for_ssh_reachable(self, timeout_secs: int = 10 * 60) -> None:
+        wait_for(
+            lambda: commands.local_cmd(["nc", "-zw5", self.hostname_or_ip, "22"], check=False).returncode == 0,
+            "Wait for ssh up on host",
+            timeout_secs=timeout_secs,
+            retry_delay_secs=5
+        )
+
+    def wait_for_xapi_enabled(self, timeout_secs: int = 30 * 60) -> None:
+        logging.info(f"Wait for XAPI to complete initialization on {self.hostname_or_ip}")
+        self.ssh(f"xapi-wait-init-complete {timeout_secs}")
+        assert self.is_enabled()
 
     def is_enabled(self) -> bool:
         try:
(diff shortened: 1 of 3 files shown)


lib/host.py
```diff
@@ -530,6 +530,14 @@ def restart_toolstack(self, verify: bool = False) -> None:
...
             self.wait_for_xapi_enabled(timeout_secs=30 * 60)
 
+    def wait_for_host_reachable(self, timeout_secs: int = 10 * 60) -> None:
+        wait_for(
+            lambda: not os.system(f"nc -zw5 {self.hostname_or_ip} 22"),
+            "Wait for ssh up on host",
+            timeout_secs=timeout_secs,
+            retry_delay_secs=5
+        )
+
```
ydirson: Maybe we should have "ssh" in the name. XAPI in v9 starts to default to "ssh disabled" mode, so we may have to do other reachability tests in the future (eg. HTTPS reachable to send test commands using XAPI)

glehmann (approved):

semarie (approved):

lib/host.py
```diff
@@ -530,7 +531,35 @@ def restart_toolstack(self, verify: bool = False) -> None:
...
+    def wait_for_ssh_reachable(self, timeout_secs: int = 10 * 60) -> None:
+        wait_for(
+            lambda: commands.local_cmd(["nc", "-zw5", self.hostname_or_ip, "22"], check=False).returncode == 0,
+            "Wait for ssh up on host",
+            timeout_secs=timeout_secs,
+            retry_delay_secs=5
+        )
+
+    def wait_for_xapi_enabled(self, timeout_secs: int = 30 * 60) -> None:
+        self.ssh(f"xapi-wait-init-complete {timeout_secs}")
```
stormi, line 562: Unless I'm mistaken, from a logging perspective, this is not equivalent to the previous code : we used to say "Wait for XAPI to be ready" or f"Wait for XAPI init to be complete on {master_hostname_or_ip}".

stormi (changes requested):

lib/pool.py
```diff
@@ -30,10 +30,7 @@ def __init__(self, master_hostname_or_ip: HostAddress) -> None:
 
         # wait for XAPI startup to be done, or we can get "Connection
         # refused (calling connect )" when calling self.hosts_uuids()
-        wait_for(lambda: commands.ssh_with_result(master_hostname_or_ip,
-                                                  'xapi-wait-init-complete 60').returncode == 0,
-                 f"Wait for XAPI init to be complete on {master_hostname_or_ip}",
-                 timeout_secs=30 * 60)
+        self.master.wait_for_xapi_enabled()
```
stormi, line 33: Here you replace code that erred on the safe side of things, not assuming that the Host object be completely ready to run methods. You're removing this security. I will assume that you're doing this on purpose and that reviewers have noticed the change, but then it seems to me that an explanation is lacking from the commit message.

CC @ydirson as this was your code initially.

See https://github.com/xcp-ng/xcp-ng-tests/commit/e21613f2b757623d6d2f3209d6af7055b38f0e0e

vxgmichel: Yes, it's been discussed with @ydirson .

The explanation is that `self.master` is already available at this point since the corresponding `Host` has been instantiated (line 25) and `master.is_master()` was called (line 26). Both of those operations require a successful ssh command to return (`cat /etc/xensource/pool.conf` and `cat /etc/xensource-inventory`). It was then decided that the `wait_for` was superfluous in this case.

I'll add this explanation in the commit message

lib/host.py
vxgmichel: Good catch, I'll fix this 👍

lib/pool.py
vxgmichel: Added here: 4f736dd73fe4e73b87a54189919e0ecc69b64969

lib/host.py
vxgmichel: Fixed here: 4f736dd73fe4e73b87a54189919e0ecc69b64969

glehmann (approved):

stormi (approved):

ydirson (approved):

PR #522, opened 2026-05-11
Add VS Code/Codium extensions recommendations for Python development

glehmann:
This could help the non-Python developers working on the project get as much help as possible from their IDE.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,10 @@
+{
+	"recommendations": [
+	  "detachhead.basedpyright",
+	  "ms-python.flake8",
+	  "charliermarsh.ruff",
+	  "bellini666.pytest-language-server"
+	],
+	"unwantedRecommendations": [
+	]
+}


.vscode/extensions.json
```diff
@@ -0,0 +1,9 @@
+{
+	"recommendations": [
+	  "detachhead.basedpyright",
```
olivierh-pro, line 3: Since we also use `mypy`, would you consider adding this one https://marketplace.visualstudio.com/items?itemName=ms-python.mypy-type-checker ?

glehmann: I'm not sure that's a good idea to have 2 extensions doing almost the same thing. It sure isn't in helix, where we have duplicated entries. Have you tried in vscodium?

mypy is way slower (in helix) when used in the editor, so I preferred recommending basedpyright.

olivierh-pro (approved):

stormi (approved):

PR #525, opened 2026-05-12
Add basic tracing functionality to tests

contificate:
Opening as a draft to receive commentary.

---

This simple change introduces a `--tracing-endpoint` option to the test framework that, if provided, will augment the host setup fixture to create an observer on each host. The effect of this is that toolstack operations will be traced, with their spans submitted to the provided endpoint (either Zipkin, Jaeger, or compatible span aggregator).

---

A simple test is provided that uses this to perform an operation and then query the endpoint to fetch the root span of the operation (with some metadata attached).

image

---

The overall idea is that the benchmarking/performance teams can modify extant tests (or write new tests) to retrofit tracing (in a non-intrusive, opt-in, way).

I'm seeking some feedback about the direction of this change as there's more to consider:

- Tests with complex pool steps may want to have better identifying metadata by default. So, you can run a test and immediately place the separate hosts' span timeline together for debugging.
- The `Host` object's `xe` method could be modified to allow for environmental variables. Then, as done in the provided sample test (albeit manually), we could supply `BAGGAGE` to attach metadata to the spans of specific toolstack operations (relying on the metadata to fetch and process the related spans later).

---

Any feedback will be much appreciated. This change is very simple and could do with being made a lot more featureful and robust, but I'm not overly familiar with the testing framework or its tests, so I've kept it small and isolated for now.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -116,6 +116,12 @@ def pytest_addoption(parser: pytest.Parser) -> None:
         default="2GiB",
         help="Maximum amount of data written to a volume"
     )
+    parser.addoption(
+        "--tracing-endpoint",
+        action="store",
+        default=None,
+        help="Specify distributed tracing endpoint."
+    )
 
 def pytest_configure(config: pytest.Config) -> None:
     global_config.ignore_ssh_banner = config.getoption('--ignore-ssh-banner')
@@ -192,6 +198,19 @@ def pytest_runtest_makereport(
 
 # END make test results visible from fixtures
 
+def setup_tracing(host: Host, endpoint: str):
+    logging.info(f'Enabling tracing on {host}, endpoint = {endpoint}')
+    host.ssh('printf "observer-endpoint-http-enabled=true\nobserver-experimental-components=\\"\\"\n" > /etc/xapi.conf.d/observer.conf')
+    host.restart_toolstack(verify=True)
+    observer_uuid = host.xe('observer-create', {'name-label': 'xcpng-test', 'endpoints': endpoint, 'enabled': 'true'})
+    host.restart_toolstack(verify=True)
+    return observer_uuid
+
+def teardown_tracing(host, observer):
+    logging.info(f'Disabling tracing on {host}')
+    host.xe('observer-destroy', {'uuid': observer})
+    host.ssh('rm -f /etc/xapi.conf.d/observer.conf')
+    host.restart_toolstack(verify=True)
 
 # fixtures
 
@@ -259,7 +278,19 @@ def hosts(pytestconfig: pytest.Config) -> Generator[list[Host], None, None]:
 
     if not host_list:
         pytest.fail("This test requires at least one --hosts parameter")
+
+    tracing_endpoint = pytestconfig.getoption("--tracing-endpoint")
+    has_tracing = tracing_endpoint is not None
+    observers = []
+    if has_tracing:
+        for h in host_list:
+            observers.append(setup_tracing(h, tracing_endpoint))
+
     yield host_list
+
+    if has_tracing:
+        for h, o in zip(host_list, observers):
+            teardown_tracing(h, o)
 
     cleanup_hosts()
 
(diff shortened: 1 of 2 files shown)


conftest.py
```diff
@@ -259,8 +278,20 @@ def cleanup_hosts() -> None:
 
     if not host_list:
         pytest.fail("This test requires at least one --hosts parameter")
+
+    tracing_endpoint = pytestconfig.getoption("--tracing-endpoint")
+    has_tracing = tracing_endpoint is not None
+    observers = []
```
stormi, line 284: There's just one observer per host, right? So rather than maintaining a list of observers, I think each host should have its optional observer as an attribute, and maybe the setup and teardown parts should be handled directly in setup_host and cleanup_hosts.

conftest.py
```diff
@@ -192,6 +198,19 @@ def pytest_runtest_makereport(
...
+    host.restart_toolstack(verify=True)
+    observer_uuid = host.xe('observer-create', {'name-label': 'xcpng-test', 'endpoints': endpoint, 'enabled': 'true'})
+    host.restart_toolstack(verify=True)
+    return observer_uuid
+
+def teardown_tracing(host, observer):
+    logging.info(f'Disabling tracing on {host}')
+    host.xe('observer-destroy', {'uuid': observer})
+    host.ssh('rm -f /etc/xapi.conf.d/observer.conf')
+    host.restart_toolstack(verify=True)
```
stormi, line 213: I think these could be added as methods of the Host object.

conftest.py
gthvn1: I think you can have several observers per host with different endpoint for example.

stormi: You can, but that's not what's in the current implementation, and I suppose that if/when we need more than one observer per host we can have a list or dict of observers attached to the host.

conftest.py
```diff
@@ -116,6 +116,12 @@ def pytest_addoption(parser: pytest.Parser) -> None:
         default="2GiB",
         help="Maximum amount of data written to a volume"
     )
+    parser.addoption(
+        "--tracing-endpoint",
+        action="store",
+        default=None,
+        help="Specify distributed tracing endpoint."
```
LuKP17, line 123: Maybe add examples of endpoints here for users?

conftest.py
LuKP17: From my understanding of the relationship between observers and hosts, it's the other way around: observers contain an optional list of hosts uuids. If this list is empty then the observer is pool-wide, we provide hosts uuids if we need to filter which hosts in the pool are affected by this observer (most importantly it dictates which ones are not).
So for migration tests with a pool of 2 hosts, there's no need to specify hosts when creating an observer, we will get the source and dest hosts traces by default (pool-wide observer). But for some tests we would maybe like only one host to be affected by the observer, then add this host's uuid in the observer's host list.
If an observer is a standalone XAPI object, maybe we can rethink how to reflect this in xcp-ng-tests (like introducing an Observer class).

conftest.py
```diff
@@ -192,6 +198,19 @@ def pytest_runtest_makereport(
 
 # END make test results visible from fixtures
 
+def setup_tracing(host: Host, endpoint: str):
+    logging.info(f'Enabling tracing on {host}, endpoint = {endpoint}')
+    host.ssh('printf "observer-endpoint-http-enabled=true\nobserver-experimental-components=\\"\\"\n" > /etc/xapi.conf.d/observer.conf')
+    host.restart_toolstack(verify=True)
+    observer_uuid = host.xe('observer-create', {'name-label': 'xcpng-test', 'endpoints': endpoint, 'enabled': 'true'})
```
LuKP17, line 205: If we don't provide the host's uuid in `host-uuids` param, we will end up with multiple pool-wide observers if the function is called multiple times, which is not what was intended here I think.

LuKP17 (commented):
Since there is interest in this PR and we would like to have tracing support for the live migration benchmarking project, we agreed with Colin to reorganize the code into a proper solution. I will open my own PR when I'm done, which will surely introduce an Observer class in `/lib` as mentioned in [one of my comments](https://github.com/xcp-ng/xcp-ng-tests/pull/525#discussion_r3388066298).

PR #527, opened 2026-05-12
Fix auto-install tests

ydirson:
Various regressions have accumulated in the last months, as this test is not regularly scheduled in Jenkins.

Tested in the `tests/install/test-sequences/` configuration on 8.3.0 june '25 ISO

### DIFF
--- a/tests/install/conftest.py
+++ b/tests/install/conftest.py
@@ -11,7 +11,7 @@ import xml.etree.ElementTree as ET
 from data import ARP_SERVER, ISO_IMAGES, ISO_IMAGES_BASE, ISO_IMAGES_CACHE, TEST_SSH_PUBKEY, TOOLS
 from lib import installer, pxe
 from lib.commands import local_cmd
-from lib.common import Defer, callable_marker, url_download, wait_for
+from lib.common import callable_marker, url_download, wait_for
 from lib.installer import AnswerFile
 
 from typing import TYPE_CHECKING, Any, Generator, Sequence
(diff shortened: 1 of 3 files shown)


olivierh-pro (approved):

rzr (approved):

tests/install/conftest.py
```diff
@@ -285,10 +284,10 @@ def vm_booted_with_installer(host: Host, create_vms: list[VM], remastered_iso: s
 
     remote_iso = host.pool.push_iso(iso)
     host_vm.insert_cd(os.path.basename(remote_iso))
-    defer(lambda: host.pool.remove_iso(remote_iso))
+    lambda: host.pool.remove_iso(remote_iso)
```
glehmann: I don't think that remove_iso would ever be called here.

tests/install/conftest.py
```diff
@@ -285,10 +284,10 @@ def vm_booted_with_installer(host: Host, create_vms: list[VM], remastered_iso: s
...
 
     host_vm.start()
-    defer(lambda: host_vm.shutdown(force=True))
+    lambda: host_vm.shutdown(force=True)
```
glehmann: same here

glehmann: It could work with
```python
    defer(lambda: host_vm.shutdown(force=True) if not host_vm.is_halted() else None)
```

I think `defer` is useful here to keep the resource in the same state as when the test failed, i.e. avoid shutting down the VM before entering the debugger.
If you prefer it without defer, it's probably safer to restore the exact version before 09c4d6e653df094cf20f8f643d92cd2ac6a2c79c

glehmann (changes requested):
The patch reverting the use of `defer` doesn't seem valid to me.
The rest looks good

tests/install/conftest.py
ydirson: > The patch reverting the use of `defer` doesn't seem valid to me.

Right, I messed up on this one.

> If you prefer it without defer, it's probably safer to restore the exact version before https://github.com/xcp-ng/xcp-ng-tests/commit/09c4d6e653df094cf20f8f643d92cd2ac6a2c79c

Right, I don't think Defer will bring us anything here, the shutdowns were originally handled by the fixture teardown, which has the intended and explicit behaviour.

ydirson: Actually the issue is that those shutdown only need to be issued on exception, while using `Defer` makes them systematic.

glehmann (approved):

PR #510, opened 2026-04-30
pre-commit: add config for checks on commit

olivierh-pro:
At the moment, the [Makefile](https://github.com/xcp-ng/xcp-ng-tests/blob/master/Makefile) is used as an entry point for running the code quality checks locally. However, there exists tools specifically designed to run code checkers. Those tools are widely adopted by open-source projects, especially in the python community. The most common approach is to use [pre-commit](https://pre-commit.com/) or [prek](https://github.com/j178/prek) to leverage so-called [git hooks](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks) in order to make sure that each commit is compliant with the quality standard of a given project. 

> [!NOTE]
> **The use of this tool is completely optional.**
> It is up to the developer to decide whether to install the git hooks locally or not. 

This PR adds a `.pre-commit-config.yaml` configuration file that can be used with [prek](https://prek.j178.dev/). It also adds [prek](https://prek.j178.dev/) to the `dev` dependencies so that the tool can be easily invoked using `uv run prek`.

**How do I use it?**

Find below a longer version of the *Pre-commit hook* section that is being added to the README:

---

#### General

The repository has a [prek](https://prek.j178.dev/) configuration, which allows for making sure that each commit created is compliant with the [code checking implemented in the CI](https://github.com/xcp-ng/xcp-ng-tests/blob/master/.github/workflows/code-checkers.yml). This feature is **opt-in** and can be run manually.

This tool is configured via the [`.pre-commit-config.yaml`](./.pre-commit-config.yaml) file.

In order to run all the checks that the CI performs, you can simply run:
```shell
$ uv run prek -a
ruff.....................................................................Passed
flake8...................................................................Passed
mypy.....................................................................Passed
pyright..................................................................Passed
```

#### Install git hooks

[prek](https://prek.j178.dev/) really shines once you start to run it automatically before each commit.
For this, we need to tell [prek](https://prek.j178.dev/) to install a [git hook](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks) in our local repository:
```bash
$ uv run prek install
```

Note that you can easily remove this hook by running:
```bash
$ uv run prek uninstall
```
  
Once the hooks are in place, `git commit` will automatically:
- stash the changes that are not staged for this commit
- perform the `ruff`, `mypy`, `flake8` and `pyright` checks (unless the staged changes do not affect any python files, in which case the checks are skipped)
- re-apply the changes that were stashed

```bash
$ git commit -m "my message"
[...]

PR #528, opened 2026-05-13
Use prek in the CI to run the code checks

vxgmichel:
This ensures consistency between pre-commit checks and the CI

Follow-up after #510

### DIFF
--- a/.github/workflows/code-checkers.yml
+++ b/.github/workflows/code-checkers.yml
@@ -4,6 +4,9 @@ on: [push]
 
 permissions: {}
 
+env:
+  PREK_COLOR: "always"
+
 jobs:
   mypy:
     runs-on: ubuntu-latest
@@ -16,7 +19,7 @@ jobs:
         run: cp data.py-dist data.py
       - name: Create a dummy vm_data.py
         run: cp vm_data.py-dist vm_data.py
-      - run: mypy --install-types --non-interactive .
+      - run: prek -a mypy
 
   pyright:
     runs-on: ubuntu-latest
@@ -29,7 +32,7 @@ jobs:
         run: cp data.py-dist data.py
       - name: Create a dummy vm_data.py
         run: cp vm_data.py-dist vm_data.py
-      - run: pyright .
+      - run: prek -a pyright
 
   ruff:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/code-checkers.yml
olivierh-pro:
```suggestion
  prek:
```

.github/workflows/code-checkers.yml
```diff
@@ -5,7 +5,7 @@ on: [push]
 permissions: {}
 
 jobs:
-  mypy:
+  prek:
     runs-on: ubuntu-latest
```
olivierh-pro, line 12:
```suggestion
    runs-on: ubuntu-latest
    env:
      FORCE_COLOR: "1"
```

Should we keep this env var like in old version?

.github/workflows/code-checkers.yml
vxgmichel: Done: a9fcf23761aeabed93fb1fe067081a855a9e7c5f

.github/workflows/code-checkers.yml
vxgmichel: Done: a9fcf23761aeabed93fb1fe067081a855a9e7c5f

olivierh-pro (approved):

glehmann (approved):
Grouping all the checks in a single job means we loose a bit of visibility when something fails, but that looks acceptable fo me

vxgmichel:
> Grouping all the checks in a single job means we loose a bit of visibility when something fails, but that looks acceptable for me

Right. It's possible to provide specific hooks to run to prek using `uv run prek run -a flake8` for instance, so an alternative would be to configure one job per hook.
- Pros: the failing hook is shown directly on the PR CI report
- Cons: we have to remember to change the CI config when adding/removing/renaming hooks

Both are fine with me.

**EDIT**: I'm investigating an alternative

.github/workflows/code-checkers.yml
```diff
@@ -16,39 +19,4 @@ jobs:
         run: cp data.py-dist data.py
```
rzr, line 19: What about scripting this in makefile, then developer can run exactly what CI is doing ?

vxgmichel: I think this a good argument for https://github.com/xcp-ng/xcp-ng-tests/pull/491
It would be great if our tools did not require this copy before they can run on a fresh clone.

olivierh-pro (approved):

glehmann (approved):

PR #534, opened 2026-05-18
Mark Broken tests

rzr:
Recent CI runs shown that some tests can not work as expected , instead of disabling them on cmdline, it thought it's better to keep trace of status, until they are fixed again.


glehmann (changes requested):
I've already excluded this SR in the jenkins config, as it was done for moosefs.
Also it's not only this file, but the whole dir that should be excluded.

stormi (approved):
Approving despite disapproving, to help with current CI jobs. We must fix this properly afterwards.

glehmann (approved):
Also disapproving! ;)

PR #535, opened 2026-05-19
Use NFS for intrapool migration in the glusterfs tests

vxgmichel:
The following tests are failing:
- test_glusterfs_sr_intrapool_migration.py::Test::test_cold_intrapool_migration
- test_glusterfs_sr_intrapool_migration.py::Test::test_live_intrapool_migration

With either one of those errors depending on the fixture resolution order:
- [AssertionError: xfsprogs must not be installed on the host at the beginning of the tests](https://github.com/xcp-ng/xcp-ng-tests/blob/377e8bd9d6ba0149728c5f4e535f55fb5fef84cf/tests/storage/conftest.py#L69)
- [AssertionError: There is already a saved package list set](https://github.com/xcp-ng/xcp-ng-tests/blob/488cbdc1258cc0e0a38bcf45ba39a7cb065a0491/lib/host.py#L621)[AssertionError: There is already a saved package list set](https://github.com/xcp-ng/xcp-ng-tests/blob/488cbdc1258cc0e0a38bcf45ba39a7cb065a0491/lib/host.py#L621) 

This is due to both mangled package installation in underlying fixtures and the lack of enough unused disks for both SR.

@glehmann, @rzr and I concluded that using NFS was much simpler and less error prone to test the glusterfs intrapool migration

### DIFF
--- a/tests/storage/glusterfs/test_glusterfs_sr_intrapool_migration.py
+++ b/tests/storage/glusterfs/test_glusterfs_sr_intrapool_migration.py
@@ -1,9 +1,13 @@
 import pytest
 
+from lib import config
 from lib.host import Host
 from lib.sr import SR
+from lib.vdi import ImageFormat
 from lib.vm import VM
 from tests.storage import cold_migration_then_come_back, live_storage_migration_then_come_back
+
+from typing import Generator
 
 # Requirements:
 # From --hosts parameter:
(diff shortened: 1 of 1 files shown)


rzr (approved):
I would put text in the PR also in the commit message. It will help the future you.

update: I did it for you and merged to test it.

glehmann (approved):

Lankou66 (approved):

PR #536, opened 2026-05-19
Add autopep8 code formatter integration

glehmann:
autopep8 is an autoformatter that focuses on fixing pep8 formatting
problems, and doesn't impose a hard coded coding style as would do black
or ruff formatter.
For example, if the author chose to use split a line at a specific
position, ruff or black might decide to reformat the code in their own
way, but autopep8 won't, as long as it respects the pep8.
So it gives the author the option to automatically fix pep8 problems but
still allows to format the code as they think is the most readable.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,6 +7,12 @@ repos:
         require_serial: true
         entry: "uv run ruff check"
         types: ["python"]
+      - id: "autopep8"
+        name: "autopep8"
+        language: "system"
+        require_serial: true
+        entry: "uv run autopep8 -i"
+        types: ["python"]
       - id: "flake8"
         name: "flake8"
         language: "system"
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,7 +5,7 @@ repos:
         name: "ruff"
         language: "system"
         require_serial: true
-        entry: "uv run ruff check"
+        entry: "uv run ruff check --fix --show-fixes"
         types: ["python"]
       - id: "autopep8"
         name: "autopep8"
--- a/Makefile
+++ b/Makefile
@@ -9,16 +9,16 @@ data.py:
 	@test -r data.py || echo "File 'data.py' does not exist. Refer to https://github.com/xcp-ng/xcp-ng-tests#configuration." && exit 1
 
 mypy: data.py
-	uv run mypy --install-types --non-interactive .
+	uv run prek -a mypy
 
 pyright: data.py
-	uv run pyright .
+	uv run prek -a pyright
 
 ruff: data.py
-	FORCE_COLOR=1 uv run ruff check --fix
+	uv run prek -a ruff
 
 flake8:
-	uv run flake8
+	uv run prek -a flake8
 
 autopep8:
-	uv run autopep8 -ir --exit-code .
+	uv run prek -a autopep8
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ pyright: data.py
 	uv run pyright .
 
 ruff: data.py
-	FORCE_COLOR=1 uv run ruff check
+	FORCE_COLOR=1 uv run ruff check --fix
 
 flake8:
 	uv run flake8
--- a/Makefile
+++ b/Makefile
@@ -8,17 +8,7 @@ all: ruff autopep8 flake8 mypy pyright
 data.py:
 	@test -r data.py || echo "File 'data.py' does not exist. Refer to https://github.com/xcp-ng/xcp-ng-tests#configuration." && exit 1
 
-mypy: data.py
-	uv run prek -a mypy
+mypy pyright ruff: data.py
 
-pyright: data.py
-	uv run prek -a pyright
-
-ruff: data.py
-	uv run prek -a ruff
-
-flake8:
-	uv run prek -a flake8
-
-autopep8:
-	uv run prek -a autopep8
+ruff autopep8 flake8 mypy pyright:
+	uv run prek -a $@
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 
 .PHONY: all mypy pyright ruff flake8
 
-all: mypy pyright ruff flake8
+all: ruff autopep8 flake8 mypy pyright
 
 data.py:
 	@test -r data.py || echo "File 'data.py' does not exist. Refer to https://github.com/xcp-ng/xcp-ng-tests#configuration." && exit 1
@@ -19,3 +19,6 @@ ruff: data.py
 
 flake8:
 	uv run flake8
+
+autopep8:
+	uv run autopep8 -ir --exit-code .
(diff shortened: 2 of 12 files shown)


.pre-commit-config.yaml
```diff
@@ -7,6 +7,12 @@ repos:
         require_serial: true
         entry: "uv run ruff check"
         types: ["python"]
+      - id: "autopep8"
+        name: "autopep8"
+        language: "system"
+        require_serial: true
+        entry: "uv run autopep8 -ir ."
```
olivierh-pro: Maybe we should use the same config as `mypy` and `pyright`, using `pass_filenames: false`.

https://prek.j178.dev/reference/configuration/?h=pass#pass_filenames

@vxgmichel what do you think?

glehmann: I think the formatting is not affected by the other files, so it should be faster to run only on the modified files.
Note that this is not what I configured…

vxgmichel: We can run autopep8 only on staged python files using:
```suggestion
        entry: "uv run autopep8 -i"
```

vxgmichel (changes requested):

.pre-commit-config.yaml
vxgmichel: Ooops I already replied here: https://github.com/xcp-ng/xcp-ng-tests/pull/536#discussion_r3267236350

glehmann: Yes I was looking into that. But would it run on all the files in the CI?

glehmann: Yes it would, `-a` is for `--all-files` in `prek -a autopep8`

lib/host.py
```diff
@@ -173,7 +173,7 @@ def stringify(key: str, value: str | bool | dict[str, str]) -> str:
             return f'{key}={shlex.quote(value)}'
 
         command: str = f'xe {action} {maybe_param_minimal} {maybe_param_force} ' + \
-                       ' '.join(stringify(key, value) for key, value in args.items())
+            ' '.join(stringify(key, value) for key, value in args.items())
```
ydirson, line 176: I would rather see that kind of thing handled by grouping with parentheses to get rid of the backslash-continuation. I'm kinda surprised modern checker do not reject that old widely-frowned-upon syntax

tests/security/test_tls.py
```diff
@@ -29,7 +29,7 @@ def test_tls_disabled(host: Host, protocol_name: str) -> None:
     with pytest.raises(ssl.SSLError):
         context = ssl.SSLContext(protocol)
         with socket.create_connection((str(host), PORT), timeout=10) as sock, \
-             context.wrap_socket(sock, server_hostname=str(host)) as ssock:
+                context.wrap_socket(sock, server_hostname=str(host)) as ssock:
```
ydirson, line 32: same, `with` can now work with parentheses (several occurrences of this one, won't repeat)

.pre-commit-config.yaml
ydirson: In the CI it maybe the check should be different than in precommit?  Since there is a single check for the whole branch it should likely take all files changed by the PR instead.

vxgmichel: > Since there is a single check for the whole branch it should likely take all files changed by the PR instead.

I've seen some projects do that, but that was mainly to automate the push of a fixing commit. In our case, I think it's simpler to have the CI run `prek` on all files (that's how I configured it #528 at least)

lib/host.py
glehmann: This is not what `autopep8` would do on its own. We can do that manually, and autopep8 would let it through, but in future modifications in other places, autopep8 would introduce a backslash instead of parentheses.

.pre-commit-config.yaml
glehmann: IMO, the check should stay green for all files, and the run is already very fast (1-2 seconds).
Is there a specific reason to restrict the check to the files modified in the PR?

vxgmichel (approved):
I wonder if it's useful to maintain the `Makefile` now that we rely on the pre-commit configuration for the checks. Maybe we should remove it? LGTM otherwise 👍

.pre-commit-config.yaml
ydirson: Not necessarily here in fact. That kind of thing is often useful when the whole codebase does not pass initially, to allow for progressive application of new rules.

lib/host.py
ydirson: Quite strange, given that PEP8 itself discourage such backslash use when not strictly necessary: https://peps.python.org/pep-0008/#maximum-line-length

glehmann: I don't see an option to tweak that in autopep8.

Note: ruff is doing it by default ;)

.pre-commit-config.yaml
```diff
@@ -5,7 +5,13 @@ repos:
         name: "ruff"
         language: "system"
         require_serial: true
-        entry: "uv run ruff check"
+        entry: "uv run ruff check --fix --show-fixes"
```
olivierh-pro, line 8: I'm not very confortable with idea of autofix here because we use this hook in CI.

What does it imply to run `--fix` with github actions?
* 1/ Does Github operate a fix commit for us?
or
* 2/ Does the job fail as usual and force us to fix in local?

cc @vxgmichel

ydirson: If `--fix` exits in error when it finds something would need fixing, I guess it's good in the CI

vxgmichel: > 1/ Does Github operate a fix commit for us?

No, [some projects do that](https://github.com/fastapi/fastapi/blob/f4cafbc467c225263ad3b5b0d4a7306b42ac855b/.github/workflows/pre-commit.yml#L57-L72) but we don't.

> 2/ Does the job fail as usual and force us to fix in local?

> If --fix exits in error when it finds something would need fixing, I guess it's good in the CI

Yes that's exactly how it works. 

Also note that this behavior is the one we would get by using [the autopep8 pre-commit repo](https://github.com/pre-commit/mirrors-autopep8), see the `-i` [here](https://github.com/pre-commit/mirrors-autopep8/blob/ff31678cfc7e06898ba918005b1f6f3099b46a66/.pre-commit-hooks.yaml#L7). My point here is that `pre-commit` hooks are typically designed to work this way (although the [ruff pre-commit repo](https://github.com/astral-sh/ruff-pre-commit) makes it optional for `ruff check`)

ydirson (approved):
lgtm, and we can deal with the horrendous line continuations in another PR :)

Makefile
```diff
@@ -3,19 +3,22 @@
...
 ruff: data.py
-	FORCE_COLOR=1 uv run ruff check
+	uv run prek -a ruff
 
 flake8:
-	uv run flake8
+	uv run prek -a flake8
+
+autopep8:
+	uv run prek -a autopep8
```
ydirson: BTW, this can likely be factorized, like (untested):
```
mypy pyright ruff: data.py

mypy pyright ruff flake8 autopep8:
        uv run prek -a $@
```

or even, since the available rules are already enumerated as depedencies of `all`:
```
%:
        uv run prek -a $@

mypy pyright ruff: data.py
```

PR #540, opened 2026-05-20
Use vm.detect_package_manager() to check if apk is available

glehmann:

1. `master`
2. **"Use vm.detect_package_manager() to check if apk is available" (this PR)**
3. #542

### DIFF
--- a/lib/vm.py
+++ b/lib/vm.py
@@ -500,12 +500,11 @@ class VM(BaseVM):
         """ Heuristic to determine the package manager on a unix distro. """
         if self.file_exists('/usr/bin/rpm') or self.file_exists('/bin/rpm'):
             return PackageManagerEnum.RPM
-        elif self.file_exists('/usr/bin/apt-get'):
+        if self.file_exists('/usr/bin/apt-get'):
             return PackageManagerEnum.APT_GET
-        elif self.file_exists('/sbin/apk'):
+        if self.file_exists('/sbin/apk'):
             return PackageManagerEnum.APK
-        else:
-            return PackageManagerEnum.UNKNOWN
+        return PackageManagerEnum.UNKNOWN
 
     def insert_cd(self, vdi_name: str) -> None:
         logging.info("Insert CD %r in VM %s", vdi_name, self.uuid)
--- a/tests/storage/storage.py
+++ b/tests/storage/storage.py
@@ -5,7 +5,7 @@ from dataclasses import dataclass
 
 from lib import config
 from lib.commands import SSHCommandFailed
-from lib.common import QCOW2_MAX, VHD_MAX, Defer, GiB, MiB, TiB, strtobool, wait_for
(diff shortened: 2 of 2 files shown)


olivierh-pro (approved):

lib/vm.py
```diff
@@ -502,6 +502,8 @@ def detect_package_manager(self) -> PackageManagerEnum:
             return PackageManagerEnum.RPM
         elif self.file_exists('/usr/bin/apt-get'):
             return PackageManagerEnum.APT_GET
+        elif self.file_exists('/sbin/apk'):
+            return PackageManagerEnum.APK
         else:
             return PackageManagerEnum.UNKNOWN
```
Wescoeur:
```suggestion
        if self.file_exists('/usr/bin/apt-get'):
            return PackageManagerEnum.APT_GET
        if self.file_exists('/sbin/apk'):
            return PackageManagerEnum.APK
        return PackageManagerEnum.UNKNOWN
```

glehmann: One line shorter, I like it :)
Done.

lib/vm.py
```diff
@@ -500,10 +500,11 @@ def detect_package_manager(self) -> PackageManagerEnum:
...
         if self.file_exists('/usr/bin/rpm') or self.file_exists('/bin/rpm'):
             return PackageManagerEnum.RPM
-        elif self.file_exists('/usr/bin/apt-get'):
+        if self.file_exists('/usr/bin/apt-get'):
             return PackageManagerEnum.APT_GET
-        else:
-            return PackageManagerEnum.UNKNOWN
+        if self.file_exists('/sbin/apk'):
+            return PackageManagerEnum.APK
+        return PackageManagerEnum.UNKNOWN
```
ydirson, line 507: Although it's not directly related to this PR, I'm not particularly fond of this heuristics.

* for one, `rpm` and `apt` are not of the same level, don't we want to test for `yum` or `dnf` instead of `rpm`?
* as much as `rpm` is available in Debian, `apt` is also [available in various RPM-based distros](https://www.rpmfind.net/linux/rpm2html/search.php?query=apt&submit=Search)

olivierh-pro: IMO, pyinfra is a good example of how to figure out the package manager ([here](https://github.com/pyinfra-dev/pyinfra/blob/3.x/src/pyinfra/operations/server.py#L751)).

Under the hood, it is using `command -v <command> || true` ([ref](https://github.com/pyinfra-dev/pyinfra/blob/3.x/src/pyinfra/facts/server.py#L196)).

Maybe it can help for inspiration.

glehmann: Indeed that's not the same level, we should probably fix that.

Wescoeur (approved):

lib/vm.py
glehmann: In progress as part of https://github.com/xcp-ng/xcp-ng-tests/pull/570

PR #542, opened 2026-05-20
Move root partition resizing logic to VM.grow_root_partition()

glehmann:
And make it a bit more robust, by fetching the actual partition mounted on `/`

Also add support for other Linux distributions than Alpine.

1. `master`
2. #540
3. **"Move root partition resizing logic to VM.grow_root_partition()" (this PR)**

### DIFF
--- a/lib/vm.py
+++ b/lib/vm.py
@@ -526,8 +526,11 @@ class VM(BaseVM):
         return PackageManagerEnum.UNKNOWN
 
     def grow_root_partition(self) -> int | None:
-        if self.detect_package_manager() == PackageManagerEnum.APK:
+        pkg_manager = self.detect_package_manager()
+        if pkg_manager == PackageManagerEnum.APK:
             self.ssh('apk add util-linux e2fsprogs-extra')
+        elif pkg_manager == PackageManagerEnum.APT_GET:
+            self.ssh('apt-get update && apt-get install -y -qq util-linux e2fsprogs')
         else:
             return None
         mount_output = self.ssh('mount').strip()
--- a/lib/vm.py
+++ b/lib/vm.py
@@ -531,6 +531,8 @@ class VM(BaseVM):
             self.ssh('apk add util-linux e2fsprogs-extra')
         elif pkg_manager == PackageManagerEnum.APT_GET:
             self.ssh('apt-get update && apt-get install -y -qq util-linux e2fsprogs')
+        elif pkg_manager == PackageManagerEnum.RPM:
+            self.ssh('yum install -y util-linux e2fsprogs')
         else:
             return None
         mount_output = self.ssh('mount').strip()
--- a/lib/vm.py
+++ b/lib/vm.py
@@ -527,13 +527,7 @@ class VM(BaseVM):
 
     def grow_root_partition(self) -> int | None:
         if self.detect_package_manager() == PackageManagerEnum.APK:
-            # growpart is not available in alpine 3.12
-            # vm.ssh('apk add cloud-utils-growpart e2fsprogs-extra')
-            self.ssh('apk add gawk util-linux e2fsprogs-extra')
-            self.ssh(
-                'wget https://raw.githubusercontent.com/canonical/cloud-utils/main/bin/growpart -O /usr/bin/growpart'
-            )
-            self.ssh('chmod +x /usr/bin/growpart')
+            self.ssh('apk add util-linux e2fsprogs-extra')
         else:
             return None
         mount_output = self.ssh('mount').strip()
@@ -543,8 +537,8 @@ class VM(BaseVM):
         if not fs_type.startswith('ext'):
             logging.debug(f"Unsupported filesystem: {fs_type}")
             return None
-        growpart_returncode = self.ssh_with_result(f'growpart /dev/{disk} {partition}').returncode
-        assert growpart_returncode in [0, 1] # growpart returns 1 if the size is already the expected one
+        self.ssh(f'echo ", +" | sfdisk --no-reread --force -N {partition} /dev/{disk}')
+        self.ssh(f'partx -u -n {partition}:{partition} /dev/{disk}')
         self.ssh(f'resize2fs /dev/{disk}{p}{partition}')
         df_output = self.ssh('df /')
         return int(df_output.splitlines()[-1].split()[3]) * KiB
--- a/lib/vm.py
+++ b/lib/vm.py
@@ -4,6 +4,7 @@ import pytest
 
 import logging
 import os
+import re
 import subprocess
 import tempfile
 import uuid
@@ -12,6 +13,7 @@ import lib.commands as commands
 import lib.efi as efi
 from lib.basevm import BaseVM
 from lib.common import (
+    KiB,
     PackageManagerEnum,
     expand_scope_relative_nodeid,
     parse_xe_dict,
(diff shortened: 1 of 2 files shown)


ydirson (commented):
It would be good to add more context, not everyone is expected to know what this partition resizing is about

lib/vm.py
```diff
@@ -506,6 +508,30 @@ def detect_package_manager(self) -> PackageManagerEnum:
...
 
+    def grow_root_partition(self) -> int | None:
+        if self.detect_package_manager() == PackageManagerEnum.APK:
+            # growpart is not available in alpine 3.12
+            # vm.ssh('apk add cloud-utils-growpart e2fsprogs-extra')
+            self.ssh('apk add gawk util-linux e2fsprogs-extra')
+            self.ssh(
+                'wget https://raw.githubusercontent.com/canonical/cloud-utils/main/bin/growpart -O /usr/bin/growpart'
+            )
+            self.ssh('chmod +x /usr/bin/growpart')
```
ydirson: This still gives an external dependency on known-flaky github, I'm not sure we want to have more of them.
Is there any particular reason to use such an external tool?  Wouldn't `parted` just avoid this?

glehmann: It doesn't add a dependency, it's already there.

glehmann: I removed growpart and used sfdisk and partx instead, which were already dependencies

olivierh-pro (commented):
LGTM!

olivierh-pro (approved):

lib/vm.py
```diff
@@ -506,6 +508,29 @@ def detect_package_manager(self) -> PackageManagerEnum:
...
+    def grow_root_partition(self) -> int | None:
+        pkg_manager = self.detect_package_manager()
+        if pkg_manager == PackageManagerEnum.APK:
+            self.ssh('apk add util-linux e2fsprogs-extra')
+        elif pkg_manager == PackageManagerEnum.APT_GET:
+            self.ssh('apt-get update && apt-get install -y -qq util-linux e2fsprogs')
+        elif pkg_manager == PackageManagerEnum.RPM:
+            self.ssh('yum install -y util-linux e2fsprogs')
+        else:
+            return None
```
Millefeuille42: I'd suggest to use pattern matching as introduced by [PEP634](https://docs.python.org/3.10/whatsnew/3.10.html#pep-634-structural-pattern-matching):

Optionally I'd also suggest throwing in an [`assert_never`](https://docs.python.org/3.11/library/typing.html#typing.assert_never) to make sure all cases are handled if we ever add one in the future.

```suggestion
from typing import assert_never
[...]
        pkg_manager = self.detect_package_manager()
        match pkg_manager:
            case PackageManagerEnum.APK:
                self.ssh('apk add util-linux e2fsprogs-extra')
            case PackageManagerEnum.APT_GET:
                self.ssh('apt-get update && apt-get install -y -qq util-linux e2fsprogs')
            case PackageManagerEnum.RPM:
                self.ssh('yum install -y util-linux e2fsprogs')
            case PackageManagerEnum.UNKNOWN:
                return None
            case _:
                assert_never(pkg_manager)
```

lib/vm.py
```diff
@@ -506,6 +508,29 @@ def detect_package_manager(self) -> PackageManagerEnum:
...
+        mount_output = self.ssh('mount').strip()
+        root_match = re.search(r'/dev/(\w+?)(p?)(\d+) on / type (\w+)', mount_output)
```
Millefeuille42, line 546: I'd suggest to replace spaces by `\s+` to make it more robust at the price of readability.
```suggestion
        root_match = re.search(r'/dev/(\w+?)(p?)(\d+)\s+on\s+/\s+type\s+(\w+)', mount_output)
```

lib/vm.py
```diff
@@ -506,6 +508,29 @@ def detect_package_manager(self) -> PackageManagerEnum:
...
+        assert root_match is not None
+        disk, p, partition, typ = root_match.groups()
```
Millefeuille42: I'd prefer to have `typ` to be named more explicitly (like `mount_type` or `fs_type`, etc) instead of just `typ` to avoid collision with `type`.

lib/vm.py
```diff
@@ -506,6 +508,29 @@ def detect_package_manager(self) -> PackageManagerEnum:
...
+        if not typ.startswith('ext'):
+            logging.debug(f"Unsupported filesystem: {typ}")
+            return None
+        self.ssh(f'echo ", +" | sfdisk --no-reread --force -N {partition} /dev/{disk}')
+        self.ssh(f'partx -u -n {partition}:{partition} /dev/{disk}')
+        self.ssh(f'resize2fs /dev/{disk}{p}{partition}')
```
Millefeuille42, line 554: I'd prefer to have the variable parts to be put through`shlex.quote` or put in an array and through `shlex.join()` to prevent any issue with ill-formed parameters.

Example:
```python
self.ssh(f'echo ", +" | sfdisk --no-reread --force -N {shlex.quote(partition)} {shlex.quote(f"/dev/{disk}")}')

# OR

self.ssh(shlex.join(["partx", "-u", "-n", f"{partition}:{partition}", f"/dev/{disk}"]))
```

Millefeuille42 (changes requested):

lib/vm.py
glehmann: Have you noticed some OS where it could have more than one space?

lib/vm.py
glehmann: The values are matched just above with the regex. The chances of having ill-formed parameters seem to be approximately null. Am I missing something?

lib/vm.py
glehmann: I'm usually not a fan of Python's match: it's usually more indented and takes 1 more line.
Detecting a missing case with the type checker could make it useful, though!

lib/vm.py
glehmann: done!

lib/vm.py
glehmann: done

Millefeuille42: > I'm usually not a fan of Python's match: it's usually more indented and takes 1 more line. Detecting a missing case with the type checker could make it useful, though!

I can understand, it's not the best syntax IMHO either. However I really like it due to the "safe" factor of failing on unhandled cases!

lib/vm.py
Millefeuille42: Not necessarily, indeed if you just matched the patterns it's safe but I'd rather have it applied systematically to prevent forgetting it on a case where it's useful, I'd accept not using `shlex` without any problem if you don't agree.

lib/vm.py
Millefeuille42: Not necessarily, but this makes it safer and resilient to any change about the type of space / number of spaces if any.
I'd accept keeping as it is if you don't agree.

Millefeuille42 (approved):
The two of my remaining comments aren't blocking, but must be resolved either way (applying the suggestion or not) before merging.

AnthoineB (approved):

lib/vm.py
glehmann: I will favor readability in that case–that's a tradeoff, I'm just picking one side!

lib/vm.py
glehmann: Here, I also pick the readability. If it were in a security-sensitive code, I would also properly escape everything, but here it's better to keep things more readable IMO

PR #546, opened 2026-05-20
jobs.py: temporarily exclude glusterfs from storage-migrations

stormi:
There's a fixture problem which makes Host.yum_save_state() twice, which is not supported.

Temporary fix to allow urgent CI runs to resume.

### DIFF
--- a/jobs.py
+++ b/jobs.py
@@ -172,7 +172,7 @@ JOBS: dict[str, JobData] = {
         },
         "paths": ["tests/storage"],
         "markers": "not unused_4k_disks",
-        "name_filter": "migration and not linstor",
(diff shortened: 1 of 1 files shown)


jobs.py
```diff
@@ -172,7 +172,7 @@ class JobData(TypedDict):
         },
         "paths": ["tests/storage"],
         "markers": "not unused_4k_disks",
-        "name_filter": "migration and not linstor",
+        "name_filter": "migration and not linstor and not glusterfs", # FIXME: glusterfs temporarily excluded
```
rzr, line 175: and lvmohba if we want to align to previous commit i made

PR #551, opened 2026-05-22
commands: expose ssh's own errors separately via ssherr field

glehmann:
Use ssh's -E option to redirect ssh's own diagnostic messages to a temp file, separating them from the remote command's stdout/stderr. Expose the captured text as a new ssherr field on SSHResult and SSHCommandFailed, and log it at DEBUG level after each command.

### DIFF
--- a/lib/commands.py
+++ b/lib/commands.py
@@ -5,6 +5,7 @@ import logging
 import os
 import platform
 import subprocess
+import tempfile
 
 import lib.config as config
 from lib.netutil import wrap_ip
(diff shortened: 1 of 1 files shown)


lib/commands.py
```diff
@@ -102,26 +109,44 @@ def _ssh(
...
         opts += ['-o', 'ControlMaster no']
 
-    ssh_cmd = ['ssh', f'root@{hostname_or_ip}'] + opts + [cmd]
-
     # Fetch banner and remove it to avoid stdout/stderr pollution.
     banner_res = None
     if config.ignore_ssh_banner:
-        banner_res = subprocess.run(
-            ['ssh', f'root@{hostname_or_ip}'] + opts + ['\n'],
+        _, banner_log_file = tempfile.mkstemp(suffix='.log', prefix='ssh_err_banner_')
```
stormi: Doesn't tempfile offer a class with support for `with` and automated tempfile deletion once done? Would this be less robust than try... finally...?

lib/commands.py
```diff
@@ -102,26 +109,44 @@ def _ssh(
...
+        try:
+            banner_res = subprocess.run(
+                ['ssh', f'root@{hostname_or_ip}'] + opts + ['-E', banner_log_file] + ['\n'],
+                stdout=subprocess.PIPE,
+                stderr=subprocess.STDOUT,
+                check=False
+            )
+            with open(banner_log_file) as f:
+                banner_ssherr = f.read()
```
stormi: Is it useful to do this for every SSH call rather than just when the return code is 255?

stormi: Hmm, yes, since we want the banner text to ignore it.

lib/commands.py
glehmann: Indeed, I first used mkstemp to avoid indenting a lot of code and then used it in a try-finally anyway, containing a small amount of code.
I'll switch to a context manager

glehmann: done

stormi (approved):
Assuming that it was tested after the last push, LGTM.

vxgmichel (approved):
I'm curious, what kind of errors should we expect in `ssherr`, do you have an example? When do they occur?
LGTM otherwise 👍

PR #554, opened 2026-05-25
fix: early check with --hosts is missing

lmolina:
--hosts defaults to [], not None, so `assert is not None` could never fire. Assert now checks for empty list.

Disclaimer: it passes the static checks using pre-commit. I haven't run the actual tests. I'm not familiar with the code base.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -256,7 +256,7 @@ def hosts(pytestconfig: pytest.Config) -> Generator[list[Host], None, None]:
 
     # a list of master hosts, each from a different pool
     hosts_args = pytestconfig.getoption("hosts")
-    assert hosts_args is not None
+    assert hosts_args, "This test requires at least one --hosts parameter"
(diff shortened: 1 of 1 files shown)


conftest.py
```diff
@@ -268,7 +268,7 @@ def cleanup_hosts() -> None:
         raise
 
     if not host_list:
```
lmolina, line 270: Maybe `setup_host` does something relevant that deserves this check? If not, I think that this check is also redundant and may be removed.
```suggestion
```

conftest.py
```diff
@@ -268,7 +268,7 @@ def cleanup_hosts() -> None:
...
-        pytest.fail("This test requires at least one --hosts parameter")
+        pytest.fail("This test requires at least one valid host")
```
lmolina, line 271:
```suggestion
```

PR #558, opened 2026-05-29
storage: move create/destroy SR tests to create_destroy sub-packages

glehmann:
pytest treats each directory with __init__.py as a separate package, so
package-scoped fixtures get independent instances for create_destroy/ tests
vs. the main SR tests. This eliminates disk conflicts where a
TestXxxSRCreateDestroy test would destroy the SR while a package-scoped
fixture from TestXxxSR still held a reference to it.
No fixture scopes were changed.

This is an alternative to #557

### DIFF
--- a/tests/storage/ext/create_destroy/test_create_destroy_ext_sr.py
+++ b/tests/storage/ext/create_destroy/test_create_destroy_ext_sr.py
@@ -1,0 +1,36 @@
+from __future__ import annotations
+
+import pytest
+
+from lib.common import vm_image
+from lib.host import Host
+from lib.vdi import ImageFormat
+from tests.storage import try_to_create_sr_with_missing_device
+
+# Requirements:
+# - one XCP-ng host with an additional unused disk for the SR
+
+class TestEXTSRCreateDestroy:
+    """
+    Tests that do not use fixtures that setup the SR or import VMs,
+    because they precisely need to test SR creation and destruction,
+    and VM import.
+    """
+
+    def test_create_sr_with_missing_device(self, host: Host) -> None:
+        try_to_create_sr_with_missing_device('ext', 'EXT-local-SR-test', host)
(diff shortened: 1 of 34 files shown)


tests/storage/ext/create_destroy/test_create_destroy_ext_sr.py
```diff
@@ -0,0 +1,36 @@
...
+    Tests that do not use fixtures that setup the SR or import VMs,
+    because they precisely need to test SR creation and destruction,
+    and VM import.
+    """
+
+    def test_create_sr_with_missing_device(self, host: Host) -> None:
+        try_to_create_sr_with_missing_device('ext', 'EXT-local-SR-test', host)
+
+    def test_create_and_destroy_sr(self, host: Host,
+                                   unused_512B_disks: dict[Host, list[Host.BlockDeviceInfo]],
```
stormi, line 24: Do package-scoped fixtures such as this one work inside the expected boundaries despite not being imported directly this create_destroy package's conftest.py?

stormi (approved):
I see mostly code moved, I agree with the rationale. 

Not verified in detail so this is a conditional approval:
* package scope working as intended inside create_destroy tests (what if for example we run all create_destroy tests together without the rest of the storage tests?)
* Test requirement comments accurate for each SR type. The few I checked were fine, but I haven't check them all.

tests/storage/ext/create_destroy/test_create_destroy_ext_sr.py
glehmann: Yes it seems to work properly. `host_with_xfsprogs` for example is setup and teared down when entering or leaving each package.

Note: `unused_512B_disks` is a session-scoped fixture

Wescoeur (approved):

PR #562, opened 2026-06-02
Disable autofix on default make invocation

vxgmichel:
Here's a proposal, following up @stormi's comment:
> Please offer a simple way to verify your files without altering them automatically

```shell
# Run all checks without modifications
$ make                                               
uv run prek -a ruff
ruff.....................................................................Passed
uv run prek -a autopep8
autopep8.................................................................Passed
uv run prek -a flake8
flake8...................................................................Passed
uv run prek -a mypy
mypy.....................................................................Passed
uv run prek -a pyright
pyright..................................................................Passed


# Run all the auto-fix hooks
± make fix
uv run prek -a ruff-fix
ruff-fix.................................................................Passed
uv run prek -a autopep8-fix
autopep8-fix.............................................................Passed
```
Note: the pre-commit hook behavior hasn't changed:
```shell
$ git commit
ruff-fix.................................................................Passed
autopep8-fix.............................................................Passed
flake8...................................................................Passed
mypy.....................................................................Passed
pyright..................................................................Passed
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,14 +1,20 @@
 # Makefile to run static code checkers locally
 # Equivalent to the GitHub Actions workflow in .github/workflows/code-checkers.yml
 
-.PHONY: all mypy pyright ruff flake8
+.PHONY: all mypy pyright ruff ruff-fix flake8 autopep8 autopep8-fix
 
-all: ruff autopep8 flake8 mypy pyright
+# By default, only run the non-fix version of the hooks so it doesn't modify any files
+# It runs on all files managed by git (untracked files are not checked)
+check: ruff autopep8 flake8 mypy pyright
+
+# Use the `fix` directive to let autopep8 auto-format the code and ruff sort the imports
+# It runs on all files managed by git (untracked files are not modified)
+fix: ruff-fix autopep8-fix
 
 data.py:
 	@test -r data.py || echo "File 'data.py' does not exist. Refer to https://github.com/xcp-ng/xcp-ng-tests#configuration." && exit 1
 
 mypy pyright ruff: data.py
 
-ruff autopep8 flake8 mypy pyright:
+ruff ruff-fix autopep8 autopep8-fix flake8 mypy pyright:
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -1,14 +1,15 @@
 # Makefile to run static code checkers locally
 # Equivalent to the GitHub Actions workflow in .github/workflows/code-checkers.yml
 
-.PHONY: all mypy pyright ruff flake8
+.PHONY: all mypy pyright ruff ruff-fix flake8 autopep8
 
-all: ruff autopep8 flake8 mypy pyright
+# Do not run ruff-fix or autopep8 by default, so it doesn't modify any file
+all: ruff flake8 mypy pyright
```
ydirson: If `autopep8` is dropped too, maybe a more general `fix` target would be useful

ydirson (commented):
Another approach (simpler UX-wise) would be to have those fix commands protected from touching an unstaged file, by adding some `git` guard command to fail the check before running it (when the command cannot be told to do it by itself)

Makefile
vxgmichel: Done: [6bb460a](https://github.com/xcp-ng/xcp-ng-tests/pull/562/commits/6bb460a58fc64509af1c6a40f310814eda2bc9ce)

vxgmichel: But maybe `all` is not a good default target name now that it only performs the checks, would you prefer the following?
```make
check: ruff flake8 mypy pyright
fix: ruff-fix autopep8
all: fix check
```

vxgmichel: Done: [874047d](https://github.com/xcp-ng/xcp-ng-tests/pull/562/commits/874047d0d8898f27213965cca074d87b488ef468)

glehmann (approved):

stormi (changes requested):
LGTM but the readme was forgotten.

glehmann (approved):

stormi (approved):

PR #563, opened 2026-06-02
During install tests, avoid VM being shutdown before PDB can run

ydirson:
When booting a nested host with an installer ISO, in case of failure to get to the state where we can read it using ssh, we could not inspect the host once `--pdb` stopped the test, because the exception was firing from fixture setup code.

### DIFF
--- a/tests/install/conftest.py
+++ b/tests/install/conftest.py
@@ -280,8 +280,15 @@ def iso_in_pool_isosr(host: Host, remastered_iso: str) -> Generator[str, None, N
     host.pool.remove_iso(remote_iso)
 
 @pytest.fixture(scope='function')
-def vm_booted_with_installer(host: Host, create_vms: list[VM], iso_in_pool_isosr: str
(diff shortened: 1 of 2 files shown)


vxgmichel (commented):
I think this change is very useful, but I agree with @glehmann that it might done in a way that is more straight forward. 

In particular, having `(try_)booting_vm_installer` no longer failing when it could not boot gives the wrong impression that this might be expected for some specific tests. As far as I understand this is not the case, but a way to address the valid issue of being able to drop in a PDB before the cleanup is performed when the booting fails.

Instead, we might want to add a `defer/request.addFinalizer` to `booting_vm_installer`, i.e remove the `try-except-finally` statements and use something like this instead:
```python
    def cleanup():
        if host_vm is not None:
            host_vm.shutdown(force=True)
        if remote_iso:
            host.pool.remove_iso(remote_iso)
    defer(cleanup)
```
This should achieve the expected result.

PR #567, opened 2026-06-04
Tools/update - xapi script for system UUID

olivierh-pro:
Using `dmidecode -s system-uuid` is quite buggy (cc @rzr ).

This PR change the way of getting system uuid. Instead of using shell command (`dmidecode`), it pushes a python script onto remote server containing xapi commands.

We perform the same operation as with dmidecode but with xapi.

This PR is related to #468 . system uuid is needed to perform a snapshot of VM (nested host) from hosting pool.

### DIFF
--- a/lib/host.py
+++ b/lib/host.py
@@ -454,10 +454,17 @@ class Host:
         else:
             return self.xe('vm-list', {'uuid': vm_uuid}, minimal=True) == vm_uuid
 
-    def get_system_uuid(self) -> str:
-        """Return system uuid of current host.
+    def get_system_serial_number(self) -> str:
+        """Get system serial number.
 
-        Intended for driving current host from its "parent host" in a **nested context**.
+        Connect to host and execute a raw python script to
+        get system serial number. This function is quite
+        equivalent to the command: `dmidecode -s system-uuid`.
+
+        >>> my_host.get_serial_system()
+        1392c56b-5f8b-47a3-a077-b17900fd7922
+
+        Intended for driving current host from its "hosting pool" in a **nested context**.
 
         .. note::
             If the current host is nested, it means it is not a physical host. It is a VM living inside a real host.::
@@ -465,16 +472,30 @@ class Host:
                 [PH: Physical Host] -> [VM: emulation of an XCP-ng host] -> [vm: a vm inside nested host]
                                        |      current working host     |
 
-            So we need system-uuid of current working host (`VM`) which is
-            the uuid seen in physical host's (`PH`) scope.
-
-        Performs the following command::
-
-            dmidecode -s system-uuid
-
-        ref: `dmidecode(8) <https://man.archlinux.org/man/dmidecode.8.en#s>__`
+            So we need system serial number of current working host (`VM`) which is
+            the uuid seen in physical host's (`PH`) scope, the hosting pool.
         """
-        return self.ssh("dmidecode -s system-uuid").lower().strip()
+        raw_python_script = f"""
+import XenAPI
+def main():
+    session = XenAPI.xapi_local()
+    try:
+        session.xenapi.login_with_password('root', '')
+        current_host = session.xenapi.host.get_by_uuid('{self.uuid}')
+        uuid = session.xenapi.host.get_bios_strings(current_host)['system-serial-number']
+    except Exception:
+        pass
+    finally:
+        session.xenapi.session.logout()
+    return uuid
+if __name__ == '__main__':
+    print(main())
+       """
+        system_serial_number = self.execute_script(raw_python_script, shebang='python')
+        if not system_serial_number:
+            raise ValueError("System serial number has incorrect value.")
+
+        return system_serial_number
 
     def yum_clean_metadata(self) -> str:
         """Quietly removes cached metadata on target.
--- a/lib/tools/tasks/update.py
+++ b/lib/tools/tasks/update.py
@@ -61,5 +61,5 @@ def update_pools(inventory: Inventory) -> None:
     # Snapshot creation
     for hosting_pool, nested in nested_hosts.items():
         pool = Pool(hosting_pool) # mandatory for getting an host instance
-        vm_uuids = [h.get_system_uuid() for h in nested]
+        vm_uuids = [h.get_system_serial_number() for h in nested]
         create_snapshots(pool.master, vm_uuids)


lib/tools/tasks/common.py
stormi, line 1: > Using dmidecode -s system-uuid is quite buggy

Can you elaborate?

lib/tools/tasks/common.py
```diff
@@ -0,0 +1,39 @@
+"""Common module
+
+This module provides common utils shared between other tasks.
+"""
+from lib.host import Host
+
+def get_system_serial(master: Host) -> str:
```
stormi: 1. This is singular (serial, not serials) but the description is plural. The naming is not appropriate.
2. Semantically, you're looking for data about the pool not about the master host.  We have a Pool class.

lib/tools/tasks/update.py
```diff
@@ -61,5 +62,6 @@ def update_pools(inventory: Inventory) -> None:
     # Snapshot creation
     for hosting_pool, nested in nested_hosts.items():
         pool = Pool(hosting_pool) # mandatory for getting an host instance
-        vm_uuids = [h.get_system_uuid() for h in nested]
+        vm_uuids = [get_system_serial(h) for h in nested]
+        vm_uuids = list(filter(lambda h: h or False, vm_uuids)) # don't want empty uuid
```
glehmann: Why would a system serial be empty? Isn't it a problem if it is?

Also I think a list comprehension would be nicer here:

```python
        vm_uuids = [uuid for uuid in vm_uuids if uuid] # don't want empty uuid
```

lib/tools/tasks/common.py
olivierh-pro: I've made a reword accordingly. This function operates on each host (nested).

lib/tools/tasks/common.py
rzr: Actually dmidecode is not buggy but can not be used because of a bug underneath,  after a few investigation I suspect the root cause is related to this bit flipflop done by XEN_DEFINE_UUID  from:

https://xenbits.xen.org/gitweb/?p=xen.git;a=blob;f=xen/include/public/xen.h;h=b12fd10e6315430755e0eae9300e574af818ef70;hb=HEAD#l995

```
/*
 * XEN_DEFINE_UUID(0x00112233, 0x4455, 0x6677, 0x8899,
 *                 0xaa, 0xbb, 0xcc, 0xdd, 0xee, 0xff)
 * will construct UUID 00112233-4455-6677-8899-aabbccddeeff presented as
 * {0x00, 0x11, 0x22, 0x33, 0x44, 0x55, 0x66, 0x77, 0x88,
 * 0x99, 0xaa, 0xbb, 0xcc, 0xdd, 0xee, 0xff};
 *
 * NB: This is compatible with Linux kernel and with libuuid, but it is not
 * compatible with Microsoft, as they use mixed-endian encoding (some
 * components are little-endian, some are big-endian).
 */
#define XEN_DEFINE_UUID_(a, b, c, d, e1, e2, e3, e4, e5, e6)            \
    {{((a) >> 24) & 0xFF, ((a) >> 16) & 0xFF,                           \
      ((a) >>  8) & 0xFF, ((a) >>  0) & 0xFF,                           \
      ((b) >>  8) & 0xFF, ((b) >>  0) & 0xFF,                           \
      ((c) >>  8) & 0xFF, ((c) >>  0) & 0xFF,                           \
      ((d) >>  8) & 0xFF, ((d) >>  0) & 0xFF,                           \
                e1, e2, e3, e4, e5, e6}}
```
It looks like xapi is handling this correctly to reorder bytes the MS way, but it's might be missing elsewhere (I stopped the investigation there), isnt this @xcp-ng/hypervisor-kernel  issue already reported ? if not I can file a ticket

dinhngtu: On Linux, the SMBIOS uuid byte order is buggy as you've pointed out.

I think you can use `/sys/hypervisor/uuid`?

lib/tools/tasks/update.py
olivierh-pro: I've replaced with a list comprehension, it is clearer, thank you!

Yes it is a problem if uuid is empty because it will create a side effect if we launch an xe command like this `xe vm-snapshot new-name-label=foo uuid=''`.

In my code (raw script), I always return empty string on error. I can't ensure xapi will not fail here.

lib/tools/tasks/common.py
olivierh-pro: Thank for your tip @dinhngtu . Unfortunately, `dmidecode -s system-uuid` and `cat /sys/hypervisor/uuid` have a different result! :(

dinhngtu: `/sys/hypervisor/uuid` should give the right UUID corresponding to XAPI VM UUID on Linux VMs, or at least it does on my Debian 12 VM.

tperard: Yes, this bug is already reported:
[XCPNG-2775 \[xen,xapi\] Endianness mismatch on SMBIOS UUID with UEFI guest](https://project.vates.tech/vates-global/browse/XCPNG-2775/)
With some work upstream: [\[PATCH v2 0/3\] Update to SMBIOS 2.6](https://lore.kernel.org/xen-devel/cover.1756460430.git.teddy.astie@vates.tech/)

But it's all related to guest running under UEFI. (Well the upstream work would be a solution to fix the issue everywhere.)

With guest started with BIOS, Linux managed to get the UUID in the correct order, but Windows doesn't if I recall correctly.

The source of the system-uuid is xapi, that's the only one you can fully trust to be correct. Next, if I remember correctly, `/sys/hypervisor/uuid` should be fine and the same as the one is xapi. And the value that `dmidecode` uses can be wrong, depending on how the guest was started.

rzr: > Yes, this bug is already reported: [XCPNG-2775 [xen,xapi] Endianness mismatch on SMBIOS UUID with UEFI guest](https://project.vates.tech/vates-global/browse/XCPNG-2775/) With some work upstream: [[PATCH v2 0/3] Update to SMBIOS 2.6](https://lore.kernel.org/xen-devel/cover.1756460430.git.teddy.astie@vates.tech/)

Excellent, I would be happy to test it because what was puzzling me is that the bug is not deterministic (some systems have correct uuid others not), anyway using the proposed fallback seems a good alternative to replace the dmidecode call, maybe the xapi script could be used for other tasks, but if we could avoid it at this stage it will make maintenance ship easier

TSnake41: When XCP-ng is installed, XAPI generates a random UUID that will be used as (here nested) Dom0 UUID, this is unrelated to SMBIOS UUID hence unrelated to "nested XCP-ng as a guest" UUID.
This UUID is then sent to L1 Xen, and `/sys/hypervisor/uuid` will return this one which very certainly will not match "nested XCP-ng as a guest" UUID.

You can get "nested XCP-ng as a guest" UUID (through SMBIOS, you may have more luck with "serial number" which is stringified UUID rather than the SMBIOS UUID field) but it will be a different one to the "nested XCP-ng as a host in pool".

lib/tools/tasks/update.py
glehmann: I mean, if it's a problem, maybe we shouldn't only discard the value but instead raise an exception?

lib/tools/tasks/common.py
stormi: One thing this whole discussion tells me is that this PR doesn't accurately describe the problem it wants to solve. I'm putting it as draft until it does.

lib/tools/tasks/common.py
```diff
@@ -0,0 +1,37 @@
...
+import socket
+def main():
+    session = XenAPI.xapi_local()
+    try:
+        session.xenapi.login_with_password('root', '')
+        uuid = ''
+        for h in session.xenapi.host.get_all():
+            hostname = session.xenapi.host.get_hostname(h)
+            if hostname == socket.gethostname():
+                uuid = session.xenapi.host.get_bios_strings(h)['system-serial-number']
```
stormi: 1. Do we really need to iterate over all hosts in order to get the one that we are interested in? I would be surprised if XAPI offered no way to query a host directly based on its object UUID.

2. I think getting the host's system serial number should be a method of the Host class.

lib/tools/tasks/common.py
rzr: > `cat /sys/hypervisor/uuid` have a different result! :(

~~Because this one is what we want/need, as suggested above just try to replace dmidecode call with this and it should be enough ~~ 

If It works, then this whole xapi script PR can be closed as there is a simpler solution to address the identification issue (this is the fix to the bug introduced by usage of dmidecode in earlier merged PR).

Update: I tested it looks it's a different id (as @TSnake41  pointed out) the correct/desired one is 30424f38-7efd-f527-0d20-71dba677c549)

```
[10:21 pcl-49 ~]# dmidecode -s system-uuid 
384F4230-FD7E-27F5-0D20-71DBA677C549

[10:23 pcl-49 ~]# cat /sys/hypervisor/uuid 
d81b7bfd-b1a8-46da-8757-f0999520b1d7
```


The fallback option I can propose is to check if  system-serial number is a uuid and use it in place and if not just return system-uuid.

TSnake41: > > `cat /sys/hypervisor/uuid` have a different result! :(
> 
> ~~Because this one is what we want/need, as suggested above just try to replace dmidecode call with this and it should be enough ~~
> 
> If It works, then this whole xapi script PR can be closed as there is a simpler solution to address the identification issue (this is the fix to the bug introduced by usage of dmidecode in earlier merged PR).
> 
> Update: I tested it looks it's a different id (as @TSnake41 pointed out) the correct/desired one is 30424f38-7efd-f527-0d20-71dba677c549)
> 
> ```
> [10:21 pcl-49 ~]# dmidecode -s system-uuid 
> 384F4230-FD7E-27F5-0D20-71DBA677C549
> 
> [10:23 pcl-49 ~]# cat /sys/hypervisor/uuid 
> d81b7bfd-b1a8-46da-8757-f0999520b1d7
> ```

Does `dmidecode -s system-serial-number` works ?
It's a bit of a implementation detail abuse, but it workaround the endianness issue.

rzr: IIRC dmidecode -s system-serial-number works but should be only used on nested, as suggested above:

"The fallback option I can propose is to check if system-serial number is a uuid and use it in place and if not just return system-uuid."

Pseudo code:
```
# TODO: temp workaround for XCPNG-2775 (ideally prefer an upstream bug id if any exists)
if sudo dmidecode -s bios-vendor == Xen then
   return dmidecode -s system-serial-number
else 
   return dmidecode -s system-serial-uuid
```


Alternatively you can work directly with sys files:

```
cat /sys/devices/virtual/dmi/id/bios_vendor 
Xen

dmidecode -s system-serial-number 
30424f38-7efd-f527-0d20-71dba677c549
cat /sys/devices/virtual/dmi/id/product_serial  
30424f38-7efd-f527-0d20-71dba677c549

dmidecode -s system-uuid 
384F4230-FD7E-27F5-0D20-71DBA677C549
cat /sys/devices/virtual/dmi/id/product_uuid 
384f4230-fd7e-27f5-0d20-71dba677c549
```


Then mention the PR is related ticket to make sure we revert this change once the real fix is landing in our supported product(s).

TSnake41: If you're not managing a nested XCP-ng (e.g bare-metal), I don't think the SMBIOS UUID has any meaning. XAPI doesn't use it.

lib/host.py
```diff
@@ -459,27 +454,48 @@ def pool_has_vm(self, vm_uuid: str, vm_type: str = 'vm') -> bool:
...
             return self.xe('vm-list', {'uuid': vm_uuid}, minimal=True) == vm_uuid
 
-    def get_system_uuid(self) -> str:
-        """Return system uuid of current host.
+    def get_system_serial_number(self) -> str:
+        """Get system serial number.
+
+        Connect to host and execute a raw python script to
+        get system serial number. This function is quite
+        equivalent to the command: `dmidecode -s system-uuid`.
```
stormi, line 462: Anyone reading this in the future risks wondering why we're doing a complicated script if it's equivalent to `dmidecode -s system-uuid`. Maybe you mean that it is close to it, but more predictable?

lib/tools/tasks/common.py
rzr: > If you're not managing a nested XCP-ng (e.g bare-metal), I don't think the SMBIOS UUID has any meaning. XAPI doesn't use it.

ok but it costs nothing to implement for a high level API

PR #568, opened 2026-06-04
Add ipdb with IPython debugger support for pytest

glehmann:
Configure pytest to use IPython's TerminalPdb as the default post-mortem debugger for an improved debugging experience:
- display a few lines around the current line
- syntax coloring
- completion

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,6 +16,7 @@ dependencies = [
     "pytest>=9.1.1",
     "pytest-dependency",
     "requests",
+    "ipdb",
 ]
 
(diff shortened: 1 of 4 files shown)


stormi (changes requested):
Marking as "changes requested" to ensure it is discussed first.

dinhngtu (changes requested):
There's still a fixme in here? It also needs to be tested in Jenkins, but as long as it works in there, it's fine for me.

rzr (approved):

dinhngtu (approved):

PR #570, opened 2026-06-04
XCP-ng 9 fixes

glehmann:
Adapt to XCP-ng 9 specificities

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -426,7 +426,7 @@ def host_less_than_8_3(host: Host) -> None:
     if not host.xcp_version < version.parse(version_str):
         pytest.skip(f"This test requires an XCP-ng < {version_str} host")
 
-@pytest.fixture(scope='session')
+@pytest.fixture(scope='function')
 def host_with_hsts(host: Host) -> Generator[Host, None, None]:
     host.enable_hsts_header()
     yield host
--- a/lib/common.py
+++ b/lib/common.py
@@ -192,17 +192,17 @@ def callable_marker(value: T | Callable[..., T], request: pytest.FixtureRequest)
     else:
         return value
 
-def wait_for(fn: Callable[[], object], msg: str | None = None, timeout_secs: int = 2 * 60, retry_delay_secs: int = 2,
-             invert: bool = False) -> None:
+def wait_for(fn: Callable[[], T], msg: str | None = None, timeout_secs: int = 2 * 60, retry_delay_secs: int = 2,
+             invert: bool = False) -> T:
     if msg is not None:
         logging.info(msg)
     start_time = time.perf_counter()
     while True:
         ret = fn()
         if not invert and ret:
-            return
+            return ret
         if invert and not ret:
-            return
+            return ret
         if time.perf_counter() - start_time >= timeout_secs:
             expected = 'True' if not invert else 'False'
             raise TimeoutError(
--- a/lib/host.py
+++ b/lib/host.py
@@ -55,6 +55,7 @@ def host_data(hostname_or_ip: str) -> dict[str, str]:
 class Host:
     xe_prefix = "host"
     pool: "Pool"
+    package_manager: str
 
     # Data extraction is automatic, no conversion from str is done.
     @dataclass
@@ -89,6 +90,13 @@ class Host:
         self._dom0: VM | None = None
 
         self.rescan_block_devices_info()
+
+        if self.file_exists('/usr/bin/yum'):
+            self.package_manager = 'yum'
+        elif self.file_exists('/usr/bin/dnf'):
+            self.package_manager = 'dnf'
+        else:
+            raise Exception("No yum or dnf on the host")
 
     def __str__(self) -> str:
         return self.hostname_or_ip
@@ -509,7 +517,7 @@ class Host:
             yum clean metadata -q
         """
         logging.info(f"[{self}] Removing cache metadata...")
-        return self.ssh("yum clean metadata -q")
+        return self.ssh(f'{self.package_manager} clean metadata -q')
 
     def yum_update(self, enablerepos: list[str] = []) -> str:
         """Updates packages on target.
@@ -522,7 +530,7 @@ class Host:
 
         :param enablerepos: Enable one or more repositories (default: []).
         """
-        base_command = "yum update -y"
+        base_command = f'{self.package_manager} update -y'
 
         logging.info(f"[{self}] Updating packages...")
         if enablerepos:
@@ -601,7 +609,7 @@ class Host:
     def has_updates(self) -> bool:
         try:
             # yum check-update returns 100 if there are updates, 1 if there's an error, 0 if no updates
-            self.ssh('yum check-update')
+            self.ssh(f'{self.package_manager} check-update')
             # returned 0, else there would have been a SSHCommandFailed
             return False
         except commands.SSHCommandFailed as e:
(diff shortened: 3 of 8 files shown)


lib/host.py
```diff
@@ -90,6 +91,13 @@ def __init__(self, pool: Pool, hostname_or_ip: str):
 
         self.rescan_block_devices_info()
 
+        if self.file_exists('/usr/bin/yum'):
```
rzr, line 94: woultnt it make sense to prefer dnf over yum ?

lib/host.py
```diff
@@ -625,41 +633,52 @@ def get_last_yum_history_tid(self) -> int:
         [...]
         """
         try:
-            history_str = self.ssh('yum history list --noplugins')
+            history_str = self.ssh(f'{self.package_manager} history list --noplugins')
         except commands.SSHCommandFailed:
```
stormi, line 637: It's not directly related to your PR, and is probably code I wrote initially, so I just make it a suggestion: we should probably only tolerate failures that we can verify to be actually caused by an empty history, rather than any command failure that could happen for whatever reason. And let the exception through in the other cases.

lib/host.py
```diff
@@ -625,41 +633,52 @@ def get_last_yum_history_tid(self) -> int:
...
+                    break
+            if line_index is None:
+                raise Exception('Unable to get yum transactions')
+            return history[line_index + 1:]
+
+        history = split_history(history_str)
+        if not history:
+            # we need a transaction to already exist. Install a small package with no deps, and remove it immediately.
+            logging.info(f"[{self}] Install and remove a small package to workaround empty yum history.")
+            small_pkg = 'gpm-libs' if self.xcp_version.major == 8 else 'bubblewrap'
```
stormi, line 661: I suggest `dummypkg`, a package that must be present in our repos because other tests rely on it. If not present in XCP-ng 9, we should add it there too.

rzr (approved):

lib/host.py
gduperrey: I’d say that for now, yum is the default, since we’re on version 8.3 and are adding dnf alongside it for 9.0. And as we don't have dnf on the 8.3 (and before).

If we wanted dnf to be the default over yum for 9.0, we would need to base that on version detection first, before the presence of the file.

conftest.py
```diff
@@ -426,7 +426,7 @@ def host_less_than_8_3(host: Host) -> None:
     if not host.xcp_version < version.parse(version_str):
         pytest.skip(f"This test requires an XCP-ng < {version_str} host")
 
-@pytest.fixture(scope='session')
+@pytest.fixture(scope='function')
```
stormi, line 429: Most host_xxx fixtures in this global conftest.py are session-scoped. I think that to change the scope we should change the fixture name too to avoid wrong expectations. And/or bring the fixture closer to where it's used.

lib/host.py
glehmann: There is no `yum` command on XCP-ng 9, so it doesn't matter much in practice :)

gduperrey (approved):

dinhngtu (approved):

lib/common.py
```diff
@@ -192,17 +192,17 @@ def callable_marker(value: T | Callable[..., T], request: pytest.FixtureRequest)
...
+def wait_for(fn: Callable[[], T], msg: str | None = None, timeout_secs: int = 2 * 60, retry_delay_secs: int = 2,
+             invert: bool = False) -> T:
     if msg is not None:
         logging.info(msg)
     start_time = time.perf_counter()
     while True:
         ret = fn()
         if not invert and ret:
-            return
+            return ret
```
stormi, line 203: nitpick: it took me time to understand this part of the commit message: "Enhance wait_for to return the callback value ", because I thought it meant you returned the callback itself, as a value. 

Due to the use of word "callback", I thought there was a function we'd call back once returned, but you just mean you return the value produced by the function executed by wait_for.

Suggestion: "Enhance wait_for to return the polled function's result"

lib/vm.py
```diff
@@ -784,6 +784,9 @@ def is_in_uefi_shell(self) -> bool:
         tmp_file = res_host.ssh('mktemp')
         session = f"detached-cat-{self.uuid}"
         ret = False
+        # screen install is broken on RHEL, see https://bugzilla.redhat.com/show_bug.cgi?id=2385964
```
stormi, line 787: Okay, another "I got confused by the wording" item: I thought there was something about display issues 😂.

But you're talking about `screen`, the tool.

Suggestion: 
```
The `screen` tool may be broken on RHEL, see https://bugzilla.redhat.com/show_bug.cgi?id=2385964
```

lib/common.py
stormi, line 1: I forgot to ask: was the PR tested on XCP-ng 8 too? Probably so but it would be best to mention it in the PR description so that we don't ask :)

lib/host.py
glehmann: It's not in xcp-ng 9 yet
We could use the same package on xcp-ng 9 and xcp-ng 8.3 :+1:

lib/common.py
glehmann: I tested both for the tests I modified, but I haven't run a complete run on XCP-ng 8.3 yet.

glehmann:
I integrated the requested changes and updated on master

gduperrey (approved):

lib/host.py
ydirson: There is actually a `yum` compatibility wrapper (in package `yum`). It's surely safer to look for `dnf` first, just in case.

lib/host.py
```diff
@@ -633,41 +641,54 @@ def get_last_yum_history_tid(self) -> int:
...
+            # we need a transaction to already exist. Install a small package with no deps, and remove it immediately.
+            logging.info(f"[{self}] Install and remove a small package to workaround empty yum history.")
+            self.yum_install(['dummypkg'])
+            self.yum_remove(['dummypkg'])
+            history = split_history(self.ssh(f'{self.package_manager} history list --noplugins'))
 
         try:
-            return int(history[line_index + 1].split()[0])
+            return int(history[0].split()[0])
         except ValueError:
```
ydirson, line 678: (nitpick?) This change could have been easier to read by introducing `split_history()` in a preliminary commit (which involves some reindentation), and then the effective difference specific to v9 would have stood out more clearly in the diff.

lib/host.py
```diff
@@ -633,41 +641,54 @@ def get_last_yum_history_tid(self) -> int:
...
-        line_index = None
-        for i in range(len(history)):
-            if history[i].startswith('--------'):
-                line_index = i
-                break
+            history_str = self.ssh(f'{self.package_manager} history list --noplugins')
+        except commands.SSHCommandFailed as e:
+            if 'Error: Failed history list' not in e.stdout:
+                # this is not the expected error for an empty history list, re-raise the exception
+                raise e
```
ydirson: `raise` without argument would be more correct I think

lib/host.py
```diff
@@ -633,41 +641,54 @@ def get_last_yum_history_tid(self) -> int:
...
+                raise Exception('Unable to get yum transactions')
+            return history[line_index + 1:]
+
+        history = split_history(history_str)
+        if not history:
+            # we need a transaction to already exist. Install a small package with no deps, and remove it immediately.
+            logging.info(f"[{self}] Install and remove a small package to workaround empty yum history.")
+            self.yum_install(['dummypkg'])
+            self.yum_remove(['dummypkg'])
+            history = split_history(self.ssh(f'{self.package_manager} history list --noplugins'))
```
ydirson, line 674: A few words in the commit message about that particular behavior change and how it is restructured would be nice (yes the can read the code and actually understand, but a "reading guide" would actually allow the reader to spend less time)

lib/host.py
```diff
@@ -700,9 +721,12 @@ def yum_restore_saved_state(self) -> None:
...
 
-        self.ssh(
-            f'yum history rollback --enablerepo=xcp-ng-base,xcp-ng-testing,xcp-ng-updates {self.saved_rollback_id} -y'
-        )
+        repositories = ['xcp-ng-base']
+        if self.xcp_version.major == 8:
+            # TODO: activate those repositories in xcp-ng 9. For now they are not available.
+            repositories += ['xcp-ng-testing', 'xcp-ng-updates']
+        self.ssh(f'{self.package_manager} history rollback'
+                 f' --enablerepo={",".join(repositories)} {self.saved_rollback_id} -y')
```
ydirson, line 729: This also is not implied by "use the right package manager".  Sure this is v9 support, and relates to repositories, so likely has a place grouped as such, but would be nice in the commit message

lib/host.py
```diff
@@ -633,41 +641,54 @@ def get_last_yum_history_tid(self) -> int:
...
         """
         try:
-            history_str = self.ssh('yum history list --noplugins')
-        except commands.SSHCommandFailed:
-            # yum history list fails if the list is empty, and it's also not possible to rollback
-            # to before the first transaction, so "0" would not be appropriate as last transaction.
-            # To workaround this, create transactions: install and remove a small package.
-            logging.info(f"[{self}] Install and remove a small package to workaround empty yum history.")
-            self.yum_install(['gpm-libs'])
-            self.yum_remove(['gpm-libs'])
```
ydirson, line 643: switching to `dummypkg` because this package is not available in v9?

tests/misc/test_export.py
```diff
@@ -19,15 +19,15 @@ def export_test(host: Host, vm: VM, filepath: str, compress: Literal['none', 'gz
...
 
     if compress == 'none':
-        check_file_type('application/x-tar')
+        check_file_type(['application/x-tar'])
     elif compress == 'gzip':
-        check_file_type('application/x-gzip')
+        check_file_type(['application/x-gzip', 'application/gzip'])
     elif compress == 'zstd':
-        check_file_type('application/octet-stream')
+        check_file_type(['application/octet-stream', 'application/zstd'])
```
ydirson: I disagree this makes things "more robust".  `application/octet-stream` is "I did not recognize this file", which should be an error when the more specific type is known.  I like my "test_export: support modern distro knowing about zstd and better gzip" patch in `ydi/9` better 😄

lib/installer.py
```diff
@@ -96,7 +96,7 @@ def inner():
         if failed:
             raise InstallationFailed(failed)
         return cmd()
-    return wait_for(inner, msg, timeout_secs=timeout_secs)
+    wait_for(inner, msg, timeout_secs=timeout_secs)
```
ydirson, line 99: It feels awkward to have `wait_for` gain a real return value, and just not propagate it here.
Any reason not to change this function to behave similarly to ` (possibly by just changing the return type annotation)?

lib/common.py
```diff
@@ -192,17 +192,17 @@ def callable_marker(value: T | Callable[..., T], request: pytest.FixtureRequest)
     else:
         return value
 
-def wait_for(fn: Callable[[], object], msg: str | None = None, timeout_secs: int = 2 * 60, retry_delay_secs: int = 2,
-             invert: bool = False) -> None:
+def wait_for(fn: Callable[[], T], msg: str | None = None, timeout_secs: int = 2 * 60, retry_delay_secs: int = 2,
+             invert: bool = False) -> T:
```
ydirson, line 196: This use of the return value should be documented, a function docstring would seem adequate.

Especially the behavior with `invert=True` may need some notice, that only False-like values will be returned in this case (so not very useful).

lib/vm.py
```diff
@@ -784,6 +784,9 @@ def is_in_uefi_shell(self) -> bool:
         tmp_file = res_host.ssh('mktemp')
         session = f"detached-cat-{self.uuid}"
         ret = False
+        # The screen package installation is broken on RHEL, see https://bugzilla.redhat.com/show_bug.cgi?id=2385964
```
ydirson: s/RHEL/el10/ ?

ydirson (changes requested):

lib/host.py
glehmann: Are you sure about that? I added it yesterday :)

lib/host.py
glehmann: Done. This feels very much not useful to me, but if that's useful to you…

lib/host.py
glehmann: Agreed, because it would avoid changing the stack trace 👍

lib/installer.py
glehmann: Because the function to call returns a `bool` (`Callable[[], bool]`), so, given the behavior of `wait_for`, the only possible returned value is `Literal[True]`, which is not very useful.
I'm fine with both versions, though

tests/misc/test_export.py
glehmann: I replace this commit with yours

stormi (approved):
Approving based on my previous review + comments addressed by @glehmann. I rely on other reviewers for the rest.

PR #572, opened 2026-06-05
xo_cli changes

semarie:
the PR is instroducing some `xo_cli` changes:
- use `TOOLS` variable in `data.py` to get the path of `xo-cli` command-line : it permits to use custom path. the default is to use `xo_cli` as currently
- add `logging.debug()` when `xo-cli` is invoked in order to trace it

### DIFF
--- a/lib/xo.py
+++ b/lib/xo.py
@@ -1,6 +1,7 @@
 import json
 import subprocess
 
+from data import TOOLS
 from lib.typing import JSONType
 
 from typing import Dict, Literal, overload
@@ -28,7 +29,7 @@ def xo_cli(action: str, args: Dict[str, str] = {}, *, check: bool = True, simple
 def xo_cli(
     action: str, args: dict[str, str] = {}, check: bool = True, simple_output: bool = True, use_json: bool = False
 ) -> subprocess.CompletedProcess[bytes] | JSONType | str:
-    run_array = ['xo-cli', action]
+    run_array = [TOOLS.get('xo-cli', 'xo-cli'), action]
     if use_json:
         run_array += ['--json']
     run_array += ["%s=%s" % (key, value) for key, value in args.items()]
--- a/lib/xo.py
+++ b/lib/xo.py
@@ -1,51 +1,32 @@
 import json
-import subprocess
 
 from data import TOOLS
+from lib.commands import local_cmd
 from lib.typing import JSONType
 
-from typing import Dict, Literal, overload
-
-# TODO: either:
-#   * replace simple_output and use_json by a single output type
-#   * make sure that simple_output=False and use_json=True are not being used together
+from typing import Literal, overload
 
 @overload
-def xo_cli(action: str, args: Dict[str, str] = {}, *, check: bool = True, simple_output: Literal[True] = True,
-           use_json: Literal[False] = False) -> str:
+def xo_cli(action: str, args: dict[str, str] = {}, *, check: bool = True, use_json: Literal[False] = False) -> str:
     ...
 @overload
-def xo_cli(action: str, args: Dict[str, str] = {}, *, check: bool = True, simple_output: Literal[True] = True,
-           use_json: Literal[True]) -> JSONType:
+def xo_cli(action: str, args: dict[str, str] = {}, *, check: bool = True, use_json: Literal[True]) -> JSONType:
     ...
-@overload
-def xo_cli(action: str, args: Dict[str, str] = {}, *, check: bool = True, simple_output: Literal[False],
-           use_json: bool = False) -> subprocess.CompletedProcess[bytes]:
-    ...
-@overload
-def xo_cli(action: str, args: Dict[str, str] = {}, *, check: bool = True, simple_output: bool = True,
-           use_json: bool = False) -> subprocess.CompletedProcess[bytes] | JSONType | str:
-    ...
-def xo_cli(
-    action: str, args: dict[str, str] = {}, check: bool = True, simple_output: bool = True, use_json: bool = False
-) -> subprocess.CompletedProcess[bytes] | JSONType | str:
-    run_array = [TOOLS.get('xo-cli', 'xo-cli'), action]
+
+def xo_cli(action: str, args: dict[str, str] = {}, *, check: bool = True, use_json: bool = False) -> JSONType | str:
+    cmd = [TOOLS.get('xo-cli', 'xo-cli'), action]
     if use_json:
-        run_array += ['--json']
-    run_array += ["%s=%s" % (key, value) for key, value in args.items()]
-    res = subprocess.run(
-        run_array,
-        stdout=subprocess.PIPE,
-        stderr=subprocess.STDOUT,
-        check=check
-    )
-    if simple_output:
-        output = res.stdout.decode().strip()
-        if use_json:
-            return json.loads(output)
-        return output
-    return res
+        cmd += ['--json']
+    cmd += ["%s=%s" % (key, value) for key, value in args.items()]
+
+    res = local_cmd(cmd, check=check)
+
+    if use_json:
+        return json.loads(res.stdout)
+
+    return res.stdout
 
 def xo_object_exists(uuid: str) -> bool:
-    lst = json.loads(xo_cli('--list-objects', {'uuid': uuid}))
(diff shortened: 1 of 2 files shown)


lib/xo.py
```diff
@@ -33,6 +34,8 @@ def xo_cli(
     if use_json:
         run_array += ['--json']
     run_array += ["%s=%s" % (key, value) for key, value in args.items()]
+
+    logging.debug(f"xo_cli: {' '.join(run_array)}")
```
glehmann: Maybe use `local_cmd()` in place of `subprocessrun()`? It also does the logging.

glehmann (approved):

lib/xo.py
```diff
@@ -1,50 +1,30 @@
...
-        check=check
-    )
-    if simple_output:
-        output = res.stdout.decode().strip()
-        if use_json:
-            return json.loads(output)
-        return output
-    return res
+        cmd += ['--json']
+    cmd += ["%s=%s" % (key, value) for key, value in args.items()]
```
olivierh-pro, line 20: What if *value* is splitted by spaces here? Resulting command line would be wrong, isn't it?

```python
>>> l = ["%s=%s" % (key, value) for key, value in d.items()]
>>> l
['foo=foo foo', 'bar=bar bar']
```
With my example above, final command would be `"xo-cli foo=foo foo bar=bar bar"` right?

lib/xo.py
ydirson: Does it behave correctly in case of error when `check=False`?  It looks like it would return stdout and not the `LocalCommandResult`.

lib/xo.py
```diff
@@ -1,48 +1,30 @@
...
     ...
-@overload
-def xo_cli(action: str, args: Dict[str, str] = {}, *, check: bool = True, simple_output: bool = True,
-           use_json: bool = False) -> LocalCommandResult[bytes] | JSONType | str:
-    ...
-def xo_cli(
-    action: str, args: dict[str, str] = {}, check: bool = True, simple_output: bool = True, use_json: bool = False
-) -> LocalCommandResult[bytes] | JSONType | str:
+
+def xo_cli(action: str, args: dict[str, str] = {}, *, check: bool = True, use_json: bool = False) -> Any:
```
ydirson: This intended simplification is not explained in the commit message, what is the goal, and what is the commit doing?
- it notably replaces careful type-hinting of the return value with `Any`, preventing the typecheckers to detect errors, that looks like a misguided idea

ydirson (changes requested):

lib/xo.py
semarie: First, I didn't rewrite `xo_cli` but keep it "as it", so it isn't a behavior change.

Next, I don't see any problem here: you are passing 2 arguments (containing spaces) to underlined command. They will not break in 4 arguments.

```sh
$ cat /tmp/test.sh
#!/bin/sh
printf '[%s]\n' "$@"
$ uv run python3
>>> from lib.commands import local_cmd
>>> print(local_cmd(['/tmp/test.sh', 'foo=foo foo', 'bar=bar bar']).stdout)
[foo=foo foo]
[bar=bar bar]
```

lib/xo.py
semarie: 1. The simplification came as second step. I first just added logging but @glehmann noted that [local_cmd could be just used](https://github.com/xcp-ng/xcp-ng-tests/pull/572#discussion_r3364180935).

2. Regarding the typage, I agree that I switched from `JSONType` to `Any`. The reason are:

- follow what `json.loads()` actually returns
- it simplifies **a lot** the usage of `xo_cli` as depending the arguments the output type isn't the same and typecheckers doesn't see it:
  - `xo_cli('list-objects', use_json=True)` : returns a list of dict of ...
  - `xo_cli('list-commands', {})` : returns a dict of ...
  - `xo_cli('plugin.load', {'id': 'sdn-controller'})` : returns a bool

Here the implementation of `xo_object_exists` with `xo_cli` returning `JSONType`:

```py
def xo_object_exists(uuid: str) -> bool:
    lst = xo_cli('list-objects', {'uuid': uuid}, use_json=True)
    assert isinstance(lst, list)
    return len(lst) > 0
```

Without the `assert`, mypy will complains that `Argument 1 to "len" has incompatible type "int | float | str | list[JSONType] | dict[str, JSONType] | None"; expected "Sized"  [arg-type]`.

lib/xo.py
semarie: you are commenting outdated code. it isn't on the latest changes.

lib/xo.py
olivierh-pro: For a reason I ignore, `subprocess` escapes correctly the space:

```python
>>> subprocess.run(args=["xo-cli", "list-objects", "foo=foo foo"])
[
]
CompletedProcess(args=['xo-cli', 'list-objects', 'foo=foo foo'], returncode=0)
```

So, it's ok ;)

lib/xo.py
semarie: I extended the commit message regarding `xo_cli` simplification.
if you prefer the type to stick to `JSONType` instead of `Any`, I will update it (and callers).

ydirson: When there are several structural changes like this, creating separate commits for each of them allows to better understand each of them.

>     2. Regarding the typage, I agree that I switched from `JSONType` to `Any`.

Not only that, `Any` here, just to handle one case where the type is not known with precision, gets to replace other cases `LocalCommandResult[bytes] | str` too.  Those may become moot from other changes in this commit, but it's hard to see since everything is in the same commit.  And if they don't, then the type-checker is left with less information to deal with the case when JSON is not directly returned.

Note, the original annotation looks quite unoptimal, I'd think the JSON types could only be declared as returned when `use_json` is `Literal[True]`, which would isolate to a limited number of overrides the issue you're trying to solve.

BTW I'm not a huge fan of `JSONType` either, it really looks like an nearly-`Any` alias hiding behind an `import`. I'm not objecting to getting rid of it, but this should be done in a more chirurgical way.


> Here the implementation of `xo_object_exists` with `xo_cli` returning `JSONType`:
> 
> ```python
> def xo_object_exists(uuid: str) -> bool:
>     lst = xo_cli('list-objects', {'uuid': uuid}, use_json=True)
>     assert isinstance(lst, list)
>     return len(lst) > 0
> ```
> 
> Without the `assert`, mypy will complains that `Argument 1 to "len" has incompatible type "int | float | str | list[JSONType] | dict[str, JSONType] | None"; expected "Sized" [arg-type]`.

The use of `assert` to sanitize the input received by `xo-cli` rather looks like good practice, at the same time documenting the assumption of the code using that output, and checking early that they actually stand.  It should needed also when using `Any`, or the type checkers would not be able to do any check.

If we want anything better, I'd rather see us using the XO API, it should be possible to generate python binding like what is done with XenAPI, and we should also be able to extract typing information from there. But this idea looks a bit out of scope for this PR :wink:

lib/xo.py
ydirson: I'm commenting on lines that are added by the second commit.  Getting this rewritten by a 3rd commit does not make it "outdated code": it would still be merged into our history, and `git bisect` will want to use it some day.  Each commit should result in valid code.

This seems to relate to changes not being self-contained and checked to be so.  I agree it's not necessarily easy to reach when dealing with several changes that touch the same lines, but [rebase-rewind](https://lore.kernel.org/git/139173043.431119331.1615653441685.JavaMail.root@zimbra39-e7/) can help.

lib/xo.py
```diff
@@ -1,50 +1,32 @@
...
-from lib.typing import JSONType
-
-from typing import Dict, Literal, overload
+from typing_extensions import Literal, overload
 
-# TODO: either:
-#   * replace simple_output and use_json by a single output type
-#   * make sure that simple_output=False and use_json=True are not being used together
+from data import TOOLS
+from lib.commands import local_cmd
```
glehmann, line 4: Is there a reason to the types import from `typing_extensions`?

semarie: no specific reason. my editor (zed) did it automatically. fixed.

olivierh-pro (approved):

glehmann (approved):

ydirson (approved):
lgtm

Though the last commit really does not mention that it is also changing `Dict` to `dict` and `run_array` to `cmd`.  Both are valuable changes, but making such cosmetic changes in a separate commit always helps with readability of the core changes.

PR #575, opened 2026-06-08
lib/host: add attribute in host to distinguish nested

olivierh-pro:
In this PR, I add a new Host attribute to determine whether a host is nested or not (it's physical if so).

### DIFF
--- a/lib/host.py
+++ b/lib/host.py
@@ -81,6 +81,7 @@ class Host:
         self.saved_packages_list: list[str] | None = None
         self.saved_rollback_id: int | None = None
         self.inventory = self._get_xensource_inventory()
+        self._bios_vendor: str | None = None
         self.uuid = self.inventory['INSTALLATION_UUID']
         self.xcp_version = version.parse(self.inventory['PRODUCT_VERSION'])
         self.xcp_version_short = f"{self.xcp_version.major}.{self.xcp_version.minor}"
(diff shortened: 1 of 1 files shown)


glehmann (approved):

rzr (approved):

lib/host.py
```diff
@@ -277,6 +278,18 @@ def _get_xensource_inventory(self) -> dict[str, str]:
             inventory[key] = raw_value.strip('\'')
         return inventory
 
+    def _get_bios_vendor(self) -> str:
```
rzr, line 281: Any reason to have the _ prefix ? a private API ? I would make it public ?

lib/host.py
```diff
@@ -81,6 +81,7 @@ def __init__(self, pool: Pool, hostname_or_ip: str):
         self.saved_packages_list: list[str] | None = None
         self.saved_rollback_id: int | None = None
         self.inventory = self._get_xensource_inventory()
+        self.nested = 'Xen' in self._get_bios_vendor()
```
rzr: Is it a good practice to have this in constructor ? I would set it to None and update it (once) when needed

lib/host.py
olivierh-pro: Yes, it is a python convention to make it private.

Currently, this bios information is not needed by anyone else (I guess). So I prefer to put it private.

Am I wrong? :)

rzr: up to you, I would make it public to avoid a later change when there is need for, since it should be sable as it is just a wrapper. but nevermind it's not a big deal.

rzr (approved):
Thanks for taking into account my after review suggestions

lib/host.py
```diff
@@ -277,6 +278,26 @@ def _get_xensource_inventory(self) -> dict[str, str]:
...
+
+            dmidecode -s bios-vendor
+
+        >>> my_host._get_bios_vendor()
+        Xen
+        """
+        return self.ssh('dmidecode -s bios-vendor')
+
+    @property
+    def nested(self):
```
```suggestion
    def is_nested(self) -> bool:
```
to be more consistent with how it's done in other places—`VM.is_windows` for example.

olivierh-pro: Done!

PR #576, opened 2026-06-08
tools/update: fix use of dmidecode in context of nested hosts

olivierh-pro:
This PR fixes an unexpected behaviour for command `dmidecode -s system-uuid` in nested hosts (Some bits are shifted): See comment [here](https://github.com/xcp-ng/xcp-ng-tests/pull/567#discussion_r3356172811).

Reminder: "system uuid" is used to target a nested host in its hosting pool. This work follows up previous work in snapshot creation (#468).

### DIFF
--- a/lib/host.py
+++ b/lib/host.py
@@ -481,26 +481,25 @@ class Host:
             return self.xe('vm-list', {'uuid': vm_uuid}, minimal=True) == vm_uuid
 
     def get_system_uuid(self) -> str:
-        """Return system uuid of current host.
+        """Get system uuid of current host.
 
-        Intended for driving current host from its "parent host" in a **nested context**.
+        In case host is nested, it uses `system-serial-number` instead of `system-uuid`.
 
-        .. note::
-            If the current host is nested, it means it is not a physical host. It is a VM living inside a real host.::
-
-                [PH: Physical Host] -> [VM: emulation of an XCP-ng host] -> [vm: a vm inside nested host]
-                                       |      current working host     |
-
-            So we need system-uuid of current working host (`VM`) which is
-            the uuid seen in physical host's (`PH`) scope.
+        If command result is empty or None, it raises an Error.
 
         Performs the following command::
 
-            dmidecode -s system-uuid
+            dmidecode -s [system-uuid|system-serial-number]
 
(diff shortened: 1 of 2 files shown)


glehmann (approved):

lib/host.py
```diff
@@ -480,26 +480,24 @@ def pool_has_vm(self, vm_uuid: str, vm_type: str = 'vm') -> bool:
...
-            dmidecode -s system-uuid
+            dmidecode -s [system-uuid|system-serial-number]
 
         ref: `dmidecode(8) <https://man.archlinux.org/man/dmidecode.8.en#s>__`
         """
-        return self.ssh("dmidecode -s system-uuid").lower().strip()
+        dmi_key = "system-serial-number" if self.is_nested else "system-uuid"
+        system_uuid = self.ssh(f"dmidecode -s {dmi_key}").lower().strip()
+        if not system_uuid:
+            raise ValueError(f"The system uuid '{system_uuid}' incorrect.")
```
```suggestion
            raise ValueError(f"The system uuid '{system_uuid}' is incorrect.")
```

lib/host.py
```diff
@@ -480,26 +480,24 @@ def pool_has_vm(self, vm_uuid: str, vm_type: str = 'vm') -> bool:
             return self.xe('vm-list', {'uuid': vm_uuid}, minimal=True) == vm_uuid
 
     def get_system_uuid(self) -> str:
-        """Return system uuid of current host.
+        """Get system uuid of current host.
 
-        Intended for driving current host from its "parent host" in a **nested context**.
+        In case host is nested, it uses `system-serial-number` instead of `system-uuid`.
```
rzr, line 486:
```suggestion
        In case host is nested, it uses `system-serial-number` instead of `system-uuid`.
        
        (TODO: This workaround is tracked in XCPNG-2775 ) 
```
(ideally prefer an upstream bug id if any exists)

rzr (approved):
Let me try to test it, before merge.

rzr (approved):
I tested, the snapshot is working and noticed that yum errors are ignored, I would raise an expection on failed ssh commands (it can be done in later PR), let's merge

lib/host.py
olivierh-pro: Done

PR #579, opened 2026-06-09
conftest: split vm-type fixtures into require and skip variants

d3athjest3r:
Opened new pull-request based on https://github.com/xcp-ng/xcp-ng-tests/pull/410
Ping: @stormi 

unix_vm, windows_vm, and uefi_vm now call pytest.fail() if the VM is the wrong type, matching the expectation that these fixtures enforce a requirement. The new skip_if_not_unix_vm, skip_if_not_windows_vm, and skip_if_not_uefi_vm fixtures call pytest.skip() instead, for use in multi-VM jobs where different VMs types are expected and non-matching ones should be silently skipped rather than failed. So, updated all multi_vms-annotated classes to use the skip variants. The markable_fixtures list and jobs.py marker expressions are updated accordingly.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -170,21 +170,26 @@ def pytest_collection_modifyitems(items: list[pytest.Item], config: pytest.Confi
     # Automatically mark tests based on fixtures they require.
     # Check pytest.ini or pytest --markers for marker descriptions.
 
-    markable_fixtures = [
-        'uefi_vm',
-        'unix_vm',
-        'windows_vm',
-        'hostA2',
-        'hostB1',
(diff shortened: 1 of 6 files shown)


casasnovas (changes requested):
The commit description is a bit misleading (or I'm missing something), in that the current `unix_vm`, `windows_vm` and `uefi_vm` actually skip the test and don't fail it - it's only after your change that they do fail the test.  It does seem to imply that tests that were previously skipped will now fail (since you haven't changed all to use the new `skip_if_not_*` variant, is this intentional?

PR #582, opened 2026-06-11
t/g/u/test_xen_guest_agent.py: Use yum on older distros or dnf if available

rzr:
On modern systems that uses dnf, there is a compatibility yum wrapper, some refactoring can be considered to support more package managers.

Observed issue was:

    tests/guest_tools/unix/test_xen_guest_agent.py::TestXenGuestAgent::test_agent_running_after_reboot[...]
    lib.commands.SSHCommandFailed: SSH command (dnf install -y xen-guest-agent) failed with return code 127: bash: dnf: command not found


dinhngtu (approved):

glehmann (approved):
I'm preparing a version with the package manager detection in the VM

d3athjest3r (approved):
lgtm.

PR #583, opened 2026-06-11
tools/update: stop script on update failure

olivierh-pro:
Rework multithread part of update script to be
able to stop when there is a host that cannot be
updated.

### DIFF
--- a/lib/tools/tasks/update.py
+++ b/lib/tools/tasks/update.py
@@ -4,7 +4,7 @@ This module is intended for performing update actions on existing remote targets
 """
 from __future__ import annotations
 
-from concurrent.futures import ThreadPoolExecutor
+from concurrent.futures import ThreadPoolExecutor, as_completed
 
 from lib.host import Host
 from lib.pool import NotAMasterHostError, Pool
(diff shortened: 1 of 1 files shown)


lib/tools/tasks/update.py
```diff
@@ -45,17 +45,38 @@ def update_pools(inventory: Inventory) -> None:
 
     # update master hosts
     with ThreadPoolExecutor() as executor:
-        for p in pools:
-            executor.submit(p.master.update, inventory_hosts[p.master.hostname_or_ip]["repositories"])
+        future_updated_masters = {executor.submit(
+            p.master.update, inventory_hosts[p.master.hostname_or_ip]["repositories"]): p.master for p in pools}
+        for future in as_completed(future_updated_masters):
+            updated_master = future_updated_masters[future]
```
rzr: nitpicking:


```suggestion
            host = updates[future]
```


`updated_master` name can be be misleading, since it can not be (successfully) updated

Same for dict `future_updated_masters`  name is not obvious and i would name  it same with in both functions

lib/tools/tasks/update.py
```diff
@@ -45,17 +45,38 @@ def update_pools(inventory: Inventory) -> None:
...
+            try:
+                future.result()
+            except Exception as exc:
+                logger.error(f"Updating pool has failed! The master {updated_master} cannot be updated.")
+                logger.info(
+                    "*** Updating pool will stop due to previous error. Waiting for other masters to finish. ***"
```
rzr:
```suggestion
                    "*** Due to previous error, the pool updating task will stop, Let finish running updates if any.***"
```

olivierh-pro: Done

lib/tools/tasks/update.py
olivierh-pro: I prefer not renaming my variables here

glehmann (approved):

rzr (approved):

PR #588, opened 2026-06-15
Fix test_checkpoint on Windows

dinhngtu:
The series of failures of start_background_process on Windows has shown that this function is particularly unreliable on this platform.

The symptom is the test timing out on `test -f /tmp/bg_process.pid`.

Replace with the Windows equivalents for better reliability.

### DIFF
--- a/lib/vm.py
+++ b/lib/vm.py
@@ -818,17 +818,19 @@ class VM(BaseVM):
             f"Write-Output (Start-Process -Wait -PassThru {program} -ArgumentList '{args}').ExitCode")
         return int(output)
 
-    def start_background_powershell(self, cmd: str) -> None:
+    def start_background_powershell(self, cmd: str) -> str:
         """
-        Run command under powershell in the background.
+        Run command under powershell in the background. Return the PID as string.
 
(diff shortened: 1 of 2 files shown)


lib/vm.py
```diff
@@ -818,17 +831,19 @@ def run_powershell_command(self, program: str, args: str) -> int:
             f"Write-Output (Start-Process -Wait -PassThru {program} -ArgumentList '{args}').ExitCode")
         return int(output)
 
-    def start_background_powershell(self, cmd: str) -> None:
+    def start_background_powershell(self, cmd: str) -> str:
         """
-        Run command under powershell in the background.
+        Run command under powershell in the background. Returns the PID as string.
```
rzr:
```suggestion
        Run command under powershell in the background. Return the PID as string.
```
Align to infinitive (Run)

rzr (approved):
This tests was tested on Windows and linux, it is desirable for next stable branch update.

lib/vm.py
dinhngtu: Done

PR #590, opened 2026-06-18
tests/guest_tools: Skip unsupported RPM distros in xen-guest-agent test

d3athjest3r:
Skip yum-only distros (e.g. CentOS 7). RPM packages are built against Fedora 37 and won't install on old RHEL-like distros. Also skip SUSE-based distros where no packages are published. A multi-VM job may include such VMs alongside supported ones, so skip rather than fail.

### DIFF
--- a/tests/guest_tools/unix/test_xen_guest_agent.py
+++ b/tests/guest_tools/unix/test_xen_guest_agent.py
@@ -27,6 +27,7 @@ def _vif_published_ips(host: Host, xs_prefix: str, vif_id: int, proto: str) -> l
     ]
 
 
+@pytest.mark.skip(reason="Test suite is currently unstable — skipped until fixed")
 @pytest.mark.multi_vms
 @pytest.mark.usefixtures("unix_vm")
 class TestXenGuestAgent:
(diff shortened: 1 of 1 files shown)


casasnovas (approved):

tests/guest_tools/unix/test_xen_guest_agent.py
```diff
@@ -50,6 +50,16 @@ def agent_install(self, running_vm: VM, xen_guest_agent_urls: dict[str, str]) ->
             vm.ssh('apt-get remove -y xe-guest-utilities')
 
         if pkg_mgr == PackageManagerEnum.RPM:
+            # Skip these tests for certain unsupported platforms:
+            # - RPM packages are built against Fedora 37 and won't install on
```
rzr, line 55: they won't install using dnf but does it also mean the binaries won't run if installed with yum?

d3athjest3r: I assume not. They probably expect a different libc, and also other libraries. Fedora 37 is just sooo far apart from a Cent OS 7.

rzr (commented):
change is improving a bit the current situation, but it wont be enough to make it run on ci (using dnf), is the dependency on gitlab credential, would it be possible to skip it for now, until it's fully runnable everywhere ?

tests/guest_tools/unix/test_xen_guest_agent.py
rzr: I was thinking more about later distros (closer to f37) still using yum

tests/guest_tools/unix/test_xen_guest_agent.py
```diff
@@ -27,6 +27,7 @@ def _vif_published_ips(host: Host, xs_prefix: str, vif_id: int, proto: str) -> l
     ]
 
 
+@pytest.mark.skip(reason="Test suite is currently unstable — skipped until fixed")
```
rzr, line 30: Thank you

rzr (approved):

tests/guest_tools/unix/test_xen_guest_agent.py
rzr: I have tested this branch in CI it's now passing, Let's merge

PR #591, opened 2026-06-18
Detect yum and dnf in the VMs

glehmann:
And act accordingly.

### DIFF
--- a/lib/common.py
+++ b/lib/common.py
@@ -78,9 +78,11 @@ Defer: TypeAlias = Callable[[Callable[[], object]], None]
 
 class PackageManagerEnum(Enum):
     UNKNOWN = 1
-    RPM = 2
+    YUM = 2
     APT_GET = 3
     APK = 4
+    DNF = 5
+    ZYPPER = 6
 
 # Common VM images used in tests
 def vm_image(vm_key: str) -> str:
--- a/lib/vm.py
+++ b/lib/vm.py
@@ -511,12 +511,16 @@ class VM(BaseVM):
 
     def detect_package_manager(self) -> PackageManagerEnum:
         """ Heuristic to determine the package manager on a unix distro. """
-        if self.file_exists('/usr/bin/rpm') or self.file_exists('/bin/rpm'):
-            return PackageManagerEnum.RPM
+        if self.file_exists('/usr/bin/dnf'):
+            return PackageManagerEnum.DNF
+        if self.file_exists('/usr/bin/yum'):
+            return PackageManagerEnum.YUM
         if self.file_exists('/usr/bin/apt-get'):
             return PackageManagerEnum.APT_GET
         if self.file_exists('/sbin/apk'):
             return PackageManagerEnum.APK
+        if self.file_exists('/usr/bin/zypper'):
+            return PackageManagerEnum.ZYPPER
         return PackageManagerEnum.UNKNOWN
 
     def insert_cd(self, vdi_name: str) -> None:
--- a/tests/guest_tools/unix/test_xen_guest_agent.py
+++ b/tests/guest_tools/unix/test_xen_guest_agent.py
@@ -12,7 +12,7 @@ from lib.vm import VM
 # From --hosts parameter:
 # - host(A1): first XCP-ng host >= 8.0
 # From --vm parameter:
-# - A Linux VM with systemd and a supported package manager (RPM or APT)
+# - A Linux VM with systemd and a supported package manager (DNF or APT)
 
 
 def _vif_published_ips(host: Host, xs_prefix: str, vif_id: int, proto: str) -> list[str]:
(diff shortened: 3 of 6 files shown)


tests/guest_tools/unix/test_xen_guest_agent.py
```diff
@@ -38,18 +38,18 @@ def agent_install(self, running_vm: VM, xen_guest_agent_urls: dict[str, str]) ->
...
         pkg_mgr = vm.detect_package_manager()
-        if pkg_mgr not in (PackageManagerEnum.RPM, PackageManagerEnum.APT_GET):
+        if pkg_mgr not in (PackageManagerEnum.YUM, PackageManagerEnum.DNF, PackageManagerEnum.APT_GET):
             pytest.skip(f"Package manager '{pkg_mgr}' not supported in this test")
 
         # Remove conflicting xe-guest-utilities if present
         logging.info("Removing xe-guest-utilities if present")
-        if pkg_mgr == PackageManagerEnum.RPM:
+        if pkg_mgr not in (PackageManagerEnum.YUM, PackageManagerEnum.DNF):
             vm.ssh('rpm -qa | grep xe-guest-utilities | xargs --no-run-if-empty rpm -e')
```
d3athjest3r, line 51: If the yum and/or dnf binary exists, this ALWAYS means there is an rpm binary as well? I assume yes? Otherwise this might fail, no? Or I should switch to yum/dnf to uninstall the package...

d3athjest3r (approved):
In general it looks good to me. I was just wondering if removing the check for RPM could cause trouble?

tests/guest_tools/unix/test_xen_guest_agent.py
glehmann: Assuming `rpm` is there when the system uses `yum` or `dnf` is a very reasonable assumption IMO

tests/guest_tools/unix/test_xen_guest_agent.py
```diff
@@ -38,18 +38,18 @@ def agent_install(self, running_vm: VM, xen_guest_agent_urls: dict[str, str]) ->
...
         elif pkg_mgr == PackageManagerEnum.APT_GET and \
                 vm.ssh_with_result('dpkg -l xe-guest-utilities').returncode == 0:
             vm.ssh('apt-get remove -y xe-guest-utilities')
 
-        if pkg_mgr == PackageManagerEnum.RPM:
+        if pkg_mgr not in (PackageManagerEnum.YUM, PackageManagerEnum.DNF):
```
semarie: the conditional doesn't seems right: the `not` seems wrong.

tests/guest_tools/unix/test_xen_guest_agent.py
```diff
@@ -38,18 +38,18 @@ def agent_install(self, running_vm: VM, xen_guest_agent_urls: dict[str, str]) ->
...
 
         pkg_mgr = vm.detect_package_manager()
-        if pkg_mgr not in (PackageManagerEnum.RPM, PackageManagerEnum.APT_GET):
+        if pkg_mgr not in (PackageManagerEnum.YUM, PackageManagerEnum.DNF, PackageManagerEnum.APT_GET):
             pytest.skip(f"Package manager '{pkg_mgr}' not supported in this test")
 
         # Remove conflicting xe-guest-utilities if present
         logging.info("Removing xe-guest-utilities if present")
-        if pkg_mgr == PackageManagerEnum.RPM:
+        if pkg_mgr not in (PackageManagerEnum.YUM, PackageManagerEnum.DNF):
```
semarie: the conditional doesn't seems right: the `not` seems wrong.

semarie (changes requested):

semarie (approved):
LGTM

tests/guest_tools/unix/test_xen_guest_agent.py
```diff
@@ -12,7 +12,7 @@
 # From --hosts parameter:
 # - host(A1): first XCP-ng host >= 8.0
 # From --vm parameter:
-# - A Linux VM with systemd and a supported package manager (RPM or APT)
+# - A Linux VM with systemd and a supported package manager (YUM or APT)
```
stormi:
```suggestion
# - A Linux VM with systemd and a supported package manager (YUM/DNF or APT)
```

lib/common.py
```diff
@@ -78,9 +78,10 @@ def parse_size(size_str: str) -> int:
 
 class PackageManagerEnum(Enum):
     UNKNOWN = 1
-    RPM = 2
+    YUM = 2
     APT_GET = 3
     APK = 4
+    DNF = 5
```
rzr, line 84: what about zypper ?

olivierh-pro (approved):

lib/common.py
glehmann: I've added zypper

lib/common.py
```diff
@@ -78,9 +78,11 @@ def parse_size(size_str: str) -> int:
...
+    ZIPPER = 6
```
d3athjest3r: I think it's called ZYPPER.

lib/vm.py
```diff
@@ -511,12 +511,16 @@ def file_exists(self, filepath: str, regular_file: bool = True) -> bool:
...
+        if self.file_exists('/usr/bin/dnf'):
+            return PackageManagerEnum.DNF
+        if self.file_exists('/usr/bin/yum'):
+            return PackageManagerEnum.YUM
         if self.file_exists('/usr/bin/apt-get'):
             return PackageManagerEnum.APT_GET
         if self.file_exists('/sbin/apk'):
             return PackageManagerEnum.APK
+        if self.file_exists('/usr/bin/zypper'):
+            return PackageManagerEnum.ZIPPER
```
d3athjest3r: Same as above

lib/common.py
glehmann: fixed!

lib/vm.py
glehmann: Both are in the same column, just one line apart. I should have looked more carefully!

PR #595, opened 2026-06-22
linstor: test VM startup on disk failure

Kuruyia:
This adds a test to the LINSTOR SR test suite to make sure that a with a VDI on a shared LINSTOR SR can still start and shut down when a physical disk of that SR has failed.

The test does the following:
- Fails a physical disk of the LINSTOR SR pool on a random host.
- Ensures a VM can still start up and shut down on all hosts.

This uses a device mapper to avoid relying on the capabilities of underlying block device.

Supersedes #312.


lib/common.py
```diff
@@ -215,6 +216,28 @@ def wait_for_not(
 ) -> None:
     return wait_for(fn, msg, timeout_secs, retry_delay_secs, True)
 
+def run_with_timeout(fn: Callable[[], Any], timeout_secs: int = 2 * 60) -> None:
+    queue = multiprocessing.Queue()
```
glehmann:
```suggestion
    queue: multiprocessing.Queue[Exception] = multiprocessing.Queue()
```

tests/storage/linstor/unhealthy/test_failed_disk_linstor_sr.py
```diff
@@ -0,0 +1,45 @@
...
+
+        try:
+            for host in sr.pool.hosts:
+                logging.info(f'Checking VM on host {host.hostname_or_ip}')
+
+                run_with_timeout(lambda: vm.start(on=host.uuid), timeout_secs=60)
+                vm.wait_for_os_booted()
+                vm.shutdown(verify=True)
+        finally:
+            flakey_unused_512B_disk[random_host].repair()
```
glehmann: Using `defer(lambda: flakey_unused_512B_disk[random_host].repair())` line 36 would avoid the try/finally block and keep the disk in its failed state in the debugger, in case of failure

tests/storage/linstor/unhealthy/test_failed_disk_linstor_sr.py
```diff
@@ -0,0 +1,45 @@
...
+    @pytest.mark.small_vm # run with a small VM to test the features
+    def test_linstor_sr_fail_disk(
+        self,
+        vm_on_linstor_sr: VM,
+        flakey_unused_512B_disk: dict[Host, FlakeyDisk],
+        linstor_sr: SR,
+    ) -> None:
+        sr = linstor_sr
+        vm = vm_on_linstor_sr
+        random_host = random.choice(sr.pool.hosts)
```
glehmann: Introducing randomness might make the test less stable. Is there a reason to pick the host randomly?

tests/storage/linstor/unhealthy/test_failed_disk_linstor_sr.py
```diff
@@ -0,0 +1,45 @@
...
+        # Let xcp-persistent-database come in sync across the nodes.
+        random_host.ssh('drbdadm wait-sync xcp-persistent-database')
+
+        flakey_unused_512B_disk[random_host].fail()
+
+        try:
+            for host in sr.pool.hosts:
+                logging.info(f'Checking VM on host {host.hostname_or_ip}')
+
+                run_with_timeout(lambda: vm.start(on=host.uuid), timeout_secs=60)
```
glehmann: `xe vm-start` could be hanging forever here? I'm surprised we need `run_with_timeout`.

tests/storage/linstor/unhealthy/conftest.py
```diff
@@ -0,0 +1,109 @@
...
+
+        self._apply_dm_table(self._build_dm_table(True))
+        self._host.ssh('sync')
+        self._host.ssh('echo 3 > /proc/sys/vm/drop_caches')
+
+    def repair(self) -> None:
+        logging.info(f'Repairing device {self._device.path} on {self._host.hostname_or_ip}')
+
+        self._apply_dm_table(self._build_dm_table(False))
+        cmd_res = self._host.ssh('linstor -m --controllers `xe host-list params=address --minimal` r l -n `hostname` --props DrbdOptions/SkipDisk') # noqa: E501
```
glehmann: You can just split the line instead of disabling the E501 rule
```suggestion
        cmd_res = self._host.ssh('linstor -m --controllers `xe host-list params=address --minimal` r l'
                                 ' -n `hostname` --props DrbdOptions/SkipDisk')
```

tests/storage/linstor/unhealthy/test_failed_disk_linstor_sr.py
Kuruyia: I have no reasons myself, this is more of an artifact from the previous PR. The test should be successful on all hosts equally.

We can either fix this to e.g. the first host of the list, or fail the disk and do the test on all hosts one-by-one if we want to be thorough (if we're ready to pay the extra time complexity, which would be $n^2$ in total with $n$ being the number of hosts).

tests/storage/linstor/unhealthy/test_failed_disk_linstor_sr.py
Kuruyia: When I was testing for failure (by running this test in an unsupported configuration; with 2 hosts and replication set to 1), I had the `VM.start_on` task hang seemingly indefinitely at 50% a couple of times. Most of the time, this simply fails with some "VDI is not available" error.

So this is more of a precaution against the test hanging. I was unable to reproduce that behavior reliably, but I've seen it happen. The original PR also included something similar, so I guess Rishi must have seen that behavior as well.

tests/storage/linstor/unhealthy/test_failed_disk_linstor_sr.py
glehmann: Using the first host is fine IMO

glehmann (approved):

PR #599, opened 2026-06-24
limits/vif_limit: Run 'apt update' before installing iperf3

last-genius:
Otherwise stale caches will be used and iperf3 installation will fail, on some VMs.

Tested this by running the following command, which passes with this fix:
```
pytest tests/limits/test_vif_limit.py::TestVIFLimit::test_vif_limit --vm=http://10.1.0.6/images/debian-12-created_8.2-zstd.xva
```

### DIFF
--- a/tests/limits/test_vif_limit.py
+++ b/tests/limits/test_vif_limit.py
@@ -69,6 +69,9 @@ class TestVIFLimit:
             if vm.ssh_with_result('systemctl restart networking').returncode != 0:
                 assert False, "Failed to configure networking"
 
+            # Open needed ports in the firewall
+            host.ssh(f'iptables -I INPUT -p tcp --match multiport --dports 5100:{5100+VIF_LIMIT-1} -j ACCEPT')
+
             # Test iperf on all interfaces in parallel
             # Clean up on exceptions
             logging.info('Create separate iperf servers on the host')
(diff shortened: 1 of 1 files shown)


tests/limits/test_vif_limit.py
```diff
@@ -59,6 +59,8 @@ def test_vif_limit(self, host_with_saved_yum_state: Host, vm_with_vcpu_count: VM
             vm.ssh(f'echo "{config}" >> /etc/network/interfaces')
 
             logging.info('Install iperf3 on VM and host')
+            if vm.ssh_with_result('apt update').returncode != 0:
```
stormi: The `if` and `assert` are probably unnecessary as any failure in the ssh command will raise anyway and by caught by pytest to detect failed state.

gduperrey (approved):
Job successfully executed manually and in Jenkins.

tests/limits/test_vif_limit.py
last-genius: I've tested and that's not actually the case, `ssh_with_result` does not raise on non-zero exit codes, it needs to be handled by the caller.

`vm.ssh` does raise though

stormi: Oh, you're right. But then it's not the right tool. `ssh` raises. `ssh_with_result` is here only if you really need to check the result manually, and avoiding this is the whole point of my comment.

Our tests are full of "unchecked" `ssh` commands, because `ssh` raises and pytest is good with converting exceptions into test failures (with proper teardown of fixtures afterwards etc.).

So unless when there are specific requirements, the test development principle is to send simple comments that may fail by raising, let pytest handle failures, and keep the test code simple as a consequence.

ydirson: Right, but the `iperf` launch just below was likely used as a model, and has the same "issue"

stormi: I saw that, yes. We could probably go with a follow-up PR to simplify it.

stormi (approved):
Approving, my comment can be seen as a suggestion given the context (the rest of the test already applies the principle I ask to change). A follow-up PR after the merge, to simplify the test, would be nice (not necessarily from @last-genius directly).

glehmann (approved):

tests/limits/test_vif_limit.py
last-genius: Moved to `vm.ssh`

gduperrey (approved):
It ran successfully both manually and via Jenkins on the latest commit of this PR.

tests/limits/test_vif_limit.py
```diff
@@ -100,3 +105,5 @@ def test_vif_limit(self, host_with_saved_yum_state: Host, vm_with_vcpu_count: VM
             for vif in vifs:
                 vif.destroy()
             host.ssh('killall iperf3 || true')
+            # Restore firewall
+            host.ssh('iptables -D INPUT 1')
```
stormi, line 109: This finally part is vulnerable to failures that would leave the next commands unexecuted, but that's not in this PR's scope. 

We now have a "defer" fixture, introduced by @glehmann, which can help making this. A complex finally can also be the sign that setup and teardown would better be handled in a fixture. 

So, no blocker here, but this is another improvement we can bring to this test.

stormi (approved):

glehmann (approved):

PR #600, opened 2026-06-24
Run pytest command with execvp in jobs.py

vxgmichel:
Fixes: https://github.com/xcp-ng/xcp-ng-tests/issues/596

With `subprocess.run`, both the Python parent and the pytest child are in the same process group and receive SIGINT. But `subprocess.run` internal logic then terminates the child with SIGKILL before pytest can finish its teardown. With `execvp`, the current process is replaced by pytest entirely so pytest handles SIGINT on its own and teardown runs normally.

### DIFF
--- a/jobs.py
+++ b/jobs.py
@@ -2,6 +2,7 @@
 
 import argparse
 import json
+import os
 import subprocess
 import sys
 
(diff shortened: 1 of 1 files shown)


jobs.py
```diff
@@ -811,9 +812,15 @@ def action_run(args: argparse.Namespace) -> None:
...
-    if res.returncode:
-        sys.exit(1)
+    # Use `execvp` instead of `subprocess.run` to avoid signal handling issues.
+    # With `subprocess.run`, both the Python parent and the pytest child are in
+    # the same process group and receive SIGINT. But `subprocess.run` internal
+    # logic then terminates the child with SIGKILL before pytest can finish its
+    # teardown. With `execvp`, the current process is replaced by pytest entirely
+    # so pytest handles SIGINT on its own and teardown runs normally.
+    # execvp: "v" = args as a list, "p" = resolve program via PATH.
+    os.execvp(cmd[0], cmd)
```
olivierh-pro, line 822: `cmd[0]` is supposed to be used for "program via PATH" right ? so `pytest` value I guess

glehmann (approved):

jobs.py
vxgmichel: Exactly: `cmd` is created by `build_pytest_cmd` so `cmd[0]` is `pytest`. Any issue with that?

ydirson (approved):

jobs.py
olivierh-pro: It was just for my own comprehension! :)

olivierh-pro (approved):

PR #601, opened 2026-06-25
glusterfs: Fix race condition on test_volume_stopped.

Ythogtha:
Retry sr.plug_pbds after 1s then 2s before proceding with test failure.

### DIFF
--- a/tests/storage/glusterfs/test_glusterfs_sr.py
+++ b/tests/storage/glusterfs/test_glusterfs_sr.py
@@ -54,8 +54,8 @@ class TestGlusterFSSR:
             except SSHCommandFailed:
                 logging.info("SR scan failed as expected.")
             host.ssh('gluster --mode=script volume start vol0')
+            wait_for(sr.try_plug_pbds, "Wait for glusterfs volume to start")
             volume_running = True
-            sr.plug_pbds(verify=True)
             sr.scan()
         finally:
             if not volume_running:
--- a/lib/sr.py
+++ b/lib/sr.py
@@ -80,6 +80,13 @@ class SR:
         if verify:
             wait_for(self.all_pbds_attached, "Wait for PBDs attached")
 
+    def try_plug_pbds(self) -> bool:
+        try:
+            self.plug_pbds(verify=True)
+            return True
+        except commands.SSHCommandFailed:
+            return False
+
     def vdi_uuids(self, managed: bool = False, name_label: str | None = None) -> list[str]:
         args: dict[str, str | bool | dict[str, str]] = {
             'sr-uuid': self.uuid,


DatBewar (approved):

stormi (changes requested):
The conventional commits convention is not applied on this repository (in the XCP-ng team it's mostly used by the storage team on their own repos).

tests/storage/glusterfs/test_glusterfs_sr.py
```diff
@@ -54,8 +55,17 @@ def test_volume_stopped(self, host: Host, glusterfs_sr: SR) -> None:
...
+            # as the volume is not yet started.
+            # This problem is test-specific, hence the retry here instead of upstream.
+            for i in range(3):
+                sleep(i)
+                try:
+                    sr.plug_pbds(verify=True)
+                    volume_running = True
+                except SSHCommandFailed:
+                    continue
+                break
```
Millefeuille42: You can use the `wait_for` function instead:

```suggestion
            def try_plug_pbds():
                try:
                    sr.plug_pbds(verify=True)
                    return True
                except SSHCommandFailed:
                    return False

            wait_for(
                try_plug_pbds,
                "Wait for system installation to succeed",
                timeout_secs=40 * 60,
                retry_delay_secs=1
            )
            volume_running = True
```

You could even put `try_plug_pbds` in the lib or something.

Millefeuille42 (changes requested):

tests/storage/glusterfs/test_glusterfs_sr.py
Ythogtha: You're really suggesting to retry for 40 minutes ?
Isn't it a bit too much for a race condition ?
Shouldn't we have a failed test, and a proper investigation into such slowness, if even 1 minute is required for that ?

Ythogtha: I pushed with default values, which seems ok to me (2*60, retry 2s), but if you have any good reason to change that, I'm listening.

stormi (changes requested):
LGTM.

Let's squash the commits before we risk merging as is.

The commit message will need updating, too.

stormi (approved):

tests/storage/glusterfs/test_glusterfs_sr.py
Millefeuille42: Wrong copy-paste I didn't intended for this delay, you can put whatever you see fit. My bad.

Millefeuille42 (approved):

glehmann (approved):

rzr (approved):

PR #603, opened 2026-06-30
Allow concurrent use of url_download

vxgmichel:
When two pytest instances run in parallel and download the same ISO using `url_download`, the resulting file ends up corrupted. This PR uses `tempfile.NamedTemporaryFile` to avoid writing to the same file.

The new implementation also remove the temporary `.part` file if the download fails.

It also increases the chunksize to 64k to improve performance:
```
# 3GB download with 128 chunksize
83,80s user 12,64s system 96% cpu 1:40,10 total
# 3ith 64k chunksize
4,34s user 5,94s system 24% cpu 41,330 total
```

### DIFF
--- a/lib/common.py
+++ b/lib/common.py
@@ -10,11 +10,13 @@ import os
 import random
 import string
 import sys
+import tempfile
 import time
 import traceback
 from datetime import datetime
 from enum import Enum
 from functools import lru_cache
+from pathlib import Path
 from uuid import UUID
 
 import requests
@@ -317,13 +319,30 @@ def strtobool(val: str | None) -> bool:
     raise ValueError("invalid truth value '{}'".format(val))
 
 def url_download(url: str, filename: str) -> None:
-    r = requests.get(url, stream=True)
-    r.raise_for_status()
-    tempfilename = filename + ".part"
-    with open(tempfilename, 'wb') as fd:
-        for chunk in r.iter_content(chunk_size=128):
-            fd.write(chunk)
-    os.rename(tempfilename, filename)
+    """
+    Download the content of `url` to the `filename` destination.
+
+    A randomized filename is used during download to prevent file corruption on
+    concurrent use. If the download fails then the temporary file is removed.
+    """
+    destination = Path(filename)
+    destination.parent.mkdir(parents=True, exist_ok=True)
+    with requests.get(url, stream=True) as r:
+        r.raise_for_status()
+        temp_name: str | None = None
+        try:
+            with tempfile.NamedTemporaryFile(
+                dir=destination.parent, prefix=f"{destination.name}.", suffix=".part", delete=False
+            ) as fd:
+                temp_name = fd.name
+                for chunk in r.iter_content(chunk_size=128):
+                    fd.write(chunk)
+        except BaseException:
+            if temp_name is not None:
+                os.unlink(temp_name)
+            raise
+        else:
+            os.rename(temp_name, filename)
 
 def randid(length: int = 6) -> str:
     """
--- a/lib/common.py
+++ b/lib/common.py
@@ -335,7 +335,7 @@ def url_download(url: str, filename: str) -> None:
                 dir=destination.parent, prefix=f"{destination.name}.", suffix=".part", delete=False
             ) as fd:
                 temp_name = fd.name
-                for chunk in r.iter_content(chunk_size=128):
+                for chunk in r.iter_content(chunk_size=64 * 1024):
                     fd.write(chunk)
         except BaseException:
             if temp_name is not None:


lib/common.py
```diff
@@ -317,13 +319,24 @@ def strtobool(val: str | None) -> bool:
...
+                dir=target.parent, prefix=f"{target.name}.", suffix=".part", delete=False
+            ) as fd:
+                temp_name = fd.name
+                for chunk in r.iter_content(chunk_size=64 * 1024):
+                    fd.write(chunk)
+        except BaseException:
+            if temp_name is not None:
+                os.unlink(temp_name)
+            raise
+        else:
```
olivierh-pro, line 344: Does the `else` really needed here? In previous code, rename was done outside of "with".

lib/common.py
```diff
@@ -317,13 +319,24 @@ def strtobool(val: str | None) -> bool:
...
+    target.parent.mkdir(parents=True, exist_ok=True)
+    with requests.get(url, stream=True) as r:
+        r.raise_for_status()
+        temp_name: str | None = None
+        try:
+            with tempfile.NamedTemporaryFile(
+                dir=target.parent, prefix=f"{target.name}.", suffix=".part", delete=False
+            ) as fd:
+                temp_name = fd.name
+                for chunk in r.iter_content(chunk_size=64 * 1024):
```
olivierh-pro, line 338: Is there a reason to change chunk_size from 128 to "64 x 1024"?

lib/common.py
```diff
@@ -317,13 +319,24 @@ def strtobool(val: str | None) -> bool:
...
 
 def url_download(url: str, filename: str) -> None:
-    r = requests.get(url, stream=True)
-    r.raise_for_status()
-    tempfilename = filename + ".part"
-    with open(tempfilename, 'wb') as fd:
-        for chunk in r.iter_content(chunk_size=128):
-            fd.write(chunk)
-    os.rename(tempfilename, filename)
+    target = Path(filename)
```
olivierh-pro: nitpick: I would have renamed "target" into "dest_file" (imho)

lib/common.py
```diff
@@ -317,13 +319,24 @@ def strtobool(val: str | None) -> bool:
     raise ValueError("invalid truth value '{}'".format(val))
 
 def url_download(url: str, filename: str) -> None:
```
olivierh-pro, line 321: Should be add a short docstring to describe the new implementation logic?

lib/common.py
vxgmichel: > Does the else really needed here?

`else` is not needed, I typically use to convey the meaning of:
- either it fails and we remove the tempfile
- or it succeeds and we rename the tempfile

If you find it confusing I can simply remove it.

>  In previous code, rename was done outside of "with".

It's sill is done outside of the temporary file context, but yes I added a new context for the request (so we close the socket before leaving the function).

lib/common.py
vxgmichel: Yes, notice the separate commit and the performance comparison in the description:
```
# 3GB download with 128 chunksize
83,80s user 12,64s system 96% cpu 1:40,10 total
# 3ith 64k chunksize
4,34s user 5,94s system 24% cpu 41,330 total
```

lib/common.py
vxgmichel: Renamed: ec5699abe92b7ecd8dc34c61769a263271d49819

lib/common.py
vxgmichel: Added: ec5699abe92b7ecd8dc34c61769a263271d49819

lib/common.py
rzr: at the price of more RAM used right?  I would rely on standard values unless it's a problem we want to address. but i am not blocking this change

lib/common.py
```diff
@@ -317,13 +319,30 @@ def strtobool(val: str | None) -> bool:
...
-    tempfilename = filename + ".part"
-    with open(tempfilename, 'wb') as fd:
-        for chunk in r.iter_content(chunk_size=128):
-            fd.write(chunk)
-    os.rename(tempfilename, filename)
+    """
+    Download the content of `url` to the `filename` destination.
+
+    A randomized filename is used during download to prevent file corruption on
+    concurrent use. If the download fails, the temporary file is removed.
```
rzr:
```suggestion
    concurrent use. If the download fails then the temporary file is removed.
```

rzr (approved):

lib/common.py
vxgmichel: > at the price of more RAM used right?

Python alone is 15MB in RAM, I think we can afford 64KB :)

> I would rely on standard values

I don't know any standard value for this. I'd say anything between 1KB and 1MB is pretty common for this kind of use case

> unless it's a problem we want to address

I'd say that 2x slowdown on ISO download is a problem we want to address

lib/common.py
olivierh-pro: LGTM!

olivierh-pro (approved):

glehmann (approved):

lib/common.py
```diff
@@ -317,13 +319,30 @@ def strtobool(val: str | None) -> bool:
...
+        temp_name: str | None = None
+        try:
+            with tempfile.NamedTemporaryFile(
+                dir=destination.parent, prefix=f"{destination.name}.", suffix=".part", delete=False
+            ) as fd:
+                temp_name = fd.name
+                for chunk in r.iter_content(chunk_size=64 * 1024):
+                    fd.write(chunk)
+        except BaseException:
+            if temp_name is not None:
```
ydirson, line 341: `if temp_name:` would be sufficient

ydirson (changes requested):
OK on the principle, but commit messages should give context about why we need the change.

lib/common.py
vxgmichel: Updated: [4e15fde](https://github.com/xcp-ng/xcp-ng-tests/pull/603/commits/4e15fde8e4b6b328faafac6189ecf1d7c55ad2a7)

lib/common.py
vxgmichel: I deliberately used `is not None` since I'm testing for types rather than truthfulness. Do you see any issue with that?

For reference, a comment from [PEP8](https://peps.python.org/pep-0008/) about this very subject:
> Also, beware of writing `if x` when you really mean `if x is not None` – e.g. when testing whether a variable or argument that defaults to None was set to some other value. The other value might have a type (such as a container) that could be false in a boolean context!

I also added descriptions to both commit :+1:

ydirson: > I deliberately used is not None since I'm testing for types rather than truthfulness.

The only False-like value of `str` should never be returned, so checking the precise value does not add much.

> Do you see any issue with that?

It's just that it would feels (to me) a bit more pythonic, but that may be a matter of taste.

ydirson (approved):

PR #604, opened 2026-06-30
Fix automated install tests to be compatible with the postinstall job

vxgmichel:
The `postinstall` job requires those three tests which depends on the `unused_512B_disks` fixture:
```
tests/misc/test_export.py::TestExport::test_export_zstd[mini-linux-x86_64-bios]
tests/misc/test_export.py::TestExport::test_export_gzip[mini-linux-x86_64-bios]
tests/misc/test_export.py::TestExport::test_export_uncompressed[mini-linux-x86_64-bios]
```

In order to create nested hosts that are compliant with the requirements of the `postinstall` job, an extra disk has to be added.

Also fixes the regression described in #602

### DIFF
--- a/lib/commands.py
+++ b/lib/commands.py
@@ -163,7 +163,7 @@ def _ssh(
 
     # Even if check is False, we still raise in case of return code 255, which means a SSH error.
     if res.returncode == 255:
-        return SSHCommandFailed(255, "SSH Error: %s" % ssherr, cmd, ssherr=ssherr)
+        return SSHCommandFailed(255, "SSH Error: %s" % output_for_errors, cmd, ssherr=ssherr)
 
     output: bytes = res.stdout
     if banner_res:
--- a/lib/installer.py
+++ b/lib/installer.py
@@ -4,7 +4,7 @@ import logging
 import time
 import xml.etree.ElementTree as ET
 
-from lib.commands import SSHCommandFailed, ssh
+from lib.commands import ssh
 from lib.common import wait_for
 
 from typing import Any, Self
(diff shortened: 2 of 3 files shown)


lib/commands.py
```diff
@@ -163,7 +163,7 @@ def _ssh(
 
     # Even if check is False, we still raise in case of return code 255, which means a SSH error.
     if res.returncode == 255:
-        return SSHCommandFailed(255, "SSH Error: %s" % ssherr, cmd, ssherr=ssherr)
+        return SSHCommandFailed(255, "SSH Error: %s" % output_for_errors, cmd, ssherr=ssherr)
```
olivierh-pro, line 166:
```suggestion
        return SSHCommandFailed(255, f"SSH Error: {output_for_errors}", cmd, ssherr=ssherr)
```

vxgmichel: At the moment this commit is just a revert of this change:
https://github.com/xcp-ng/xcp-ng-tests/commit/336403d3517c7b71e4550ec976e5ed01f80e3fbb#diff-9f19e50d8cf1b7e0e462f14cdc77200e48d53a73eb1bbd78e574d24d16d5def7L140-R166

I'm waiting for a discussion on #602 before implementing anything else than the required revert.

olivierh-pro (approved):

glehmann (approved):

ydirson (changes requested):
As written in another comment I'd prefer we deal differently with the file exclusion, otherwise I'm OK with the first 3 commits - though they really don't look like they belong in a PR with the current title, that title should be fixed too :)

pyproject.toml
```diff
@@ -38,6 +38,10 @@ dev = [
 
 [tool.pyright]
 typeCheckingMode = "standard"
+exclude = [".venv/", "installimg.*"]
+
+[tool.mypy]
+exclude = ["^installimg\\."]
```
glehmann: We should probably also exclude them in .gitignore

vxgmichel: Agreed :+1:  I moved  the commit here: https://github.com/xcp-ng/xcp-ng-tests/pull/628

glehmann (approved):

ydirson (approved):

PR #606, opened 2026-07-02
Update dependabot configuration

stormi:
- target branch: dependabot-staging, so that we can merge dependabot PRs faster while still having a buffer on which to run CI first before merging to master branch.
- regroup PRs into one per ecosystem to reduce the noise.

I don't really know how to test the configuration without merging it first and triggering a dependabot update.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,13 +6,29 @@ updates:
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
-      interval: "daily"
+      interval: "weekly"
+      day: "sunday"
     cooldown:
       default-days: 7
+    groups:
+      ci-deps:
+        patterns:
+          - "*"
+
   - package-ecosystem: "uv"
     directory: "/"
     schedule:
-      interval: "daily"
+      interval: "weekly"
+      day: "sunday"
     open-pull-requests-limit: 10
     cooldown:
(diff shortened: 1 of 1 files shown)


LucienLassalle (approved):
LGTM

vxgmichel (approved):
LGTM, I think the use of groups is the proper way to reduce the dependabot PR noise.

One suggestion, maybe for a future PR: it might make sense to have two groups:
1. the [base dependencies](https://github.com/xcp-ng/xcp-ng-tests/blob/f4de6f73b56093d19664ba420c58330b7b683da0/pyproject.toml#L7-L17) group (pytest, requests, pydantic, etc.) 
2. the [dev dependencies](https://github.com/xcp-ng/xcp-ng-tests/blob/f4de6f73b56093d19664ba420c58330b7b683da0/pyproject.toml#L20-L37) group (pyright, ruff, prek, etc.)

The second group can be validated and merged much more easily, since having the github CI validation should already give a strong signal that the dependency update doesn't bring any regressions.

glehmann (approved):
LGTM
I also like the split in base and dev groups for uv, but we can do that later

LucienLassalle (approved):

vxgmichel (approved):

glehmann (approved):

PR #625, opened 2026-07-07
Add test for rolling pool update with VM availability check on updated host

DatBewar:
New pr based on https://github.com/xcp-ng/xcp-ng-tests/pull/310
```
Added test_linstor_sr_pool_update for rolling pool update with VM availability check.

It performs

    A rolling update and reboot of all hosts in a LINSTOR SR pool (starting with the master).
    Verifies VM can start and shutdown successfully on each host after update.
    Ensures SR remains usable throughout the process.
```

### DIFF
--- a/lib/host.py
+++ b/lib/host.py
@@ -480,11 +480,6 @@ class Host:
         else:
             return self.xe('vm-list', {'uuid': vm_uuid}, minimal=True) == vm_uuid
 
-    def install_updates(self, enablerepo: str | None = None) -> str:
-        logging.info("Install updates on host %s" % self)
-        enablerepo_arg = f" --enablerepo={enablerepo}" if enablerepo is not None else ""
-        return self.ssh(f"yum update -y{enablerepo_arg}")
-
     def get_system_uuid(self) -> str:
         """Get system uuid of current host.
 
@@ -611,11 +606,10 @@ class Host:
             # If XAPI is not ready yet, or the host is down, this will throw. We return False in that case.
             return False
 
-    def has_updates(self, enablerepo: str | None = None) -> bool:
-        enablerepo_arg = f" --enablerepo={enablerepo}" if enablerepo is not None else ""
+    def has_updates(self) -> bool:
         try:
             # yum check-update returns 100 if there are updates, 1 if there's an error, 0 if no updates
-            self.ssh(f"yum check-update{enablerepo_arg}")
+            self.ssh('yum check-update')
             # returned 0, else there would have been a SSHCommandFailed
             return False
         except commands.SSHCommandFailed as e:
--- a/lib/host.py
+++ b/lib/host.py
@@ -480,6 +480,11 @@ class Host:
         else:
             return self.xe('vm-list', {'uuid': vm_uuid}, minimal=True) == vm_uuid
 
+    def install_updates(self, enablerepo=None):
+        logging.info("Install updates on host %s" % self)
+        enablerepo_cmd = ['--enablerepo=%s' % enablerepo] if enablerepo is not None else []
+        return self.ssh(f"yum update -y {enablerepo_cmd}")
+
     def get_system_uuid(self) -> str:
         """Get system uuid of current host.
 
@@ -606,10 +611,11 @@ class Host:
             # If XAPI is not ready yet, or the host is down, this will throw. We return False in that case.
             return False
 
-    def has_updates(self) -> bool:
+    def has_updates(self, enablerepo=None):
+        enablerepo_cmd = ['--enablerepo=%s' % enablerepo] if enablerepo is not None else []
         try:
             # yum check-update returns 100 if there are updates, 1 if there's an error, 0 if no updates
-            self.ssh('yum check-update')
+            self.ssh(f"yum check-update {enablerepo_cmd}")
             # returned 0, else there would have been a SSHCommandFailed
             return False
         except commands.SSHCommandFailed as e:
--- a/lib/host.py
+++ b/lib/host.py
@@ -480,10 +480,10 @@ class Host:
         else:
             return self.xe('vm-list', {'uuid': vm_uuid}, minimal=True) == vm_uuid
 
-    def install_updates(self, enablerepo=None):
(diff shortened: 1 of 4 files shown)


lib/host.py
```diff
@@ -480,6 +480,11 @@ def pool_has_vm(self, vm_uuid: str, vm_type: str = 'vm') -> bool:
         else:
             return self.xe('vm-list', {'uuid': vm_uuid}, minimal=True) == vm_uuid
 
+    def install_updates(self, enablerepo: str | None = None) -> str:
```
glehmann, line 483: We now have a `yum_update()`, I don't think we need this new method.

tests/storage/linstor/conftest.py
```diff
@@ -130,7 +146,10 @@ def check_linstor_installed(host: Host) -> None:
     def install_linstor(host: Host) -> None:
         logging.info(f"Installing {LINSTOR_PACKAGE} on host {host}...")
         host.yum_install([LINSTOR_RELEASE_PACKAGE])
-        host.yum_install([LINSTOR_PACKAGE], enablerepo="xcp-ng-linstor-testing")
+        if _linstor_upgrade_test:
+            host.yum_install([LINSTOR_PACKAGE], enablerepo="xcp-ng-linstor-testing")
```
glehmann, line 150: This is a `package`-scoped fixture, so wouldn't this be run only once, independently of the marker on the test?

lib/host.py
DatBewar: Yes I'll use that thanks

tests/storage/linstor/conftest.py
DatBewar: You are right, I will change that too
Thanks

Ythogtha (approved):
Feels the way I'd be doing it manually for storage, so I'm ok with this test.

tests/storage/linstor/conftest.py
```diff
@@ -130,7 +131,9 @@ def ensure_linstor_not_installed(host: Host) -> None:
     def install_linstor(host: Host) -> None:
         logging.info(f"Installing {LINSTOR_PACKAGE} on host {host}...")
         host.yum_install([LINSTOR_RELEASE_PACKAGE])
-        host.yum_install([LINSTOR_PACKAGE], enablerepo="xcp-ng-linstor-testing")
+        # Always install from the stable repo. Upgrade tests pull newer packages
+        # from xcp-ng-linstor-testing themselves via yum_update(enablerepos=...).
+        host.yum_install([LINSTOR_PACKAGE])
```
glehmann: This would impact all the other tests as well, wouldn't it?
Do we want to test the stable packages in the other tests?

tests/storage/linstor/test_linstor_sr.py
```diff
@@ -251,6 +253,57 @@ def test_linstor_missing(self, linstor_sr: SR, host: Host) -> None:
...
+                executor.map(reboot_updated, updates_applied)
+            wait_for(sr.all_pbds_attached, "Wait for PBD attached")
+
+        # Ensure VM is able to boot on all the hosts
+        for h in hosts:
+            vm.start(on=h.uuid)
+            vm.wait_for_os_booted()
+            vm.shutdown(verify=True)
+
+        sr.scan()
```
glehmann: What is this SR scan for?

pytest.ini
```diff
@@ -5,6 +5,7 @@ markers =
     default_vm: mark a test with a default VM in case no --vm parameter was given.
 
     # *** Markers used to select tests at collect stage ***
+    upgrade_test: mark a test which will upgrade packages from testing repo
```
glehmann: Do we need that marker for something specific?

tests/storage/linstor/test_linstor_sr.py
DatBewar: It's a kind of check of the SR, we can remove it since we already check the VM

tests/storage/linstor/conftest.py
DatBewar: Yes you are right, I will let the behavior like before.
I don't think the goal was to test the stable packages in the other tests

pytest.ini
DatBewar: It seems not, it was only used for this new test so we can remove it for now

glehmann (approved):

tests/storage/linstor/conftest.py
```diff
@@ -113,9 +113,10 @@ def pool_with_linstor(
...
     lvm_disks: None,
     pool_with_saved_yum_state: Pool,
-    _linstor_config: LinstorConfig
+    _linstor_config: LinstorConfig,
 ) -> Generator[Pool, None, None]:
     import concurrent.futures
+
     pool = pool_with_saved_yum_state
 
     def ensure_linstor_not_installed(host: Host) -> None:
```
stormi, line 122: This doesn't seem to be directly related to this PR. Should at list be in a dedicated commit?

stormi (changes requested):
The full history should be squashed before approval, IMO, to avoid merging it as is. I know that it was initially started by Rishi, but the code changed so much since that I don't think it's necessary to keep the history. We don't want PRs that do things then undo them. I would at least expect reviewers to say something about it.

But my biggest concern here is that I don't see how this test can produce any other result than "No updates available for host". The fixture installs linstor from xcp-ng-linstor-testing, then we try to update from xcp-ng-linstor-testing. Or am I missing something?

In my opinion, this PR started with an implementation when it should have started with a design. Then we asked @DatBewar to fix it but should have stepped back instead, and designed it first.

The core of  the test is probably fine. But update tests depend on the setup and here that's the main obstacle. We must **ENSURE** that there will be updates to test. We must even make sure we know which updates we're actually testing.

Discussions on the design, involving the platform team, would have also shown that testing updates is a vast topic, that raises logistical questions as it depends entirely on the initial state of the pool that we're testing. And is really useful if we make it a rolling pool update with VMs running.

My proposition for what's next is: a call, with the relevant people from storage team and platform team.

Now, maybe I missed something obvious and the test is actually useful the way it's been approved, don't hesitate to tell me.

CC @glehmann

PR #626, opened 2026-07-09
test_private_network: don't assume imported_vm to have only 1 NIC

semarie:
`imported_vm` is taken from `--vm` parameter and could have more than 1 NIC, resulting the test to fail for no good reason.

### DIFF
--- a/tests/network/test_network.py
+++ b/tests/network/test_network.py
@@ -26,11 +26,13 @@ class TestNetwork:
         vm2 = imported_vm.clone()
         defer(lambda: vm2.destroy())
 
-        vif_1_1 = vm1.create_vif(1, network_uuid=network.uuid)
-        vif_2_1 = vm2.create_vif(1, network_uuid=network.uuid)
+        n1 = len(vm1.vifs())
+        n2 = len(vm2.vifs())
+        vif_1_1 = vm1.create_vif(n1, network_uuid=network.uuid)
+        vif_2_1 = vm2.create_vif(n2, network_uuid=network.uuid)
 
-        assert len(vm1.vifs()) == 2, "VM1 should have 2 NICs"
-        assert len(vm2.vifs()) == 2, "VM2 should have 2 NICs"
+        assert len(vm1.vifs()) == n1 + 1, "VM1 should have 1 more NICs"
+        assert len(vm2.vifs()) == n2 + 1, "VM2 should have 1 more NICs"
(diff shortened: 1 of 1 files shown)


tests/network/test_network.py
```diff
@@ -26,11 +26,9 @@ def test_private_network(self, host: Host, empty_network: Network, imported_vm:
         vm2 = imported_vm.clone()
         defer(lambda: vm2.destroy())
 
-        vif_1_1 = vm1.create_vif(1, network_uuid=network.uuid)
-        vif_2_1 = vm2.create_vif(1, network_uuid=network.uuid)
+        vif_1_1 = vm1.create_vif(len(vm1.vifs()), network_uuid=network.uuid)
+        vif_2_1 = vm2.create_vif(len(vm2.vifs()), network_uuid=network.uuid)
 
-        assert len(vm1.vifs()) == 2, "VM1 should have 2 NICs"
-        assert len(vm2.vifs()) == 2, "VM2 should have 2 NICs"
```
last-genius, line 33: Should you still verify that the VIF was created successfully?

last-genius (changes requested):

tests/network/test_network.py
semarie: not necessary. there is other test for specifically check that VIF management is working.
but it doesn't hurt. I will fix it.

last-genius (approved):

tests/network/test_network.py
```diff
@@ -26,11 +26,13 @@ def test_private_network(self, host: Host, empty_network: Network, imported_vm:
...
-        vif_1_1 = vm1.create_vif(1, network_uuid=network.uuid)
-        vif_2_1 = vm2.create_vif(1, network_uuid=network.uuid)
+        n1 = len(vm1.vifs())
+        n2 = len(vm2.vifs())
+        vif_1_1 = vm1.create_vif(n1, network_uuid=network.uuid)
+        vif_2_1 = vm2.create_vif(n2, network_uuid=network.uuid)
 
-        assert len(vm1.vifs()) == 2, "VM1 should have 2 NICs"
-        assert len(vm2.vifs()) == 2, "VM2 should have 2 NICs"
+        assert len(vm1.vifs()) == n1 + 1, "VM1 should have 1 more NICs"
```
rzr, line 34: nitpicking: I would move assert after each creation

rzr (approved):

glehmann (approved):

PR #628, opened 2026-07-10
Exclude installimg directories from versioning and type checking

vxgmichel:
Follow up after https://github.com/xcp-ng/xcp-ng-tests/pull/604#issuecomment-4915916318

The `installimg.XXXXXX` directories are created at the root of the repository by `iso-remaster.sh` which is used by the `remastered_iso` fixture in the context of the automated installation tests. Those directories contain python2 files so they have to be excluded from type checking. It's not necessary to exclude them for the other checks (ruff, flake8, autopep8) since those checks explicitly run on versioned python files. The `installimg.*` pattern is also added to `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ data.py
 vm_data.py
 /scripts/guests/windows/id_rsa.pub
 .envrc
+installimg.*
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,6 +41,17 @@ dev = [
 
 [tool.pyright]
 typeCheckingMode = "standard"
+# Pyright doesn't support excluding files from .gitignore like mypy does.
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -4,3 +4,4 @@ data.py
 vm_data.py
 /scripts/guests/windows/id_rsa.pub
 .envrc
+installimg.*
```
glehmann: A new line char is missing here

glehmann (approved):

olivierh-pro (approved):

.gitignore
vxgmichel: Good catch thanks: 88bd812175c5c18dcffc1797548efa4c6447370e

PR #630, opened 2026-07-13
Run iso-remaster in temporary iso directory 

vxgmichel:
Supersede #628

Without this change, the following temporary files and directories are created at the root of the repository:
- fakerootsave.XXXXXX
- installimg.XXXXXX
- isorw.XXXXXX

They get removed in most cases, but they might remain when an error occurs during the remastering. This is especially a problem for type checking tools that might fail due to the python2 files in `installimg`. Since the `remastered_iso` fixture already creates a temporary directory in `/tmp` for other files used or produced by `iso-remaster.sh`, it makes sense to run it in this directory so those temporary files are created in there.

Also, this PR adds a `remastered-iso-` prefix to the corresponding temporary directory. This way, the temporary directory created by the `remastered_iso` fixture can be found more easily in the `/tmp` directory.

### DIFF
--- a/lib/commands.py
+++ b/lib/commands.py
@@ -316,13 +316,22 @@ def sftp(
     return res
 
 @overload
-def local_cmd(cmd: List[str], *, check: bool = True, decode: Literal[True] = True) -> LocalCommandResult[str]:
-    ...
-@overload
-def local_cmd(cmd: List[str], *, check: bool = True, decode: Literal[False]) -> LocalCommandResult[bytes]:
-    ...
 def local_cmd(
-    cmd: List[str], *, check: bool = True, decode: bool = True
+    cmd: List[str], *, check: bool = True, decode: Literal[True] = True,
+    cwd: str | os.PathLike[str] | None = None,
+) -> LocalCommandResult[str]:
+    ...
+
+@overload
+def local_cmd(
+    cmd: List[str], *, check: bool = True, decode: Literal[False],
+    cwd: str | os.PathLike[str] | None = None,
+) -> LocalCommandResult[bytes]:
+    ...
+
+def local_cmd(
+    cmd: List[str], *, check: bool = True, decode: bool = True,
+    cwd: str | os.PathLike[str] | None = None,
 ) -> LocalCommandResult[str] | LocalCommandResult[bytes]:
     """ Run a command locally on tester end. """
     logging.debug("[local] %s", (cmd,))
@@ -330,7 +339,8 @@ def local_cmd(
         cmd,
         stdout=subprocess.PIPE,
         stderr=subprocess.STDOUT,
-        check=False
+        check=False,
+        cwd=cwd,
     )
 
     # get a decoded version of the output in any case, replacing potential errors
--- a/tests/install/conftest.py
+++ b/tests/install/conftest.py
@@ -268,7 +268,7 @@ sed -i "${{SED_COMMANDS[@]}}" \
                    "--install-patcher", img_patcher_script,
                    "--iso-patcher", iso_patcher_script,
                    iso_file, remastered_iso
-                   ])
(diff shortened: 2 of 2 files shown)


lib/commands.py
```diff
@@ -316,21 +316,31 @@ def sftp(
     return res
 
 @overload
-def local_cmd(cmd: List[str], *, check: bool = True, decode: Literal[True] = True) -> LocalCommandResult[str]:
+def local_cmd(
+    cmd: List[str], *, check: bool = True, decode: Literal[True] = True,
+    cwd: str | bytes | os.PathLike[str] | os.PathLike[bytes] | None = None,
```
ydirson: Do we really need the `bytes` variants?

ydirson (commented):
Basically OK but

lib/commands.py
glehmann: I agree, we don't need it currently. Let's keep things simple, and add it later if we need it :)

vxgmichel: I simply copied the type that `_typeshed` reports:
```python
[...]
    def run(
        [...]
        cwd: StrOrBytesPath | None = None,
[...]
StrOrBytesPath: TypeAlias = str | bytes | PathLike[str] | PathLike[bytes]
```

I agree however that this long type can be confusing, what do you think about importing `StrOrBytesPath` from `_typeshed` instead? (no new requirements needed, it's used by both mypy and pyright)

```diff
diff --git a/lib/commands.py b/lib/commands.py
index 961bde2..54621d9 100644
--- a/lib/commands.py
+++ b/lib/commands.py
@@ -13,6 +13,8 @@ from lib.netutil import wrap_ip
 from typing import TYPE_CHECKING, Generic, List, Literal, TypeVar, overload

 if TYPE_CHECKING:
+    from _typeshed import StrOrBytesPath
+
     from lib.common import HostAddress

 class BaseCommandFailed(Exception):
@@ -318,20 +320,20 @@ def sftp(
 @overload
 def local_cmd(
     cmd: List[str], *, check: bool = True, decode: Literal[True] = True,
-    cwd: str | bytes | os.PathLike[str] | os.PathLike[bytes] | None = None,
+    cwd: StrOrBytesPath | None = None,
 ) -> LocalCommandResult[str]:
     ...

 @overload
 def local_cmd(
     cmd: List[str], *, check: bool = True, decode: Literal[False],
-    cwd: str | bytes | os.PathLike[str] | os.PathLike[bytes] | None = None,
+    cwd: StrOrBytesPath | None = None,
 ) -> LocalCommandResult[bytes]:
     ...

 def local_cmd(
     cmd: List[str], *, check: bool = True, decode: bool = True,
-    cwd: str | bytes | os.PathLike[str] | os.PathLike[bytes] | None = None,
+    cwd: StrOrBytesPath | None = None,
 ) -> LocalCommandResult[str] | LocalCommandResult[bytes]:
     """ Run a command locally on tester end. """
     logging.debug("[local] %s", (cmd,))
```

glehmann (approved):

lib/commands.py
glehmann: > I agree however that this long type can be confusing, what do you think about importing `StrOrBytesPath` from `_typeshed` instead? (no new requirements needed, it's used by both mypy and pyright)

But it's a private module, we're not supposed to do that. Maybe rather just define that type in xcp-ng-tests?

```python
StrOrBytesPath: TypeAlias = str | bytes | PathLike[str] | PathLike[bytes]  # stable
```

vxgmichel: Alright I restricted the type :+1: 
https://github.com/xcp-ng/xcp-ng-tests/pull/630/changes/d7a09da104263eb7c1f77d44f3a64a577e127990

ydirson (approved):

PR #634, opened 2026-07-20
Better error when importing a VM using clone:// but the provided base VM does not exist

vxgmichel:
A confusing error is shown when one tries to run a test on a nested host using a "cached" VM to be cloned as reference (typically produced by the auto install procedure in `tests/install`) but the reference VM does not exist:

```shell
$ uv run pytest --hosts "cache://some-non-existing-nested-host" --nest 10.30.36.1
[...]
E           lib.commands.SSHCommandFailed: SSH command (xe vm-import url=clone://some-non-existing-nested-host sr-uuid=61b22c4f-35f3-234a-e94a-96e5a4a0b696) failed with return code 1: The server failed to handle your request, due to an internal error. The given message may give details useful for debugging the problem.
E           message: Failure("Unsupported URI scheme: clone")
```

With this fix applied, the following error is produced instead:
```shell
$ uv run pytest --hosts "cache://some-non-existing-nested-host" --nest 10.30.36.1
[...]
E                   RuntimeError: VM 'some-non-existing-nested-host' not in cache (in SR 61b22c4f-35f3-234a-e94a-96e5a4a0b696)
```

### DIFF
--- a/lib/host.py
+++ b/lib/host.py
@@ -393,7 +393,6 @@ class Host:
         vm: VM | None = None
 
         if uri.startswith("clone://") or uri.startswith("clone+start://"):
-            assert use_cache, "clone URIs require cache enabled"
             assert sr_uuid is not None
             protocol, filename = uri.split("://", maxsplit=1)
             base_vm = self.cached_vm(filename, sr_uuid)
--- a/lib/host.py
+++ b/lib/host.py
@@ -392,8 +392,7 @@ class Host:
     def import_vm(self, uri: str, sr_uuid: str | None = None, use_cache: bool = False) -> VM:
         vm: VM | None = None
 
-        # Handling of 'clone://' and 'clone+start://' URI
-        if '://' in uri and uri.startswith("clone"):
+        if uri.startswith("clone://") or uri.startswith("clone+start://"):
             assert use_cache, "clone URIs require cache enabled"
             assert sr_uuid is not None
             protocol, filename = uri.split("://", maxsplit=1)
@@ -402,7 +401,7 @@ class Host:
                 raise RuntimeError(f"VM {filename!r} not in cache (in SR {sr_uuid})")
             vm = base_vm.clone()
             vm.param_clear('name-description')
-            if protocol.startswith("clone+start"):
+            if protocol == "clone+start":
                 vm.start()
                 wait_for(vm.is_running, f"[{self}] Wait for VM running ({vm.uuid})")
             return vm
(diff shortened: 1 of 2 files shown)


lib/host.py
```diff
@@ -391,25 +391,27 @@ def cached_vm(self, uri: str, sr_uuid: str) -> VM | None:
 
     def import_vm(self, uri: str, sr_uuid: str | None = None, use_cache: bool = False) -> VM:
         vm: VM | None = None
+
+        # Handling of 'clone://' and 'clone+start://' URI
+        if '://' in uri and uri.startswith("clone"):
```
```suggestion
        if uri.startswith('clone://') or uri.startswith('clone+start://'):
```
to factorize the comment and the check

lib/host.py
```diff
@@ -391,25 +391,27 @@ def cached_vm(self, uri: str, sr_uuid: str) -> VM | None:
...
+            assert use_cache, "clone URIs require cache enabled"
```
glehmann: Does it fail if we pass an existing VM name not generated with the cache feature?

lib/host.py
vxgmichel: I wasn't sure if there could be another `clone+something://` protocol used somewhere, but that seems unlikely. I'll apply your suggestion  👍

lib/host.py
vxgmichel: No it does not, that's typically the case for the nested hosts created by `tests/install`: they are generated using `vm.save_to_cache` and not `host.import_vm`.

Are you suggesting to get rid of this assert? I think it would make more sense too, I kept it because I meant for this commit to be a refactoring and not change the existing logic.

lib/host.py
vxgmichel: Applied here: 212b46ee0e047c57af414e2aeedc35f3355c8924

lib/host.py
glehmann: It's just that I don't see the point in preventing the usage of `clone://`, unless there is actually something wrong to do so.

So I would indeed recommend to not putting this assert.

ydirson: Right, there could be one more commit for this :)

ydirson (approved):
lgtm.
Note for later reviews: the 2nd commit contains changes to code that gets moved that the same time, it would have been much more rapidly reviewed if the move was in a separate commit that contains only the changes pertaining to the move, as `git show --color-moved --color-moved-ws=allow-indentation-change` would better visualize the move

lib/host.py
vxgmichel: Removed here: dbd8d9b33f7fb079ad15745ddf56372c5282bc6f

glehmann (approved):

PR #638, opened 2026-07-23
Add test for VDI resize support on LINSTOR SR

Kuruyia:
This adds the `test_resize_vdi` test in the LINSTOR SR storage test suite. This tests the VDI resize feature on this SR, and makes sure the VM can still start up successfully afterwards.

This also unifies the names of the tests for VDI resizing on the EXT and LVM SRs so that they follow the naming convention established by the related test for the LINSTOR SR.

This is a rebase of #273.

### DIFF
--- a/tests/storage/linstor/conftest.py
+++ b/tests/storage/linstor/conftest.py
@@ -118,14 +118,14 @@ def pool_with_linstor(
     import concurrent.futures
     pool = pool_with_saved_yum_state
 
-    def check_linstor_installed(host: Host) -> None:
+    def ensure_linstor_not_installed(host: Host) -> None:
         if host.is_package_installed(LINSTOR_PACKAGE):
             raise Exception(
                 f'{LINSTOR_PACKAGE} is already installed on host {host}. This should not be the case.'
             )
 
     with concurrent.futures.ThreadPoolExecutor() as executor:
-        executor.map(check_linstor_installed, pool.hosts)
+        executor.map(ensure_linstor_not_installed, pool.hosts)
 
     def install_linstor(host: Host) -> None:
         logging.info(f"Installing {LINSTOR_PACKAGE} on host {host}...")
--- a/tests/storage/linstor/test_linstor_sr.py
+++ b/tests/storage/linstor/test_linstor_sr.py
@@ -189,6 +189,23 @@ class TestLinstorSR:
         finally:
             vm.shutdown(verify=True)
 
+    @pytest.mark.small_vm
+    @pytest.mark.big_vm
+    def test_resize_vdi(self, vm_on_linstor_sr: VM) -> None:
+        vm = vm_on_linstor_sr
+        if vm.is_running():
+            vm.shutdown(verify=True)
+
+        for vdi in vm.vdis:
+            size = vdi.get_virtual_size()
+            vdi.resize(size * 2)
+            assert vdi.get_virtual_size() == size * 2
+
+        # Make sure the VM still starts up successfully after resizing
+        vm.start()
+        vm.wait_for_os_booted()
+        vm.shutdown(verify=True)
+
     # *** tests with reboots (longer tests).
 
     @pytest.mark.reboot
--- a/tests/storage/ext/test_ext_sr.py
+++ b/tests/storage/ext/test_ext_sr.py
@@ -106,7 +106,7 @@ class TestEXTSR:
 
     @pytest.mark.small_vm
     @pytest.mark.big_vm
-    def test_resize(self, vm_on_ext_sr: VM) -> None:
+    def test_resize_vdi(self, vm_on_ext_sr: VM) -> None:
         vm = vm_on_ext_sr
         vdi = VDI(vm.vdi_uuids()[0], host=vm.host)
         old_size = vdi.get_virtual_size()
@@ -118,7 +118,7 @@ class TestEXTSR:
 
     @pytest.mark.small_vm
     @pytest.mark.big_vm
-    def test_failing_resize(self, host: Host, ext_sr: SR, vm_on_ext_sr: VM, exit_on_fistpoint: None) -> None:
+    def test_failing_resize_vdi(self, host: Host, ext_sr: SR, vm_on_ext_sr: VM, exit_on_fistpoint: None) -> None:
         vm = vm_on_ext_sr
         vdi = VDI(vm.vdi_uuids()[0], host=vm.host)
         old_size = vdi.get_virtual_size()
--- a/tests/storage/lvm/test_lvm_sr.py
+++ b/tests/storage/lvm/test_lvm_sr.py
@@ -65,7 +65,7 @@ class TestLVMSR:
 
     @pytest.mark.small_vm
     @pytest.mark.big_vm
-    def test_failing_resize_on_inflate_after_setSize(
+    def test_failing_resize_vdi_on_inflate_after_setSize(
         self, host: Host, lvm_sr: SR, vm_on_lvm_sr: VM, exit_on_fistpoint: None
     ) -> None:
         vm = vm_on_lvm_sr
@@ -93,7 +93,7 @@ class TestLVMSR:
 
     @pytest.mark.small_vm
     @pytest.mark.big_vm
-    def test_failing_resize_on_inflate_after_setSizePhys(
+    def test_failing_resize_vdi_on_inflate_after_setSizePhys(
         self, host: Host, lvm_sr: SR, vm_on_lvm_sr: VM, exit_on_fistpoint: None
     ) -> None:
         vm = vm_on_lvm_sr


lib/vm.py
```diff
@@ -549,6 +552,24 @@ def test_snapshot_on_running_vm(self) -> None:
         finally:
             snapshot.destroy(verify=True)
 
+    def vdi_resize(self, vdi_uuid: str, new_size: Optional[int] = None):
```
glehmann:
```suggestion
    def vdi_resize(self, vdi_uuid: str, new_size: int | None = None):
```

lib/vm.py
```diff
@@ -549,6 +552,24 @@ def test_snapshot_on_running_vm(self) -> None:
...
+        assert self.is_halted(), "Expected halted, got running"
+
+        vdi = VDI(vdi_uuid, sr=self.get_sr())
+        if new_size is None:
+            # Resize by double, ideally it should be within min and max limits reference
+            vdi_size = int(vdi.param_get("virtual-size"))
+            new_size = vdi_size * 2
+
+        vdi.resize(new_size)
+        resized_vdi_size = int(vdi.param_get("virtual-size"))
```
glehmann:
```suggestion
        resized_vdi_size = vdi.get_virtual_size()
```

lib/vm.py
```diff
@@ -549,6 +552,24 @@ def test_snapshot_on_running_vm(self) -> None:
...
+        assert resized_vdi_size == new_size, f"Expected {new_size}, got {resized_vdi_size}"
```
glehmann: resize() silently round the expected size up so the real size is a multiple of the block size, so this assert might fail.
IMO it's better to return the actual size and let the test check if this value fits the test needs.

lib/vm.py
```diff
@@ -549,6 +552,24 @@ def test_snapshot_on_running_vm(self) -> None:
...
+
     def get_messages(self, name: str) -> List[str]:
```
glehmann, line 552: Not used in this PR? If yes, it could be removed.

lib/vm.py
```diff
@@ -549,6 +552,24 @@ def test_snapshot_on_running_vm(self) -> None:
...
+        # Reference
+        # MAX_VHD_SIZE = 2088960 * 1024 * 1024 # 2088960 MB
+        # MIN_VHD_SIZE = 1 * 1024 * 1024 # 1MB
+
+        # Following assert may be removed in future if we support online vdi resize
+        assert self.is_halted(), "Expected halted, got running"
+
+        vdi = VDI(vdi_uuid, sr=self.get_sr())
+        if new_size is None:
+            # Resize by double, ideally it should be within min and max limits reference
```
glehmann: We now have `QCOW2_MAX` and `VHD_MAX` in `lib.common`, so it can be validated (if we want to).

lib/vm.py
```diff
@@ -549,6 +552,24 @@ def test_snapshot_on_running_vm(self) -> None:
...
+            vdi_size = int(vdi.param_get("virtual-size"))
```
glehmann:
```suggestion
            vdi_size = vdi.get_virtual_size()
```

lib/vm.py
```diff
@@ -549,6 +552,24 @@ def test_snapshot_on_running_vm(self) -> None:
...
 
+    def vdi_resize(self, vdi_uuid: str, new_size: Optional[int] = None):
+        # Reference
+        # MAX_VHD_SIZE = 2088960 * 1024 * 1024 # 2088960 MB
+        # MIN_VHD_SIZE = 1 * 1024 * 1024 # 1MB
+
+        # Following assert may be removed in future if we support online vdi resize
+        assert self.is_halted(), "Expected halted, got running"
+
+        vdi = VDI(vdi_uuid, sr=self.get_sr())
```
glehmann: We should check that this VDI is indeed linked to the current VM

tests/storage/linstor/test_linstor_sr.py
```diff
@@ -189,6 +189,18 @@ def test_resynchronization(
...
+    @pytest.mark.small_vm
+    @pytest.mark.big_vm
+    def test_resize_vdi(self, vm_on_linstor_sr: VM) -> None:
+        if vm_on_linstor_sr.is_running():
+            vm_on_linstor_sr.shutdown(verify=True)
+
+        vm = vm_on_linstor_sr
+        logging.info("VDI Resize started")
+        for vdi_uuid in vm.vdi_uuids():
+            vm.vdi_resize(vdi_uuid)
```
glehmann: I'm not sure we need the VM.vdi_resize() method at all. I could just be:

```suggestion
        assert self.is_halted(), "Expected halted, got running"
        for vdi in vm.vdis:
            size = vdi.get_virtual_size()
            vm.resize(size * 2)
            assert vdi.get_virtual_size() == size * 2
```

It could also be interesting to boot the VM to validate that it runs properly.

glehmann (commented):
I'm not sure we need the first commit, please see the comment on test_linstor_sr.py

tests/storage/linstor/test_linstor_sr.py
Kuruyia: I went ahead with your suggestion and dropped the first commit entirely, as I don't think it was needed either.

I'd like to take into account one of your other comments about block size, but can't find a way to get this information. Perhaps you know how to do that?

glehmann: I don't know, but I don't think it matters now that the code just doubles the size—doubling the size that was already a multiple of the block size would still be a multiple of the block size :)

glehmann (approved):

tests/storage/linstor/test_linstor_sr.py
Kuruyia: Oh yeah I'm dumb, that's true. For some reason I was thinking about what would happen if the original size wasn't a multiple of the block size, but didn't stop to think about whether that scenario was possible ^^'

tests/storage/linstor/conftest.py
```diff
@@ -118,14 +118,14 @@ def pool_with_linstor(
     import concurrent.futures
     pool = pool_with_saved_yum_state
 
-    def check_linstor_installed(host: Host) -> None:
+    def assert_linstor_not_installed(host: Host) -> None:
```
rzr:
```suggestion
    def ensure_linstor_not_installed(host: Host) -> None:
```

nitpicking: I prefer the original name to assert, because technically it's not an assert, but a function throwing an exception. 

Imagine someday we decide to scan/deal with asserts in codebase, then this one would be a false positive.

tests/storage/linstor/conftest.py
```diff
@@ -118,14 +118,14 @@ def pool_with_linstor(
...
         if host.is_package_installed(LINSTOR_PACKAGE):
             raise Exception(
                 f'{LINSTOR_PACKAGE} is already installed on host {host}. This should not be the case.'
             )
 
     with concurrent.futures.ThreadPoolExecutor() as executor:
-        executor.map(check_linstor_installed, pool.hosts)
+        executor.map(assert_linstor_not_installed, pool.hosts)
```
rzr:
```suggestion
        executor.map(ensure_linstor_not_installed, pool.hosts)
```

tests/storage/linstor/test_linstor_sr.py
```diff
@@ -189,6 +189,23 @@ def test_resynchronization(
...
+        vm = vm_on_linstor_sr
+        if vm.is_running():
+            vm.shutdown(verify=True)
+
+        for vdi in vm.vdis:
+            size = vdi.get_virtual_size()
+            vdi.resize(size * 2)
+            assert vdi.get_virtual_size() == size * 2
+
+        # Make sure the VM still starts up successfully after resizing
```
rzr, line 204:
```suggestion
        # Make sure the VM still start up successfully after resizing
```

Kuruyia: The grammar on the original comment is correct :)

tests/storage/linstor/test_linstor_sr.py
```diff
@@ -189,6 +189,23 @@ def test_resynchronization(
         finally:
             vm.shutdown(verify=True)
 
+    @pytest.mark.small_vm
+    @pytest.mark.big_vm
+    def test_resize_vdi(self, vm_on_linstor_sr: VM) -> None:
```
Wescoeur, line 194: There are these tests in `test_ext_sr.py`: `test_resize` and `test_failing_resize`.
Regarding the naming, I would tend to say that these should be modified to include the `_vdi` suffix, as shown here.

Wescoeur (changes requested):

tests/storage/linstor/test_linstor_sr.py
Wescoeur: I suppose "still" was read as "will". 🕵️

tests/storage/linstor/test_linstor_sr.py
Kuruyia: I don't know whether this is in-scope for this PR?

Wescoeur: Adding new functions similar to those of another test but using a different name raises questions for future maintenance. This standardization of this naming could be handled in a dedicated commit.
To be a bit cheeky: renaming "assert_linstor_not_installed" isn't within the scope either, but it's the kind of change that's acceptable in a dedicated commit. 😛

tests/storage/linstor/conftest.py
glehmann: Technically, `assert` raises an exception in Python, so we could discuss if the `assert` prefix makes sense ;)

Wescoeur (approved):

rzr (approved):

PR #644, opened 2026-07-29
l/host.py: Fix issue that update from non declared repositories

rzr:
While updating some pool, I noticed that some packages from non listed repo were also updated, this is caused by previously enabled repos on host. This behavior makes the update unpredictable (unless if all/undesired repos are disabled on host explicitly before the update).

With this change all desired repos must be explicitly declared into inventory file, non declared repos will not be used anymore.

To reproduce the issue, on an fresh host do the update with

    [default]
    repositories = ["xcp-ng-base","xcp-ng-testing", "xcp-ng-candidates"]
    [hosts."$host"]

Then:

    ssh root@$host yum-config-manager --enable "xcp-ng-ci"

And run same update again: ci packages will be pulled while not desired in the inventory file.

### DIFF
--- a/lib/host.py
+++ b/lib/host.py
@@ -511,27 +511,33 @@ class Host:
         logging.info(f"[{self}] Removing cache metadata...")
         return self.ssh("yum clean metadata -q")
 
-    def yum_update(self, enablerepos: list[str] = []) -> str:
+    def yum_update(self, enablerepos: list[str] = [], disablerepos: list[str] = []) -> str:
         """Updates packages on target.
 
         Performs the following shell command::
 
             yum update -y
             # with enablerepos
-            yum update -y --disablerepo='*' --enablerepo=extra1 --enablerepos=extra2
+            yum update -y --enablerepo=extra1 --enablerepos=extra2
+            # with disablerepos
+            yum update -y --disablerepo=extra1 --disablerepos=extra2
 
-        :param enablerepos: Enable one or more repositories (default: []) if present disable other repos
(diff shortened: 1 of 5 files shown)


lib/host.py
```diff
@@ -518,16 +518,16 @@ def yum_update(self, enablerepos: list[str] = []) -> str:
...
-        :param enablerepos: Enable one or more repositories (default: []).
+        :param enablerepos: Enable one or more repositories (default: []) if present disable other repos
         """
         base_command = "yum update -y"
 
         logging.info(f"[{self}] Updating packages...")
         if enablerepos:
             extra = " ".join(f"--enablerepo={r}" for r in enablerepos)
-            base_command = f"{base_command} {extra}"
+            base_command = f"{base_command} --disablerepo='*' {extra}"
```
olivierh-pro: So, if I understand it correctly, it does a basic `yum update` without repo and when there are "repos", first, it disables "existing old repos" and then enables "new from args".

Do I miss something?

rzr: yes, is disables all except those mentioned in toml file, see commit message there is a test case you can try

olivierh-pro (approved):
LGTM! Thanks you for this fix! :)

gduperrey (approved):

glehmann (commented):
We this change we wouldn't have a way to just activate some repos in addition to the default ones. Maybe it's better to add an option to disable some repos, that would support `*`?

glehmann (approved):

PR #647, opened 2026-08-03
Add clean tool to remove all VMs and local VDIs

glehmann:
Add a new 'clean' subcommand to tools.py that removes all VMs (with all
their VDIs, regardless of storage location) from the target pools and
then clears the remaining orphan VDIs on the local (non-shared) storage
SRs. Supports a --dry-run mode, and verifies that VMs and local SRs are
actually empty before returning so downstream scripts can rely on a
clean state.

### DIFF
--- a/lib/tools/cli.py
+++ b/lib/tools/cli.py
@@ -11,6 +11,7 @@ from pathlib import Path
 from lib.common import HostAddress
 from lib.tools import logger
 from lib.tools.inventory import into_inventory, load_inventory
+from lib.tools.tasks.clean import clean_pools
 from lib.tools.tasks.update import update_pools
 
 def _command_update(args: argparse.Namespace) -> None:
@@ -20,6 +21,15 @@ def _command_update(args: argparse.Namespace) -> None:
         inventory = into_inventory(args.hosts, args.repos, args.hosting_pool)
 
     update_pools(inventory)
+
+
+def _command_clean(args: argparse.Namespace) -> None:
+    if args.inventory:
+        inventory = load_inventory(args.inventory)
+    else:
+        inventory = into_inventory(args.hosts, [], args.hosting_pool)
+
+    clean_pools(inventory, dry_run=args.dry_run)
 
 
 def cli() -> None:
@@ -60,6 +70,31 @@ def cli() -> None:
         help="Address (hostname|ip) of hosting pool's master host (nested context)",
     )
     subparser_cmd_update.set_defaults(func=_command_update)
+
+    # subparser - command: clean
+    subparser_cmd_clean = subparsers.add_parser(
+        name="clean",
+        description="Remove all VMs and all VDIs on local storage from target pools",
+        help="Remove all VMs and all VDIs on local storage from target pools",
+    )
+    cmd_clean_excl_grp = subparser_cmd_clean.add_mutually_exclusive_group(required=True)
+    cmd_clean_excl_grp.add_argument(
+        "-H",
+        "--hosts",
+        type=HostAddress,
+        metavar="HOST",
+        nargs="+",
+        help="Address (hostname|ip) of the master host in pool",
+    )
+    cmd_clean_excl_grp.add_argument("-i", "--inventory", type=Path, help="Use an hosts inventory file")
+    subparser_cmd_clean.add_argument(
+        "-n",
+        "--dry-run",
+        action="store_true",
+        default=False,
+        help="Only display what would be removed, without deleting anything",
+    )
+    subparser_cmd_clean.set_defaults(func=_command_clean)
 
     args = parser.parse_args()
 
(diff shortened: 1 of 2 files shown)


lib/tools/tasks/clean.py
```diff
@@ -0,0 +1,90 @@
...
+        vm = VM(vm_uuid, master)
+        logger.info(f"[{master}] Removing VM {vm.uuid} ({vm.name()})")
+        if not dry_run:
+            vm.destroy(verify=True)
+
+    sr_uuids = local_sr_uuids(pool)
+    for sr_uuid in sr_uuids:
+        for vdi_uuid in orphan_vdi_uuids(sr_uuid, pool):
+            vdi = VDI(vdi_uuid, sr=SR(sr_uuid, pool))
+            logger.info(f"[{master}] Removing orphan VDI {vdi.uuid} from local SR {sr_uuid}")
```
olivierh-pro: IMO, this log should be displayed only if we are not dry.

lib/tools/cli.py
```diff
@@ -61,6 +71,37 @@ def cli() -> None:
...
+    cmd_clean_excl_grp.add_argument(
+        "-H",
+        "--hosts",
+        type=HostAddress,
+        metavar="HOST",
+        nargs="+",
+        help="Address (hostname|ip) of the master host in pool",
+    )
+    cmd_clean_excl_grp.add_argument("-i", "--inventory", type=Path, help="Use an hosts inventory file")
+    subparser_cmd_clean.add_argument(
```
olivierh-pro: It looks like you don't use "hosting pool" in the cleaning command, so, do we really need this arg here?

lib/tools/tasks/clean.py
```diff
@@ -0,0 +1,90 @@
...
+    pools: list[Pool] = []
+    for host in inventory_hosts:
+        try:
+            pools.append(Pool(host))
+        except NotAMasterHostError:
+            logger.warning(f"[{host}] Skipping: not a master host")
+
+    with ThreadPoolExecutor() as executor:
+        futures = {executor.submit(clean_pool, p, dry_run): p for p in pools}
+        for future in futures:
```
olivierh-pro, line 45: I've used the same pattern for `ThreadPoolExecutor` in *update_pools*, except for this line where I use [as_completed](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.as_completed) to get *finished* or *cancelled*.

Is there a reason for this difference? Maybe, I should refactor *update_pools* myself to be like your code 🤔

lib/tools/tasks/clean.py
glehmann: The point of dry-run is to be able to see what will be removed before actually doing it.
If we suppress it, we won't be able to do that

lib/tools/tasks/clean.py
```diff
@@ -0,0 +1,90 @@
...
+            )
+
+def local_sr_uuids(pool: Pool) -> list[str]:
+    """Return the UUIDs of the pool's local (non-shared, user) SRs."""
+    uuids = safe_split(pool.master.xe('sr-list', {'content-type': 'user'}, minimal=True))
+    return [uuid for uuid in uuids if not SR(uuid, pool).is_shared()]
+
+def orphan_vdi_uuids(sr_uuid: str, pool: Pool) -> list[str]:
+    """Return the UUIDs of the managed VDIs left on the SR."""
+    return SR(sr_uuid, pool).vdi_uuids(managed=True)
```
rzr: question: What makes you think there are orphans ? is managed=True doing the selection ?

if they are not orphaned maybe just rename to vdi_remaining or vdi_available

lib/tools/tasks/clean.py
glehmann: `as_completed()` is useful for fail-fast, a mechanism I haven't added in clean.py

lib/tools/cli.py
glehmann: done

lib/tools/tasks/clean.py
glehmann: I see that the name might be confusing, and this function does bring much. I just used the code it contains inline :)

lib/tools/tasks/clean.py
olivierh-pro: And what about adding "dry-run notification" in message log?

Something like this:

```python
if not dry_run:
    logger.info(f"[{master}] Removing orphan VDI {vdi.uuid} from local SR {sr_uuid}")
    vdi.destroy()
else:
    logger.info(f"[{master}] Orphan VDI {vdi.uuid} will be removed from local SR {sr_uuid}")
```

Or maybe directly in the *f-string*.

glehmann: done

olivierh-pro (approved):

rzr (approved):

PR #648, opened 2026-08-03
Skip host reboot when no packages were updated

glehmann:
During tools update, do not reboot a host if yum reports no package
was actually updated, avoiding unnecessary reboots when the host is
already up to date.

### DIFF
--- a/lib/host.py
+++ b/lib/host.py
@@ -536,6 +536,9 @@ class Host:
 
         An helper function that wraps update tasks on current host.
 
+        When reboot is requested, the host is only rebooted if packages were
+        actually updated.
+
         :param list[str] enablerepos:
             Repositories to enable when updating.
         :param bool reboot:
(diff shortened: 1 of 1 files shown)


lib/host.py
```diff
@@ -544,10 +544,12 @@ def update(self, enablerepos: list[str] = [], reboot: bool = True) -> None:
         logging.info(f"[{self}] Updating...")
 
         self.yum_clean_metadata()
-        self.yum_update(enablerepos=enablerepos)
-        if reboot:
+        output = self.yum_update(enablerepos=enablerepos)
+        if reboot and "No packages marked for update" not in output:
             # Everything's ok, just reboot
             self.reboot(verify=True)
+        else:
```
vxgmichel: The log message should only appear if `reboot` is `True` right?
```suggestion
        elif reboot:
```

vxgmichel (approved):
A minor comment, LGTM otherwise 👍

lib/host.py
```diff
@@ -544,10 +544,12 @@ def update(self, enablerepos: list[str] = [], reboot: bool = True) -> None:
         logging.info(f"[{self}] Updating...")
 
         self.yum_clean_metadata()
-        self.yum_update(enablerepos=enablerepos)
-        if reboot:
+        output = self.yum_update(enablerepos=enablerepos)
+        if reboot and "No packages marked for update" not in output:
```
olivierh-pro, line 551: Do you think it's a good idea to describe this behaviour in docstring of the method?

glehmann: done

gduperrey (approved):

PR #649, opened 2026-08-03
Add colored logging to tools CLI

glehmann:
Apply the colored logger used in the compat kit to the tools package,
so logs emitted by scripts/tools.py are shown with per-level ANSI colors.
Replaces the plain basicConfig logging setup with a ColoredFormatter.

### DIFF
--- a/lib/logger.py
+++ b/lib/logger.py
@@ -1,0 +1,39 @@
+"""Colored logging utilities."""
+
+import logging
+
+class ColoredFormatter(logging.Formatter):
+    """A logging formatter that adds ANSI color codes to log levels."""
+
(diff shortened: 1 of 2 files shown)


olivierh-pro (commented):
Based on [previous thread](https://github.com/xcp-ng/xcp-ng-tests/pull/431#discussion_r3159798766), would you plan to add this code snippet globally?

Maybe we should have a module `lib/logger.py` or something like that?

gduperrey (approved):

olivierh-pro (approved):
I've also tested with the `-d` (debug option). It works as expected!

Good job!

PR #650, opened 2026-08-03
update: report updated packages and verify consistency across hosts

glehmann:
After updating, display a summary of the packages updated on all hosts
plus per-host additions, and raise an error listing package set
differences if hosts end up inconsistent.

### DIFF
--- a/lib/tools/tasks/update.py
+++ b/lib/tools/tasks/update.py
@@ -13,6 +13,54 @@ from lib.tools.tasks.snapshot import create_snapshots
 
 from .. import logger
 
+def _capture_packages(pools: list[Pool]) -> dict[Host, set[str]]:
+    """Snapshot the installed packages of every host in the pools."""
+    return {h: set(h.packages()) for p in pools for h in p.hosts}
+
+def _filter_packages(pkgs: set[str]) -> set[str]:
+    return {p for p in pkgs if not p.startswith("gpg-pubkey-")}
+
+def _format_packages(pkgs: list[str]) -> str:
+    return "\n".join(f"  - {p}" for p in pkgs)
+
+def _report_updated(before: dict[Host, set[str]], after: dict[Host, set[str]]) -> None:
+    """Log a summary of the packages that were updated on each host."""
+    updated = {
+        h: _filter_packages(after[h] - pkgs) for h, pkgs in before.items()
+    }
+    common_updated = set.intersection(*updated.values()) if updated else set()
+
+    if not common_updated:
+        logger.info("No packages were updated on any host.")
+        return
+    logger.info(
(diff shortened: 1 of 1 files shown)


gduperrey (approved):

lib/tools/tasks/update.py
```diff
@@ -43,6 +43,12 @@ def update_pools(inventory: Inventory) -> None:
         except NotAMasterHostError:
             logger.warning(f"[{host}] Skipping: not a master host")
 
+    # capture package lists before updating
+    before_packages: dict[Host, set[str]] = {}
+    for p in pools:
+        for h in p.hosts:
+            before_packages[h] = set(h.packages())
```
olivierh-pro: I wonder if it could be interesting to implement a *pool's method* for this 🤔

lib/tools/tasks/update.py
```diff
@@ -80,6 +86,54 @@ def update_pools(inventory: Inventory) -> None:
...
+    }
+    if inconsistent:
+        lines = [
+            f"Not all hosts have the same set of packages "
+            f"(reference: common set of {len(after_packages)} hosts):"
+        ]
+        for h, extra_pkgs in inconsistent.items():
+            lines.append(f"  [{h}] additional packages:\n{_format_packages(sorted(extra_pkgs))}")
+        raise RuntimeError("\n".join(lines))
+
```
olivierh-pro, line 136: IMO, *diff operations* implemented above increase complexity of *update_pools*. Maybe we should wrap it in a new function next to *update_pools*, isn't it?

rzr: yes please make it optional because i know from experiences that some packages may differ (ie: installed test packages) of at least just show a warning.

glehmann: > IMO, *diff operations* implemented above increase complexity of *update_pools*. Maybe we should wrap it in a new function next to *update_pools*, isn't it?

I'll try

> yes please make it optional because i know from experiences that some packages may differ (ie: installed test packages) of at least just show a warning.

IMO it is important that the hosts are consistent—we should detect that there was a leftover from a previous experiment or a non-run teardown.
Could you say more about the use cases where inconsistent packages could be useful?

olivierh-pro (approved):

lib/tools/tasks/update.py
glehmann: I kept it local to update, as I don't see another usage for now

lib/tools/tasks/update.py
glehmann: I replaced the hard error by a warning

rzr (approved):

PR #654, opened 2026-08-05
tools update: add --no-reboot and --parallel to speedup the updates

glehmann:
Allow the update command to optionally skip rebooting hosts after package
updates (--no-reboot) and to update the master and secondary hosts
concurrently (--parallel).
Both are disabled by default. It's the responsibility of the user to
determine if using them is safe.

### DIFF
--- a/lib/tools/cli.py
+++ b/lib/tools/cli.py
@@ -20,7 +20,7 @@ def _command_update(args: argparse.Namespace) -> None:
     else:
         inventory = into_inventory(args.hosts, args.repos, args.hosting_pool, disabled_repositories=args.disablerepos)
 
-    update_pools(inventory)
+    update_pools(inventory, reboot=args.reboot, parallel=args.parallel)
 
 
 def _command_clean(args: argparse.Namespace) -> None:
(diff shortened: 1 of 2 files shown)


lib/tools/cli.py
```diff
@@ -76,6 +76,19 @@ def cli() -> None:
         type=HostAddress,
         help="Address (hostname|ip) of hosting pool's master host (nested context)",
     )
+    subparser_cmd_update.add_argument(
+        "--no-reboot",
+        action="store_false",
+        dest="reboot",
+        default=True,
+        help="Skip the reboot after updating packages",
```
rzr:
```suggestion
        help="Skip rebooting host after updating packages",
```

rzr (approved):

lib/tools/cli.py
olivierh-pro: I would have said `help="Don't reboot after update operation"`.

glehmann: done

olivierh-pro (approved):

PR #655, opened 2026-08-05
update: warn about installed packages unavailable in selected repos

glehmann:
After updating hosts, compare installed packages against what is available\
in the selected repositories and report any that are no longer present,\
listing the common ones first then per-host extras. Include a yum downgrade\
command to align hosts with the available packages.

This PR is part of a stack containing 4 PRs:

1. `master`
2. #654
3. **"gln/package-inconsistency-check-zkkx" (this PR)**
4. #656
5. #657

### DIFF
--- a/lib/host.py
+++ b/lib/host.py
@@ -670,8 +670,8 @@ class Host:
         return self.ssh(f'yum remove -y {" ".join(packages)}')
 
     def packages(self) -> list[str]:
-        """ Returns the list of installed RPMs - with version, release, arch and epoch. """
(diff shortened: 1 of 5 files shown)


lib/host.py
```diff
@@ -671,7 +671,7 @@ def yum_remove(self, packages: list[str]) -> str:
 
     def packages(self) -> list[str]:
         """ Returns the list of installed RPMs - with version, release, arch and epoch. """
```
```suggestion
        """ Returns the list of installed RPMs - with epoch, version, release, arch. """
```

olivierh-pro (approved):

lib/host.py
glehmann: done

rzr (approved):

PR #656, opened 2026-08-05
tools: add exec subcommand to run a command on all hosts

glehmann:
Add an 'exec' subcommand to tools.py that runs the same command on every
host of the target pools: on the master hosts first, then on the secondary
hosts (or all at once with --parallel). Support --dry-run and --reboot, and
exit non-zero if the command fails on any host.

This PR is part of a stack containing 4 PRs:

1. `master`
2. #654
3. #655
4. **"tools: add exec subcommand to run a command on all hosts" (this PR)**
5. #657

### DIFF
--- a/lib/tools/tasks/exec.py
+++ b/lib/tools/tasks/exec.py
@@ -1,0 +1,97 @@
+"""Exec tasks.
+
+This module is intended for running the same command on existing remote targets,
+on the master hosts first, then on the secondary hosts.
+"""
+from __future__ import annotations
+
+from concurrent.futures import ThreadPoolExecutor, as_completed
+
+from lib.host import Host
+from lib.pool import NotAMasterHostError, Pool
+from lib.tools.inventory import Inventory
+
+from .. import logger
+
+def _exec_on_host(host: Host, command: str, dry_run: bool, reboot: bool) -> int:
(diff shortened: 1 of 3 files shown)


lib/tools/tasks/exec.py
```diff
@@ -0,0 +1,97 @@
...
+    with ThreadPoolExecutor() as executor:
+        futures = {
+            executor.submit(_exec_on_host, host, command, dry_run, reboot): host for host in hosts
+        }
+        for future in as_completed(futures):
+            try:
+                if future.result() != 0:
+                    failures += 1
+            except Exception:
+                logger.error("Running command has failed!")
```
rzr, line 47: I would print result here since it's not returned

rzr (approved):

lib/tools/tasks/exec.py
glehmann: This is already printed in `_exec_on_host`. This try/except is just in case there is something we wouldn't have caught—I suppose we could just not catch it and display the stack trace instead

olivierh-pro (commented):
Would you like to provide an example use case for this implementation please?

I don't understand the purpose of this.

olivierh-pro (approved):

PR #659, opened 2026-08-05
guest_tools/win: Add VifConfigureFeature tests

dinhngtu:
The XenPlus Windows guest agent now supports xe vif-configure-*. Add some quick tests for these features.

### DIFF
--- a/tests/guest_tools/win/test_vif_configure.py
+++ b/tests/guest_tools/win/test_vif_configure.py
@@ -1,0 +1,149 @@
+import pytest
+
+import logging
+
+from lib.common import Defer, wait_for
+from lib.vif import VIF
+from lib.vm import VM
+from lib.windows import (
+    vif_add_manual_configuration,
+    vif_execute_powershell_script,
+    vif_exists,
+    vif_has_static_configuration,
+    vif_uses_dhcp,
+)
+
+from typing import Generator
+
+# Requirements:
+# - Same as TestGuestToolsWindowsNondestructive.
+
+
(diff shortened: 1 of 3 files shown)


casasnovas (approved):
Isn't there a way to poll for the interface to have an IP instead of the `time.sleep(2)` in `vif._configure` (within a fixed amount of time to not hog the test-suite in case of error)?  It looks like the kind of sleep that could introduce hard-to-debug/racy tests in the future.

Otherwise lgtm!

tests/guest_tools/win/test_vif_configure.py
```diff
@@ -0,0 +1,149 @@
...
+-AutomaticMetric Disabled `
+-InterfaceMetric 9999;
+Set-NetIPInterface -InterfaceIndex $adapter.ifIndex `
+-AddressFamily IPv6 `
+-AutomaticMetric Disabled `
+-InterfaceMetric 9999""",
+        )
+        yield vif
+
+    def test_vif_configure_ipv4(self, temporary_vif: VIF) -> None:
```
glehmann, line 53: Shouldn't it be split in one test by configuration instead of aggregating them in a single test?

glehmann (approved):

tests/guest_tools/win/test_vif_configure.py
dinhngtu: Thanks for approving. It was a bit of a pain to arrange testing of the config combinations, so I took a shortcut.

PR #664, opened 2026-08-10
tests: add sshd_config.d include regression test

LucienLassalle:
sshd on XCP-ng hosts has silently ignored /etc/ssh/sshd_config.d/*.conf drop-in files due to a missing Include directive in the packaged sshd_config. Add a quick host-only test that drops a harmless Banner directive in the directory and checks it shows up in `sshd -T`'s effective configuration.

Report: https://xcp-ng.org/forum/topic/12397/change-default-ssh-port

### DIFF
--- a/tests/system/test_sshd_config_include.py
+++ b/tests/system/test_sshd_config_include.py
@@ -1,0 +1,28 @@
+from lib.common import Defer
+from lib.host import Host
+
+# Regression test for sshd silently ignoring /etc/ssh/sshd_config.d/*.conf
+# drop-in files (missing "Include" directive in the packaged sshd_config).
+#
+# A drop-in file is created, setting a directive (a pre-auth Banner), that
+# isn't set anywhere else. `sshd -T` dumps the configuration as sshd itself
+# would apply it, so finding the directive there proves the drop-in directory
+# is actually included.
+#
+# Requirements:
+# - an XCP-ng host (--hosts) >= 8.2
+
+SSHD_CONFIG_D = "/etc/ssh/sshd_config.d"
+
+def test_sshd_config_d_is_included(host: Host, defer: Defer) -> None:
+    marker = host.ssh('mktemp')
+    dropin = f"{SSHD_CONFIG_D}/99-xcp-ng-tests-include-check.conf"
+
+    host.ssh(f"echo 'Banner {marker}' > {dropin}")
+    defer(lambda: host.ssh(f"rm -f {dropin}"))
+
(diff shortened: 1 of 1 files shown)


gduperrey (approved):

tests/system/test_sshd_config_include.py
```diff
@@ -0,0 +1,28 @@
...
+# would apply it, so finding the directive there proves the drop-in directory
+# is actually included.
+#
+# Requirements:
+# - an XCP-ng host (--hosts) >= 8.2
+
+SSHD_CONFIG_D = "/etc/ssh/sshd_config.d"
+
+def test_sshd_config_d_is_included(host: Host) -> None:
+    marker = host.ssh('mktemp')
```
glehmann, line 18:
```suggestion
    marker = uuid.uuid4()
```

We don't actually need this file, so it could be just a uuid or a timestamp

tests/system/test_sshd_config_include.py
```diff
@@ -0,0 +1,28 @@
...
+    dropin = f"{SSHD_CONFIG_D}/99-xcp-ng-tests-include-check.conf"
+
+    host.ssh(f"echo 'Banner {marker}' > {dropin}")
+    try:
```
glehmann: Using `defer` in place of the try/finally block would help to debug in case of problem, by not deleting the resources before entering the debuger

tests/system/test_sshd_config_include.py
LucienLassalle: Except that `banner` expects a path to a file ^^'.

tests/system/test_sshd_config_include.py
LucienLassalle: Like that ?

tests/system/test_sshd_config_include.py
rzr: you could then just use the dropin file wouldnt it work too ?

LucienLassalle: > you could then just use the dropin file wouldnt it work too ?

Pointing to the same file seems odd to me; I think creating a temporary file is more appropriate.
It's up to you to tell me what you prefer. ;)

rzr: no preference it's only a test

glehmann: ok, my bad!

tests/system/test_sshd_config_include.py
glehmann: yes 👍

glehmann (approved):

PR #665, opened 2026-08-13
quicktest: split QT into SR specific and common suites

gthvn1:
Currently we are calling quicktest for all SRs. The problem is that it takes long time to run and some tests are not related to the type of the SR. This patch creates one new test that runs the common suites one, and we only run SR specific tests per SR.

### DIFF
--- a/lib/sr.py
+++ b/lib/sr.py
@@ -1,6 +1,11 @@
 from __future__ import annotations
 
+import pytest
+
+import enum
+import functools
 import logging
+import shlex
 import time
 
 import lib.commands as commands
@@ -25,6 +30,33 @@ from typing import TYPE_CHECKING, Literal, overload
 if TYPE_CHECKING:
     from lib.host import Host
     from lib.pool import Pool
+
+QUICKTEST_BIN = "/opt/xensource/debug/quicktest"
+
+QUICKTEST_SR_SUITES = (
+    "cbt,copy,SR tests,Quicktest_vdi,Quicktest_async_calls,"
+    "Quicktest_vm_import_export,Quicktest_vm_lifecycle,Quicktest_vm_snapshot,"
+    "Quicktest_vdi_ops_data_integrity,Quicktest_max_vdi_size,Quicktest_static_vdis"
+)
+
+QUICKTEST_COMMON_SUITES = (
+    "Quicktest_example,Quicktest_message,xenstore,event,import_raw_vdi,"
+    "Quicktest_date,Quicktest_crypt_r,http,unixext,Timer"
+)
+
+class QuicktestScoping(enum.Enum):
+    NO_PARAM = enum.auto()
+    RUN_ONLY_PARAM = enum.auto()
+    SR_ONLY_PARAM = enum.auto()
+
+@functools.lru_cache(maxsize=None)
+def _quicktest_scoping(hostname_or_ip: str) -> QuicktestScoping:
+    help_output = commands.ssh(hostname_or_ip, f"{QUICKTEST_BIN} --help", check=False)
+    if "-sr-only" in help_output:
+        return QuicktestScoping.SR_ONLY_PARAM
+    if "-run-only" in help_output:
+        return QuicktestScoping.RUN_ONLY_PARAM
+    return QuicktestScoping.NO_PARAM
 
 class SR:
     xe_prefix = 'sr'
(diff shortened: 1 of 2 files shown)


lib/sr.py
```diff
@@ -26,6 +27,17 @@
...
+    "cbt,copy,SR tests,Quicktest_vdi,Quicktest_async_calls,"
+    "Quicktest_vm_import_export,Quicktest_vm_lifecycle,Quicktest_vm_snapshot,"
+    "Quicktest_vdi_ops_data_integrity,Quicktest_max_vdi_size,Quicktest_static_vdis"
+)
+
+QUICKTEST_COMMON_SUITES = (
+    "Quicktest_example,Quicktest_message,xenstore,event,import_raw_vdi,"
+    "Quicktest_date,Quicktest_crypt_r,http,unixext,Timer"
+)
+
```
stormi, line 47: My comment on this approach: IMO it should be quicktest that offers such a distinction rather than letting the caller use a whitelist of tests. It already has a filter to select the SR, maybe it could also accept a CLI option to separate common suites from SR suites ?

gthvn1: This solution uses "-run-only" to filter tests. The main advantage is that we have it ready right now and nothing is required from XAPI. But I agree that maintaining the list of what tests should be run is the main drawback.

It is completely doable to handle this in *quicktest* itself. This will require us to:
- modify the type of the test suite to add: SR|Common
- add new CLI flags (`-sr-only`, `-common-only`)
- modify `-list-tests` to print the new information, which I think would be useful

last-genius: Relying on a new version of quicktests with the new flags will mean xcp-ng-tests will no longer work on older xapis - how do we handle this elsewhere? Do we try to keep compatibility with older versions?

gthvn1: And just to give you a quick estimation. On my "lab" I was able to test with 5 different SR:
- with the split - 1:26:52
- without: 1:47:18
So I expect to win 4/5 minutes per SR tested. Thus if you test 10 different SRs in CI we can expect a gain of 50 minutes :)

stormi: > Relying on a new version of quicktests with the new flags will mean xcp-ng-tests will no longer work on older xapis - how do we handle this elsewhere? Do we try to keep compatibility with older versions?

We skip tests that wouldn't work with the tested numbered XCP-ng release (8.2 vs 8.3) but we don't test the XAPI version at the moment. It's OK for a test to require the latest released updates.

However, here maybe we could autodetect: -run-only to benefit from the optimization straight away, -sr-only/-common-only instead when quicktest supports it.

gthvn1: We can probably detect if `-sr-only` is available by checking the output of `commands.ssh(hostname_or_ip, "/opt/xensource/debug/quicktest --help", check=False)`. The result can be cached per host, so it is only run once. And depending if it is available or not we can use the new flags or not.
So the autodetect can have three cases selected by checking the output of the ssh on the host (8.2 wihtout any flags "current code" , 8.3 with -run-only, and 8.3 with -sr-only/-common-only). Or just current code  (8.2 and "old" 8.3, and 8.3 with -sr-only/-common-only if we don't want to maintain the list of tests in pytest).

stormi: Even if not cached, it's cheap :+1:

gthvn1:
And there is a PR to add `-sr-only` and `-common-only` parameters: https://github.com/xapi-project/xen-api/pull/7227
Note: This PR is for the master (our base for 9.0). The code is a little bit different for lcm (our base for 8.3) so if we need it we will need to provide another patch (easy but just to let you know).

gthvn1:
https://github.com/xapi-project/xen-api/pull/7227 has been merged. However, after review, a more generic mechanism was introduced in Quicktest that allows us to use tags. So to run tests that depend on `-sr`, you can now use `-with-tag sr`.

I have updated the PR accordingly. Note that this is currently only merged upstream (not in 8.3). However, the current patch selects specific tests using `-run-only` if the new option is unavailable. Also note that older Quicktest versions (such as those in 8.2) do not support `-run-only`, so there is a fallback that runs all tests except test_quicktest_common if no options are available.

stormi:
@gthvn1 It's still a draft, and re-view was not re-requested. Is that on purpose?

gthvn1:
@stormi , yes I was waiting a week to be sure that it passes the Citrix CI upstream and it is not reverted before asking new reviews.

gthvn1:
PRs are merged in master and lcm. It will be in our 8.3 when we will rebase 26.1.22. So let's review it.

lib/sr.py
```diff
@@ -26,6 +32,36 @@
...
+class QuicktestScoping(enum.Enum):
+    WITH_TAG_PARAM = enum.auto()
+    RUN_ONLY_PARAM = enum.auto()
+    NO_PARAM = enum.auto()
+
+@functools.lru_cache(maxsize=None)
+def _quicktest_scoping(hostname_or_ip: str) -> QuicktestScoping:
```
glehmann, line 54: nit: the host can be passed directly as a parameter and used as `host.ssh(…)`

glehmann (approved):
A comment or commit description about which XCP-ng or xapi version is supposed to go to which scoping path would probably be useful in the future, when we'll stop supporting the older versions, and we'll want to remove the scoping mechanism

Wescoeur (changes requested):
Just regarding commit title: `QT` is redundant with `quicktest`. Also is it a common acronym? Otherwise I have the impression that we are talking about qt5 or qt6 when reading this word. 😅 

Regarding the code: nothing to add.

gthvn1:
I will change the QT. It is an acronym but yes it depends on your background :)

gthvn1:
> A comment or commit description about which XCP-ng or xapi version is supposed to go

I have updated the commit message with both versions, XAPI master and lcm.

Wescoeur (approved):

PR #675, opened 2026-08-19
Make tools.py executable

glehmann:
So we can run it without prefixing it with `uv run`.


vxgmichel (approved):

vxgmichel (commented):
Wait, `tools.py` is missing a shebang right?

glehmann:
Good catch, I forgot to squash that in this commit :sweat:

vxgmichel (approved):

gduperrey (approved):

